PR #3003, opened 2023-01-30
Issues misconfigured

wadoon:
Due to missing issue templates and misconfiguration, nobody can currently create an issue. 

This commit fixes this, and also repairs the link to the FAQs.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,7 +1,7 @@
-blank_issues_enabled: false
+blank_issues_enabled: true
 contact_links:
   - name: FAQ
-    url: https://key-project.org/docs/faq
+    url: https://www.key-project.org/docs/user/FAQ/
     about: Documentation 
   - name: KeY Discussions
     url: https://github.com/keyproject/key/discussions


mattulbrich:
Hang on.

I think it would be a very good idea that we use templates instead of free text for bug reports. @WolframPfeifer said he would look into this soon and provide templates (or even forms)

mi-ki:
I fully agree with @mattulbrich on this point. Starting the new process with (allowing for) chaotic issue reporting does not appear sensible to me. Hence, this is not a misconfiguration in my eyes.

wadoon:
> Hang on.
> 
> I think it would be a very good idea that we use templates instead of free text for bug reports. @WolframPfeifer said he would look into this soon and provide templates (or even forms)

The creation of blank issues is not that bad. I am not aware when #3004 will be ready, but currently it is not possible to create an issue! And until #3004, we should just allow it.

The difference is a small hint at the end of the list, that blank issues are allowed, [see here](https://github.com/wadoon/key-playground/issues/new/choose). 

On the other side, "issue templates" do not guarantee anything. The user can delete everything in the issue and just start writing own unstructured reports.

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -1,11 +1,11 @@
-blank_issues_enabled: false
+blank_issues_enabled: true
```
mi-ki, line 1: Blank issues are not a sensible idea for a structured development process.

mi-ki (changes requested):
The proposed pull request is neither a fix nor an improvement. Instead, e.g., issue forms with mandatory fields should be used.

mattulbrich:
I've made a suggestion to #3004 and think we can merge that now ... it'll do for the moment. ...

PR #3004, opened 2023-01-30
Update issue templates

WolframPfeifer:
added templates for bug report and feature request

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -19,13 +19,13 @@ Steps to reproduce the behavior:
 **Expected behavior**
 A clear and concise description of what you expected to happen.
 
-**Desktop (please complete the following information):**
- - OS: [e.g. iOS]
- - Browser [e.g. chrome, safari]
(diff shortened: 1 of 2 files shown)


mattulbrich:
Can we not essentially take the gitlab template?

In particular: stacktraces should be mentioned, reproducibility, and the commit SHA should be mandatory.
(And I prefer headings using `#` over `**`)

wadoon:
Are the markdown templates the first step? And we will go for the formula solution in a second step?

mattulbrich:
Yes, @wadoon, sounds good. I have copied the MDs from gitlab and adapted them to github. I'd be happy with that for the moment. Of course: Can be improved in a next iteration.

Can we undraft this, @WolframPfeifer ?

WolframPfeifer:
My original plan was to use forms for this, but I do not have the time at the moment to prepare them sufficiently. So let's use the markdown templates for the moment.
Thanks, @mattulbrich!

Since I can not be reviewer (probably because I created the PR originally), @mi-ki will do the review.

mi-ki (approved):
Seems good to me, thanks for the effort!

PR #3005, opened 2023-01-31
Fix NPE when no look and feel is configured

### LINKED ISSUE #1741: LookAndFeel - Error message
This issue was created at [git.key-project.org](https://git.key-project.org/key/key/-/issues/1741) where the discussions are preserved.
    
----
    

## Description

The look-and-feel property management adds oddly and throws an exception

## Reproducible
  
always

### Steps to reproduce 

Start Key w/o a Look-and-feel configured

### Additional information

Apparently for some reason, the String "null" seems to be used if the property is not set locally.

```
ERROR MainWindow - failed to set look and feel java.lang.ClassNotFoundException: null
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:581)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:522)
	at java.base/java.lang.Class.forName0(Native Method)
	at java.base/java.lang.Class.forName(Class.java:398)
	at java.desktop/javax.swing.SwingUtilities.loadSystemClass(SwingUtilities.java:2036)
	at java.desktop/javax.swing.UIManager.setLookAndFeel(UIManager.java:632)
	at de.uka.ilkd.key.gui.MainWindow.setLaF(MainWindow.java:485)
	at de.uka.ilkd.key.gui.MainWindow.<init>(MainWindow.java:337)
	at de.uka.ilkd.key.gui.MainWindow.getInstance(MainWindow.java:427)
	at de.uka.ilkd.key.gui.MainWindow.getInstance(MainWindow.java:415)
	at de.uka.ilkd.key.core.Main.createUserInterface(Main.java:554)
	at de.uka.ilkd.key.core.Main.main(Main.java:191)
```

---

* Commit: c86de8ee9bee905daa21bbe6cb570220b7717ee4
    
----
    
Information: 
* created_at: 2023-01-29T18:38:15.415Z
* updated_at: 2023-01-29T18:38:15.415Z
* milestone: 
* user_notes_count: 0

FliegendeWurst:
This fixes #1741.

### DIFF
--- a/key.ui/src/main/java/de/uka/ilkd/key/gui/MainWindow.java
+++ b/key.ui/src/main/java/de/uka/ilkd/key/gui/MainWindow.java
@@ -408,7 +408,7 @@ public final class MainWindow extends JFrame {
         try {
             String className =
                 ProofIndependentSettings.DEFAULT_INSTANCE.getViewSettings().getLookAndFeel();
-            if (className != null) {
+            if (className != null && !className.equals("null")) {
                 UIManager.setLookAndFeel(className);
 
                 // Workarounds for GTK+


mattulbrich:
Cool, thanks. This fixes this symptom, and perhaps it is not the time to look for the reason why we store a non-existing key as the string "null" in KeY settings. It remains strange. Could you please add a line of comment pointing out the reason for the strange check. I cannot write to your repo.

key.core/src/main/java/de/uka/ilkd/key/settings/AbstractPropertiesSettings.java
```diff
@@ -212,9 +213,12 @@ public void parseFrom(String value) {
         @Override
         public void set(T value) {
             T old = get();
-            properties.setProperty(key, toString.apply(value));
-            if (value != null && !value.equals(old)) {
-                fireSettingsChange();
+            // only store non-null values
```
mattulbrich, line 216: Caveat: This means, we cannot use "set(null)" to unset an entry. Perhaps it is not needed since it was not possible till now anyway.

mattulbrich (approved):
Thanks for adding comments and for removing the rather strange behaviour.

key.core/src/main/java/de/uka/ilkd/key/settings/AbstractPropertiesSettings.java
FliegendeWurst: `set(null)` wasn't possible prior to the LookAndFeel configuration option (see [that commit](https://github.com/KeYProject/key/commit/d0cdd23161e94ebfc02aef4bba910fa4ef0c0bb6#diff-bbf9e4af8a6ea6c88dd1c0e2fa53264e5e087c21e119ddf27dbda007f00d903cR216))

PR #3006, opened 2023-01-31
Selection history (back + forward button)

FliegendeWurst:
Continuation of https://git.key-project.org/key/key/-/merge_requests/629 (all review comments addressed)


key.ui/src/main/java/de/uka/ilkd/key/gui/MainWindow.java
```diff
@@ -85,10 +85,6 @@ public final class MainWindow extends JFrame {
      * SequentView for the current goal
      */
     public final CurrentGoalView currentGoalView;
-    /**
-     * The menu for the SMT solver options
-     */
-    public final JMenu smtOptions = new JMenu("SMT Solvers...");
```
wadoon, line 91: I guess this field was unused. And why do we have public fields in `MainWindow`?

key.ui/src/main/java/de/uka/ilkd/key/gui/SelectionHistory.java
```diff
@@ -0,0 +1,183 @@
...
+ */
+public class SelectionHistory implements KeYSelectionListener, ProofDisposedListener {
+    /**
+     * The KeY mediator.
+     */
+    private final KeYMediator mediator;
+    /**
+     * Previously selected nodes by the user.
+     */
+    private final Deque<Node> selectedNodes = new ArrayDeque<>();
```
wadoon, line 32: This is dangerous! You created a memory leak. 

Keeping proof nodes alive, results into the case where proofs are kept alive (reachable via a pointer in node.), and when proofs are reachable, everything belonging to a proof is reachable. Reachable objects are not removed by GC. 

You should either use `WeakReference`s or try to store these information inside the user data of a proof.

wadoon (changes requested):
Everything is fine. Only the memory leakage is an obstacles of acceptance.

key.ui/src/main/java/de/uka/ilkd/key/gui/SelectionHistory.java
FliegendeWurst: Oh, thanks for the heads-up. I fixed it using `WeakReference`s.

key.ui/src/main/java/de/uka/ilkd/key/gui/MainWindow.java
FliegendeWurst: Yes, the field was not used at all. I have made the other public fields private.

WolframPfeifer:
@wadoon Merging is blocked at the moment. I think you have to approve that your requested changes have been applied ...

wadoon (approved):

PR #3012, opened 2023-02-01
Proof tree view: Multiple small changes for readability

jwiesler:
I've digged these changes up from my main branch from over a year ago and made them cleaner.

* Cleanup some of the code
* Speed up ProofTreeSearchBar by caching the search strings (Makes a huge difference for large proofs)
* Fix example dialog loading previously hidden example on double-click on a folder

### Elide long rule app names and move them to the tooltip

grafik

### Normalize whitespace in rule names
After: grafik

Before: grafik

### Move applied on term of OSS nodes to the tooltip

After: 
grafik

Before: 
grafik

### DIFF
--- a/key.ui/src/main/java/de/uka/ilkd/key/gui/ExampleChooser.java
+++ b/key.ui/src/main/java/de/uka/ilkd/key/gui/ExampleChooser.java
@@ -241,10 +241,6 @@ public final class ExampleChooser extends JDialog {
         exampleList.addMouseListener(new MouseAdapter() {
             @Override
             public void mouseClicked(MouseEvent e) {
-                // select closest tree entry to ensure that double clicks
-                // always open the entry the double click was performed on
-                TreePath closest = exampleList.getClosestPathForLocation(e.getX(), e.getY());
-                exampleList.setSelectionPath(closest);
                 if (e.getClickCount() == 2) {
                     loadButton.doClick();
                 }
--- a/key.ui/src/main/java/de/uka/ilkd/key/gui/ExampleChooser.java
+++ b/key.ui/src/main/java/de/uka/ilkd/key/gui/ExampleChooser.java
@@ -241,9 +241,14 @@ public final class ExampleChooser extends JDialog {
         exampleList.addMouseListener(new MouseAdapter() {
             @Override
             public void mouseClicked(MouseEvent e) {
-                if (e.getClickCount() == 2) {
-                    loadButton.doClick();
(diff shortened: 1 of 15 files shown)


key.ui/src/main/java/de/uka/ilkd/key/gui/ExampleChooser.java
```diff
@@ -241,10 +241,6 @@ private ExampleChooser(File examplesDir) {
         exampleList.addMouseListener(new MouseAdapter() {
             @Override
             public void mouseClicked(MouseEvent e) {
-                // select closest tree entry to ensure that double clicks
-                // always open the entry the double click was performed on
-                TreePath closest = exampleList.getClosestPathForLocation(e.getX(), e.getY());
-                exampleList.setSelectionPath(closest);
```
FliegendeWurst, line 247: Why is this removed?

jwiesler: I didn't find any positive case where this actually does anything.

Other than that, as I stated, double-clicking on a folder instantly loads a previously hidden example when the list is repositioned (e.g. when fully scrolled down).  The list is expanded, repositioned, then the event handler is catched and counts the double click on a folder as a double click on the repositioned list which is completely wrong.

mattulbrich:
Actually, I do like the new tooltip mechanism. It works for me. Question would be: Why only on nodes in OSS and not on regular applications?

key.ui/src/main/java/de/uka/ilkd/key/gui/ExampleChooser.java
FliegendeWurst: This fix does something if the user clicks left of an entry (hitting the background of the tree view). In that case, the code would load the last selected example and not the one the user actually double clicked at.

jwiesler: Ah I see the problem, good catch.

The problem is that a double click *anywhere* in the tree leads to a load. I pushed a fix to check whether an actual item was doubleclicked. Could you check if it works as intended and close this conversation, please?

FliegendeWurst: Thank you. The fix works as expected.

jwiesler:
Tooltips work for non OSS nodes as well now.

wadoon (approved):

PR #3013, opened 2023-02-01
Fix non duplicate feature perf

jwiesler:
I optimized the non duplicate app feature calculation by caching a map `rule name => [RuleApp]`.
This lead to an overall performance gain of about 20% for most cost computations done by the strategies.

I made some assumptions about `Node` that are not guaranteed by the interface (although I think they should be and `Node` could easily be made immutable).
See the doc of `getRuleAppsWithName`.

Two more things I noticed: 
* `NonDuplicateAppFeature::containsRuleApp` is always inverted
* Inverting the method to "does not contain" makes the code equal to `noDuplicateFindTaclet`, so I called this one instead

Is this really the case? `NonDuplicateAppFeature::filter` does a (now useless) case distinction over whether the position in occurrence is null and calls one of the two methods... It would be very odd if they did the same thing. Why would it be there if they were?

### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/java/ServiceCaches.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/java/ServiceCaches.java
@@ -18,6 +18,7 @@ import de.uka.ilkd.key.rule.metaconstruct.arith.Polynomial;
 import de.uka.ilkd.key.strategy.IfInstantiationCachePool;
 import de.uka.ilkd.key.strategy.RuleAppCost;
 import de.uka.ilkd.key.strategy.feature.AbstractBetaFeature.TermInfo;
+import de.uka.ilkd.key.strategy.feature.AppliedRuleAppsNameCache;
 import de.uka.ilkd.key.strategy.quantifierHeuristics.ClausesGraph;
 import de.uka.ilkd.key.strategy.quantifierHeuristics.TriggersSet;
 import de.uka.ilkd.key.util.Pair;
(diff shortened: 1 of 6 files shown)


key.core/src/main/java/de/uka/ilkd/key/java/ServiceCaches.java
```diff
@@ -139,6 +140,9 @@ public class ServiceCaches {
     private final IfFormulaInstantiationCache ifFormulaInstantiationCache =
         new IfFormulaInstantiationCache();
 
+    /** applied rule apps name cache */
```
wadoon, line 143: A typical comment why I do not believe in Javadoc.

key.core/src/main/java/de/uka/ilkd/key/strategy/feature/AppliedRuleAppsNameCache.java
```diff
@@ -0,0 +1,126 @@
...
+
+/**
+ * Establishes a cache for the applied rule apps to query them by name.
+ * See the get method for additional required constraints for correctness.
+ *
+ * @author Julian Wiesler
+ */
+public class AppliedRuleAppsNameCache {
+    /** cache of all applied rules by name of a node */
+    private final LRUCache<Node, HashMap<Name, List<RuleApp>>> cache = new LRUCache<>(32);
```
wadoon, line 22: Is the number of 32 enough. Is the performance/cache hits better with higher number?

wadoon (approved):

PR #3014, opened 2023-02-01
Spec math mode

jwiesler:
This MR implements the spec math modes discussed in a recent KaKeY meeting. See [here](https://github.com/KeYProject/key/wiki/Spec-math-modes) for a reference. Furthermore, safe math is implemented using the wd checker, this mode generates additional in bounds checks when doing a wd proof and is otherwise treated like the bigint mode.

The *code* overflow behaviour remains unchanged and is to be overhauled in a later MR. This MR only includes the changes to specification translation.

Tests do exists [here](https://git.key-project.org/key/key/-/merge_requests/608/diffs#diff-content-d937fae0cf5a262edfce5b782ad8e7a1a6c33155).
All other tests are green and only minor changes to the specification of the existing examples were necessary.

@unp1 @mulbrich


wadoon:
@jwiesler Could you migrate the linked wiki page to Github?

mattulbrich (approved):

PR #3015, opened 2023-02-02
Fix branch name in Checkstyle script

FliegendeWurst:
Simple fix after the recent branch rename.


wadoon (commented):
Thank you. Sounds good.

wadoon (approved):

PR #3016, opened 2023-02-02
Draft: Fix javacc deprecation warning

jwiesler:
Fixes the warning
```
> Task :key.core:compileJava
/home/runner/work/key/key/key.core/build/generated-src/javacc/main/de/uka/ilkd/key/parser/proofjava/JavaCharStream.java:348: warning: [dep-ann] deprecated item is not annotated with @Deprecated
  static public int getColumn() {
                    ^
/home/runner/work/key/key/key.core/build/generated-src/javacc/main/de/uka/ilkd/key/parser/proofjava/JavaCharStream.java:357: warning: [dep-ann] deprecated item is not annotated with @Deprecated
  static public int getLine() {
                    ^
/home/runner/work/key/key/key.core/build/generated-src/javacc/main/de/uka/ilkd/key/parser/schemajava/JavaCharStream.java:348: warning: [dep-ann] deprecated item is not annotated with @Deprecated
  static public int getColumn() {
                    ^
/home/runner/work/key/key/key.core/build/generated-src/javacc/main/de/uka/ilkd/key/parser/schemajava/JavaCharStream.java:357: warning: [dep-ann] deprecated item is not annotated with @Deprecated
  static public int getLine() {
                    ^
```


key.core/src/main/javacc/de/uka/ilkd/key/parser/proofjava/JavaCharStream.java
```diff
@@ -0,0 +1,588 @@
...
+           buffer[bufpos] = c = (char)(hexval(c) << 12 |
+                                       hexval(ReadByte()) << 8 |
+                                       hexval(ReadByte()) << 4 |
+                                       hexval(ReadByte()));
+
+           column += 4;
+        }
+        catch(java.io.IOException e)
+        {
+           throw new Error("Invalid escape character at line " + line +
```
mattulbrich, line 324: As mentioned during our discussion: I wanted to overwrite the file(s) myself because of this most ugly exception here. Could you make this a runtime exception which is aware of a position in a file?

Perhaps a `de.uka.ilkd.key.util.parsing.LocatableException` might very well do the trick. There should be a wrapper in place on an outer level that adds the filename.

Could we also make this "Invalid *unicode* escape character" to remember people that \union is not allows in Java.

Same applies of course to the other file. ...

key.core/src/main/javacc/de/uka/ilkd/key/parser/proofjava/JavaCharStream.java
```diff
@@ -0,0 +1,588 @@
+/* Generated By:JavaCC: Do not edit this line. JavaCharStream.java Version 4.0 */
```
mattulbrich, line 1: I suggest:
```
Originally generated by JavaCC 4.0
This is a modified version of the generated file adapted to the needs within KeY.
```

mattulbrich (changes requested):
As discussed recenty: I like that, but would suggest we do the other item here too.

I wondered: Is it enough to add our own implementations. Would the generator not still produce its versions in the gen folder? Can we reliably guarantee that these versions have priority over the generated java files? Is this build in into gradle's build process?

jwiesler:
I added draft because this does not work. The files get ignored and regenerated. This happens for the already existing Token.java as well, it serves no purpose right now as fas as i see it.

wadoon:
> I added draft because this does not work. The files get ignored and regenerated. This happens for the already existing Token.java as well, it serves no purpose right now as fas as i see it.

Does the copy in [Line 48](https://github.com/KeYProject/key/blob/main/key.core/build.gradle#L48) does not do the trick for `Token.java`? Maybe you need to change it for your classes.

jwiesler:
Oh I didn't see that. However the log output still states that `Token.java` does not exist and is generated...

wadoon:
> Oh I didn't see that. However the log output states that `Token.java` does not exist and is generated...

JavaCC generates these files, and we are overwriting them. The internal mechanism of JavaCC 4.0 to detect these files seems to be broken.

jwiesler:
Did everything you asked for, the warnings are gone as far as I see it and I added the exception so `\union` produces a correct location info in the issue dialog.

Should be ready to be merged once the tests are done.

wadoon (approved):

mattulbrich (approved):

PR #3021, opened 2023-02-05
Use KeYParser.g4 for parsing proof scripts

wadoon:
This PR removes the hand-written parser for proof scripts and uses a few rules in `KeYParser.g4` instead.

Since the start of proof scripts, the KeYParser has changed and became an ANTLR4 grammar. This allows us to easily write a grammar for our proof scripts. (Or just copy the few rules from Sarah and mine proof script parser). This eliminates the handwritten proof script parser with the following benefits:

1. Proof Scripts are a first-class citizen in KeY files. You do not need to put your proof script into a string literal:
   ```
   \proofscript {
       andLeft;
       andRight;  
       rule a ==> b;
   }
   ```
   
   Please note an ambiguity in the grammar: `rule b ==> c` could either be interpreted as a command with one or two arguments: 
   1.  a semi-sequent `b ==>` and a term `c`, or 
   2. a sequent ` b ==> c`.
   3. or a term `b` and a semi-sequent `==> c`.
   
   Use quotes or (better) parentheses to clarify these situations.
  
   The grammar should follow the KeY grammar as best as possible. Backwards-compatibility is tried to preserve.

2. Earlier and better syntax errors, during reading of the KeY file instead of proof script execution. 

4. Better positioning information, as these are generated by using the `ParseContext`s.

5. Proper data structure: no triple of strings is pushed through KeY. 

6. You do not need to put your arguments in quotes. Literals, terms and sequents are parsed. 

### Breaking Change 

* The following is now prohibited: 

   ```
   my-command x;
   ```

   Commands should be valid KeY identifiers, hence they should be free of dashes (aka subtraction operation). 

* Single quotes are reserved for character literals. Please only use `"`. 

### Backward compatibility

Tried to achieve backward-compatibility. 

1. The proof script can still be given as a string: `\proofscript "...";` inside a KeY file. 

2. Term arguments can still be given as a string. The conversion mechanism are still in place to convert the data types in many directions (string to int, string to term, but also term to string, etc.)

   Note there is a difference between `\`f(x)\`` and `"f(x)"`. The first one is parsed directly as a term and can only be passed to term or string arguments (or any other data type for which a converter is registered). The second one `"f(x)"` is parsed as a string. By meta-information on commands, the string is lazily converted to a term if necessary. This conversion is delayed until the command execution. 

   The term `f(x)` might be parsed early with the input file, but the expression is evaluated (translated from parse tree to `Term`) on command-execution time, hence, the use of goal-local variables should be possible.

## "Higher-order proof scripts"

A proof command can receive a block of commands as  an argument. Consider this: 

```
try { auto; }
```

where `try` is a regular proof command. And `{ auto; }` is a block. This constructs replace `tryclose;` but gives you more power on what will be tried. 

This feature arose by the current complicated version of the `AllCommand` (e.g., `onAll hide \`f(x)\``) executes the sub-command specified by the arguments on all goals). In SaG scripts we used `foreach { ... }` to achieve this behavior. 

The extension adapts the command syntax to take a list of sub-commands in curly brackets.

```antlr
proofScriptCommand: 
    name  args  (  '{'   <sub-command>  '}' ';'? | ';' )
    ; 
```

The new syntax for `onAll` is also `onAll { hide \`f(x)\`; }`. Commands receive the code block using the key `#block` inside the map as a parse tree, that can be sent back to the `ScriptEngine` using `#execute(state, statements)`.  The new syntax allows building new commands like 

- `try {<sub>}` -- executes `<sub>` ignoring errors
- `repeat { <sub> }` -- executes `<sub>` as long as there are changes on the sequent. 
- `matching <term> { <sub> }` -- execute on all goals where `<term>` is prescense. 

### Removal of the XML for command documentation. 

We used an XML file to store documentation of proof script commands. This file was deleted, and we use text block literals instead. 

### Remove of `@`-prefix

* [ ] TODO


mattulbrich:
I agree that the current parser is more a proof-of-concept solution than future-proof.

However: With many ideas around proof scripts, which should possibly discuss how this should go now: The script debugger has its ways of communicating to the server, scripts can be in .key files, they were at a point interactively editable, and the plan is to have them in JML code. Do they all go to the same background linear script language or is the base language itself more than a sequence of commands?

@WolframPfeifer @jwiesler

key.core/src/main/antlr4/KeYParser.g4
```diff
@@ -851,10 +851,31 @@ preferences
...
+proofScriptParameter :  ((pname=IDENT EQUALS)? expr=proofScriptExpression);
+proofScriptExpression:
+    boolean_literal
+  | char_literal
+  | integer
+  | floatnum
+  | string_literal
+  | BACKTICK (term | seq) BACKTICK
+  | term
+  | seq;
```
mattulbrich, line 877: I really do not like the ambiguity.
Can we make this IDENTIFIER without ticks and any composed term/seq in ticks?
I would prefer single ticks compared to back ticks. I do not think we need single-character parameters (and if: Use strings ...)

key.core/src/main/java/de/uka/ilkd/key/api/ProofApi.java
mattulbrich, line 1: Documentation would be really helpful. Especially since this is a public API class.
This goes for the class and the methods.

key.core/src/main/java/de/uka/ilkd/key/control/KeYEnvironment.java
```diff
@@ -48,7 +48,7 @@ public class KeYEnvironment<U extends UserInterfaceControl> {
     /**
      * An optional field denoting a script contained in the proof file.
      */
-    private final @Nullable ProofScriptEntry proofScript;
+    private final KeyAst.@Nullable ProofScript proofScript;
```
mattulbrich, line 51: `A.@NonNull B`
Oh dear. This is the Java syntax? Not really nice. Perhaps import ProofScript to avoid this ugly syntax. ...

key.core/src/main/java/de/uka/ilkd/key/macros/scripts/AbstractCommand.java
mattulbrich, line 1: Is this a template or an abstract class? If the former, explain what implementors need to do. If the later, explain what needs to be done to inherit ... No chance to use it w/o doc,

key.core/src/main/java/de/uka/ilkd/key/macros/scripts/AllCommand.java
mattulbrich, line 1: Don't know who wrote this, but definitely needs doc.

key.core/src/main/java/de/uka/ilkd/key/macros/scripts/AllCommand.java
```diff
@@ -3,93 +3,62 @@
...
+    public String getName() {
+        return "onAll";
+    }
 
-        return false;
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String getDocumentation() {
```
mattulbrich, line 58: That xml file does not seem to be the best spot to store documentation.
Can we actually store it here inline, possibly using multiline strings?
That would actually kill two birds with one stone: Developer (javadoc-) documentation and user documentation.

We could also make this an annotation to the Parameters class for each command, and have the parameters themselves also receive that annotation. ...

key.core/src/main/java/de/uka/ilkd/key/macros/scripts/AllCommand.java
mattulbrich: What does this command do? Why can most of the code be remoived?

key.core/src/main/java/de/uka/ilkd/key/macros/scripts/ExprEvaluator.java
mattulbrich, line 1: Please add documentation to the class.

key.core/src/main/java/de/uka/ilkd/key/macros/scripts/LetCommand.java
```diff
@@ -3,48 +3,57 @@
...
 import de.uka.ilkd.key.control.AbstractUserInterfaceControl;
+import de.uka.ilkd.key.logic.Term;
+import de.uka.ilkd.key.macros.scripts.meta.ProofScriptArgument;
 import de.uka.ilkd.key.pp.AbbrevMap;
 
-public class LetCommand extends AbstractCommand<Map<String, String>> {
+/**
+ * The let command lets you introduce abbreviation.
+ * <p>
+ * weigl: add new parameter {@code force} to override bindings.
```
mattulbrich, line 17: This needs more explanation.

key.core/src/main/java/de/uka/ilkd/key/macros/scripts/LetCommand.java
```diff
@@ -3,48 +3,57 @@
...
+
+        boolean force = args.containsKey("force");
+
+        for (Map.Entry<String, Object> entry : args.entrySet()) {
             String key = entry.getKey();
-            if ("#1".equals(key)) {
-                continue;
-            }
-            if ("#literal".equals(key)) {
+            if (key.startsWith("#") && key.equals("force")) {
```
mattulbrich, line 41: This is `if(false)`

mattulbrich (commented):
First round of review. Some feedback. Could not review the whole thing.

mattulbrich:
Reply to me-from-the past:
> However: With many ideas around proof scripts, which should possibly discuss how this should go now: The script debugger has its ways of communicating to the server, scripts can be in .key files, they were at a point interactively editable, and the plan is to have them in JML code. Do they all go to the same background linear script language or is the base language itself more than a sequence of commands?

After implementing a/the script language for JML: This was compiled down to a linear script. So it seems that a list of individual proof commands looks like the right level of abstraction.

key.core/src/main/java/de/uka/ilkd/key/macros/scripts/AllCommand.java
wadoon: I also do not know who wrote it. But `git` blames you on 04.01.2022. 

I guess the command `onAll <cmd>` applies on all open goals the command `<cmd>`.

I did the following: proof commands can also have a script block as the last parameter:

```
onAll <parameter as usual> {   command1; command2 }
```

And the commands can decide what to do with it. This is a rather easy trick, but allows to implement the `repeat`, `foreach`/`onAll` or `try` blocks etc. of Sarah's scripts.

key.core/src/main/java/de/uka/ilkd/key/nparser/KeyAst.java
```diff
@@ -92,17 +93,24 @@ public static class File extends KeyAst<KeYParser.FileContext> {
...
+                KeYParser.ProofScriptEntryContext pctx = ctx.problem().proofScriptEntry();
+
+                KeYParser.ProofScriptContext ps;
+                if (pctx.STRING_LITERAL() != null) {
+                    var ctx = pctx.STRING_LITERAL().getSymbol();
+                    String text = pctx.STRING_LITERAL().getText();
+
+                    // +1 for the removal of the quote.
+                    text = StringUtil.move(StringUtil.trim(text, '"'), ctx.getLine(),
+                        ctx.getCharPositionInLine() + 1);
```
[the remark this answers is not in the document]
wadoon: This preserves the positional information by setting the string to the correct position (line/column) and I added the preservation of the source name.

key.core/src/main/antlr4/KeYParser.g4
wadoon: > I really do not like the ambiguity.

Not my language. 

>  Can we make this IDENTIFIER without ticks and any composed term/seq in ticks?

What do you mean by this IDENTIFIER?

In `macro tryclose`, the `tryclose` is still evaluated as a string and not sent to the expression evaluation. This adds unnecessary complexity. 

> I would prefer single ticks compared to back ticks

* single quotes `'` are often used currently as string quotes in proof scripts
* backticks are only necessary to avoid ambiguous situations: `command p=2` could be a formula or the parameter `p` to be 2.

> (and if: Use strings ...)

No. The advantage of separation is that the parser is eager to give feedback on syntactical correctness. Otherwise, these errors appears only on evaluation of commands.

key.core/src/main/java/de/uka/ilkd/key/api/ProofApi.java
wadoon: This class is old, and will be removed with PR on the Rest API.

key.core/src/main/java/de/uka/ilkd/key/macros/scripts/AllCommand.java
wadoon: One of the problems of Java. IMHO, it seems that we auto-inspection, like in PSDBG. So I would rather drop this method in favor of Javadoc or a website. 

For know, I removed the XML and `DescriptionFacade`.

key.core/src/main/java/de/uka/ilkd/key/control/KeYEnvironment.java
wadoon: Yes, this is the Java syntax, which is required to avoid compiler errors.

key.core/src/main/java/de/uka/ilkd/key/macros/scripts/AllCommand.java
wadoon: https://keyproject.github.io/key-docs/workbench/Proof%20Scripts/commands/

wadoon:
Where is sonarqube?

wadoon:
> After implementing a/the script language for JML: This was compiled down to a linear script. So it seems that a list of individual proof commands looks like the right level of abstraction.

This PR gives the current state a proper syntax in alignment with the KeY grammar. So the list of proof commands are still preserved. Giving the commands the possibility to receive a block of sub-commands arises from your `AllCommand` and allows future extensions w/o needing syntax changes.

We need a quoting symbol for terms for the ambiguities, `"` is used for string values, and `'` is also occupied (do we use it for char literals?). So backticks `\`` are a possible option, but we can also go for parentheses, e.g., `let @a=(f(x))` etc. I do not care.

key.ui/examples/heap/quicksort/sort.key
```diff
@@ -4,4 +4,34 @@
 
 \chooseContract "Quicksort[Quicksort::sort([I,int,int)].JML normal_behavior operation contract.0";
 
-\proofScript "script 'sort.script';"
+\proofScript {
```
mattulbrich, line 7: This example was intended to show that scripts can also be in external files.
It should remain like that to show the feature.

key.ui/examples/heap/quicksort/split.key
```diff
@@ -10,35 +10,34 @@
...
+
+	// first goal: body preserves invariant seqPerm condition
+	macro "simp-upd";
+	rule seqPermFromSwap;
+	rule andRight;
+    // the first of the two conjuncts is easy
+    auto;
+	// the 2nd requires instantiations;
+	instantiate hide var="iv" with=`i_0`;
+	instantiate hide var=jv with=j_0;
```
mattulbrich, line 30: I think this is confusing to users. When to use which quotes. Or not use them. ...

key.core/src/test/resources/de/uka/ilkd/key/macros/scripts/cases.yml
mattulbrich, line 1: This is an unrelated change.
I'd rather like to keep the original structure with many individual files.
(yaml or the original format which did a good job)

key.core/src/main/resources/de/uka/ilkd/key/proof/rules/wdFormulaRules.key
mattulbrich, line 1: important catch, but not related to the pr.

mattulbrich (changes requested):

mattulbrich:
There seem to be accidental artifacts of other PRs in this one.

The old principle was: Every argument is essentially a string. If it is a single identifier the quotes can be omitted.
Why not keep this convention? I'd vote for this.  (It's heavily inspired by "isar.")

If dropping this in favour of type safety, then probably this should be handled more strictly in order to not add confusion instead of removing it.

key.core/src/main/antlr4/KeYParser.g4
```diff
@@ -851,10 +851,32 @@ preferences
...
-  PROOFSCRIPT ps = STRING_LITERAL
+  PROOFSCRIPT
+    ( STRING_LITERAL
+    | LBRACE proofScript RBRACE
+    )
 ;
+proofScriptEOF: proofScript EOF;
+proofScript: proofScriptCommand+;
+proofScriptCommand: AT? cmd=IDENT proofScriptParameters?
+	( LBRACE sub=proofScript RBRACE SEMI?
```
mattulbrich, line 864: Is this new?

wadoon: You mean that that `onAll { tryclose; }` would be now allowed yes. I allowed that proof script can receive a block of commands as an argument. Describe above in the "Additions to proof scripts". 

Better then to implement `onAll` my quirking around.

mattulbrich:
This PR seems to change the infrastructure quite considerably. It touches > 260 files, some of them rather far from the topic at hand.

I do like the idea to reuse the KeY parser instead of the handmade parser.
Still, there is nothing wrong (I think) with the character bases handmade parser, but the ANTLR machine of course allows for more flexibility and cleaner code.

I was not able to understand the higher order feature, but that looked nice indeed.

I'd argue against the (voluntary) strong typing. The design of the language is untyped. Everything is a string. I would like to stick to this on the KeY level. It is a different story on the JML level.

Can we break this down into manageable units (by cherrypicking aspects e.g.)?

The script proof engine can an overhaul if/when the JML scripts are on master-track.

mattulbrich:
Consider force pushing this to [`weigl/psg4-cleaned`](https://github.com/KeYProject/key/tree/weigl/psg4-cleaned) where most non-related aspects have been removed by rebasing the branch.

mattulbrich:
Afterwards consider https://github.com/KeYProject/key/tree/psg4-cleaned-continued where a few ideas from other branches regarding the non-JML bit of proof scripts have come together.

mattulbrich:
The ambiguity of the parser still is really a problem. `command a = b ==>` can be at least three different calls. Should ANTLR not complain? Parser generators usually have guarantees regarding ambiguities. ...

wadoon:
I'll rebase this PR. Splitting into multiple PRs should not only bring complexity.

mattulbrich:
Just read the commit message "Introduce ScriptCommandAst to keep the ProofEngine separated from ANT…"

I'd propose to limit this PR to its original purpose and merge it into master.
Going on from that point with ScriptCommandAPI etc is then up to further development in a next PR.
This would also allow to add new feature in other PRs.

mattulbrich:
You asked me for a review: What is the solution to the ambiguity problem again?

key.ui/examples/heap/quicksort/sort.key
wadoon: Reverted

key.core/src/main/resources/de/uka/ilkd/key/proof/rules/wdFormulaRules.key
wadoon: I am also not sure, when this came to this branch, but I thought we already fixed that on `main`.

key.core/src/main/antlr4/KeYParser.g4
```diff
@@ -852,10 +852,32 @@ preferences
...
+proofScriptParameter :  ((pname=proofScriptParameterName (COLON|EQUALS))? expr=proofScriptExpression);
+proofScriptParameterName: AT? IDENT; // someone thought, that the let-command parameters should have a leading "@"
+proofScriptExpression:
+    boolean_literal
+  | char_literal
+  | integer
+  | floatnum
+  | string_literal
+  | LPAREN (term | seq) RPAREN
+  | term
```
mattulbrich, line 878: Well this still has this ambiguity of "a=b" right?

I thought this should be a identifier, or perhaps a qualified identifier, right?

mattulbrich: Or arguably even baseterm (without the infix operations)

wadoon:
```
  | abbreviation
  | accessterm
  | literals
```

Better? This prohibits also `macro split-prop;`.

mattulbrich (approved):

PR #3022, opened 2023-02-05
Exploiting JML Names in the Proof Tree

wadoon:
This PR adds the newly introduced JML names for entities to KeY. Such labels are exploited inside the KeY's proof tree as branching labels. And may later become accessible in proof scripts.

This MR brings following changes:

* JML entities have names:
  ```java
  //@ invariant NAME: true;
  //@ ensures<h1> NAME: true;
  //@ behavior NAME: requires NAME1: true; 
  ```

* `lbl` function is supported, e.g., `\lbl(NAME, x != y)`

* A new term label `name(XXX)` is added, where "XXX" holds the name given in the JML specification. 

* Terms that are created by the JMLParser/Translator receive a this term label if the entity has a name or a `\lbl` was given. 

* Change to the KeY parser: Goal templates can have computed labels. 

  Currently, the label of a goal can be a schema variable or static string. With this MR you are allowed to define labels which are computed by a Java interface `BranchNamingFunction`. 
  A taclet exploits this in the following way:
  
  ```
  andRight {
      \find (==> b & c)
      "\nameLabelOf(b)":\replacewith(==> b);
      "\nameLabelOf(c)":\replacewith(==> c)
      \heuristics(beta)
  };
  ```
  
  New branching naming functions can be added by an interface. The `\nameLabelOf` tries to find a `f«name(X)»` term label, and returns `X`. It fallbacks to the origin term label and then to the toString representation.

### TODO 

* [x] Better branch names for `origin` label
* [x] Test function for the introduction of abbreviations on sequent based on name label.
* [x] Discuss impact on proof scripts. Extend proofs scripts with abbreviation variable? `@post_abc`?
* [x] `andRightX` and `orLeftX` per Built-In rule?
---

### KeY example (working)

```
\sorts { s; }
\predicates {	p; 	q; }
\problem {
  (p<<name("A")>> -> q<<name("B")>>) -> !q<<name("C")>> -> !p<<name("D")>>
}
```

image

New rules `andRightX` (for $x \in 3..6$) allows a more comprehensible proof split. 

image

### JML

Consider following examples: 

```java
class Test {
    //@ public invariant MY_SUPER_INVARIANT: CONST == 42;

    public final int CONST = 42;
    /*@
    requires Z: this != null;
    ensures A: \result == 42;
    ensures B: \result >= 0;
    ensures C: \result != 0;
    */
    public int foo() {return CONST;}
}
```

This results into the following proof tree. At places where no name label is available we ask the origin label for a string representation. The origin label needs further improvement e.g., line numbers :exclamation: 

image

### Sum And Max
```
class SumAndMax {

    int sum;
    int max;

    /*@ normal_behaviour
      @   requires R1: (\forall int i; 0 <= i && i < a.length; 0 <= a[i]);
      @   assignable sum, max;
      @   ensures E1: (\forall int i; 0 <= i && i < a.length; a[i] <= max);
      @   ensures E2: (a.length > 0
      @           ==> (\exists int i; 0 <= i && i < a.length; max == a[i]));
      @   ensures E3: sum == (\sum int i; 0 <= i && i < a.length; a[i]);
      @   ensures E4: sum <= a.length * max;
      @*/
    void sumAndMax(int[] a) {
        sum = 0;
        max = 0;
        int k = 0;

        /*@ loop_invariant I1: 0 <= k && k <= a.length;
          @ loop_invariant I2: (\forall int i; 0 <= i && i < k; a[i] <= max);
          @ loop_invariant I3: (k == 0 ==> max == 0);
          @ loop_invariant I4: (k > 0 ==> (\exists int i; 0 <= i && i < k; max == a[i]));
          @ loop_invariant I5: sum == (\sum int i; 0 <= i && i< k; a[i]);
          @ loop_invariant I6: sum <= k * max;
          @
          @  assignable sum, max;
          @  decreases a.length - k;
          @*/
        while(k < a.length) {
            if(max < a[k]) {
                max = a[k];
            }
            sum += a[k];
            k++;
        }
    }
}
```

### Introduction of Abbreviations for Named Terms 

`Menu: Proof-> Introduce abbreviation for named formulas`

Before:

image

After actions is applied:

image


wadoon:
@mattulbrich @unp1 @WolframPfeifer @jwiesler @flo2702 

Help required :exclamation: This PR is feature complete, but some proofs do not close anymore. It seems that term label indeed affect the calculus. 

You can see this on the `SumAndMax` example. The example given in the "Example Chooser" is still provable, whereas `key.ui/examples/dynamicBranchLabels/SumAndMax` is not provable. The first difference in the proof tree I noted is a different cut formula. 

Does anyone is deeper in the topic or can look at it?

wadoon:
Still struggling about the differences in the proofs.

image

Added a GUI for finding differences in proof trees and sequents.

wadoon:
SumAndMax runs. It was a missing setting. 

Proof tree look as follows:

image

@WolframPfeifer 

Is it possible to achieve to make cut branch names "If x_5 is TRUE" more useful? Could we not add a term label (dynamical added) to `x_5` that origin is "if-cond is line X"? And make a term label "If if-cond:X (x_5) is TRUE"?

jwiesler:
Couldn't we make branch labels just any complex type instead of string and let the GUI decide what to display? This would allow more control over what is displayed, how, and where.
This type could contain origin, the cut term in this case and more.

An example that comes to my mind instantly is pretty printing branch labels in the tooltip.

wadoon:
> Couldn't we make branch labels just any complex type instead of string and let the GUI decide what to display? 

Possible. But I see negative impact on computing time and memory, and also not the use case of different renderers in the UI. 

In principle, we can today recompute all labels in the current situation, also with the branching name label.

mattulbrich:
Should that perhaps be made WIP since there are still open todos?

wadoon:
> Should that perhaps be made WIP since there are still open todos?

No. It is reviewable.

wadoon:
@WolframPfeifer it is green. Auto-merge error.

WolframPfeifer (commented):
While I like the general idea, I have some complaints with this PR:
1. It is unclear (undocumented), for which formulas/terms in JML you can define names. I know that for accessible clauses, it is not possible to name them (ambiguous syntax). Are there further restrictions?
2. To me, it is still unclear what the labels are attached to, for example for assignable clauses: The formulas that are generated from the clause? The (JavaDL representation of the) terms generated from the clause? Is it different whether you use the contract or prove the contract?
3. The documentation should be better, e.g.:
    - There are undocumented classes (MakeNamedFormulaToAbbrevAction, ProofDifferenceView, ...).
    - When it is documented, the documentation is very sparse (e.g.: In CompositionalBranchNamingFunction, how are the labels combined?)
    - Some contain pseudo JavaDoc (e.g. TextualJMLConstruct.createTermLabel()).
4. While it seems to be an easy solution at first, I think that using the term label mechanism for this introduces many problems:
    - When you (accidentally) assign the same term labels to different formulas, the proof tree looks strange, the "make abbreviation" feature does not work.
    - Term labels can "disappear" suddenly when the same formula (without the label) is already present on the sequent.
    - You can assign multiple name labels to a single term, which is also very strange.

WolframPfeifer:
I already did a review recently and have not yet received any answers to my questions from back then. I am still missing a clear concept in this PR, it seems really arbitrary at some points (questions 1 and 2 in the list). The most important point is that I think that there are serious problems with using the term labels (point 4). However, the documentation is better now, thanks.

WolframPfeifer (commented):
For me, the generation of name labels seems to work only partially: 
```java
class Labels {
    int f = 7; 
    //@ invariant inv1: f == 7;
    /*@ requires pre1: x == 5;
      @ ensures post1: \result == 42;
      @ ensures post2: \result == \lbl(prod, 6 * 7);
      @*/
    int m(int x) {
        return 42;
    }
}
```
image
Only the formula with lbl got a label as expected. Am I doing something wrong here?

mattulbrich:
I mark this "Work in progress" since a number of checks fail. Please undo if/when ready to review.

wadoon:
This is a typical example that KeY development is dysfunctional. 

It was green, and nobody reviewed it, after repeatedly merging the main branch into it, it became red.


I do not see any reason to invest further time into it, only to see the branch rot away until  the next review.

Issue #1723, opened 2022-12-23
Bug in rule application for floating point comparsion


wadoon:
This issue was created at [git.key-project.org](https://git.key-project.org/key/key/-/issues/1723) where the discussions are preserved.
    
----
    
 

## Description

When trying to apply the `lesser-or-equal than distinction` rule to a floating point comparison on the `master` branch, the rule is not applied and an exception is thrown that is only visible in the terminal log.

## Reproducible
  
always

### Steps to reproduce 

1. Open KeY
2. Load the following program:  
```java
class Bug {

    private float rPos;

    /*@ normal_behavior 
     @ requires (0.0 <= rPos & rPos <= 100.0);
     @ ensures (rPos >= 25.0);
     @ assignable rPos;
     */
    public void runController(){
        if (this.rPos <= 51.0) {
            this.rPos = 100.0;
        } else {
            this.rPos -= 1.0;
        }
    }
}

```
3. Run auto mode; this should result in the [Bug-incomplete.proof](/uploads/6456a231ea9d39dcec72eb28605fa0c3/Bug-incomplete.proof).
4. Click on `x=x_1<=51.0;` and choose `lesser-or-equal than distinction`
5. The rule will not be applied, there will be no UI error message, you will however find an error message in the terminal

### Steps to reproduce 
This depends a little bit on the current state of floating points in KeY, I think there's two options:
- Do error handling and show a UI visible error saying that this operation is currently unsupported
- Apply a distinction rule for floating points

### Additional information

**Terminal log of error:**
```
Exception in thread "AWT-EventQueue-0" de.uka.ilkd.key.logic.TermCreationException: Building a term failed. Normally there is an arity mismatch or one of the subterms' sorts is not compatible (e.g. like the '2' in "true & 2")
The top level operator was leq(Sort: Formula); its expected arg sorts were:
1.) sort: int, sort hash: 1944308833
2.) sort: int, sort hash: 1944308833

The subterms were:
1.) x_1<<origin(<none> (implicit)) ([ensures @ file Bug.java @ line 8, ensures (implicit), assignable (implicit)])>>(sort: float, sort hash: 1126995741)
2.) DFP(0(2(1(1(2(6(5(1(2(9(2(4(4(7(3(2(3(6(4(#))))))))))))))))))))(sort: double, sort hash: 839856028)

	at de.uka.ilkd.key.logic.op.AbstractSortedOperator.additionalValidTopLevel(AbstractSortedOperator.java:120)
[...]

PR #3023, opened 2023-02-05
Solving the FP completeness issues in #1723

mattulbrich:
## Related Issue

This pull request answers to issue #1723.

## Intended Change

Floats and doubles are currently incorrectly insufficiently handled when they appear in
one expression.

Casts need to be added and casts need to be dealt with. ...

- [x] introducing casts where needed
- [ ] adding rules / SMT support for such casts.

## Type of pull request

- [x] There are changes to the taclet rule base
- [ ] Other: 

## Ensuring quality

- [x] I made sure that introduced/changed code has been well documented (javadoc).
- [x] I have tested the feature as follows: small 2-liners
- [ ] I have checked that runtime performance has not deteriorated  

It's still  W I P.

@samysweb 

The contributions within this pull request will be licensed under GPL2-or-later wihtin KeY.


samysweb:
I just did some testing and the floating point balancing seems to work relatively reliably for binary operations.
I currently still see two issues with the current state:

### Edge case for binary expressions
There seems to be one edge case due to the replacement strategy of `ProgramElementReplacer` which always substitutes the first occurrence.
The bug shows up when analyzing the following program:
```
/*@ public normal_behavior
  @ requires x1 > 0.0f && x2 > 0.0f;
  @ ensures \result>0.0;
  @*/
public double myFunction05(float x1, float x2) {
    double squared1 = ((double) x1)+x1;
    return squared1;
}
```
At some point within the proof we get to the expression `squared1 = ((double)_x1)+_x1`
At this point we can either apply the `compound_addition_1` taclet (then everything works out fine) or we can apply the `cast` taclet which (applied repeatedly) results in:
- `((double)(double)_x1)+_x1`
- `((double)(double)(double)_x1)+_x1`
- ...

This is because `ProgramElementReplacer` finds `_x1` in the left child of `+` first and replaces it there.
I'm not sure about the consequences of this:
1. It seems that this introduces some non-determinism in the symbolic execution meaning if we always apply the wrong rule we could get stuck
2. From my understanding of how replacements work this does not seem easily fixable without reimplementing a position based version of `ProgramElementReplacer`

How to do position based replacement?

### Assignments/Returns
The current rules did not yet take this case into account, to this end I began drafting rules in #3062 (On my own fork since I don't have writing rights on this branch).
The added taclets allow for the treatment of most widening assignments, however there are still some limitations:
- [ ] Auto mode currently breaks on the first appearance of such an assignment, as the `assignment` taclet seems to take precedence but fails upon application due to type incompatibilities
- [ ] If I understand this correctly, a binary operation between two integral types (e.g. two ints) is not a simple expression and thus the rules added in #3062 currently do not apply. Currently this means the addition rule for adding 2 ints is applied and fails with an exception due to a type mismatch. I'm not sure if this can be fixed without replicating all `assignmentMultiplicationInt`,`assignmentAdditionInt` etc. taclets for float and long

**Notes from our conversation just now:**
- Maybe add type sanity checks to all currently existing assignment rules and only allow assignment if same sort or subtype (currently this is only discovered upon application and prevented through an exception which aborts auto mode)
- If we do this then new rules are needed:
  - Binary Operation between two simple expression with unbalanced types: Adjust `unbalanced_float_expression` from this branch
  - Assignment `d = se0 + se1` where `se0` and `se1` have same type but `d` differs: `e=se0+se1;d=e` where `e` has type of `se0`
  - `d=e` with `e` simple expression: type cast

samysweb:
Huh interesting, so what really happens when you click GitHub's innocently looking "rebase" button is a force-push...

mattulbrich:
> Huh interesting, so what really happens when you click GitHub's innocently looking "rebase" button is a force-push...

Ok, anything lost or was it a fast-forward with a more aggressive name?

samysweb:
I don't think anything should have gotten lost.
Probably just a more aggressive name because the `main` commits are inserted before the commits on this branch.

On that note: Unfortunately, I just realized #3027 is not merged yet

samysweb:
I'm saving the runtimes of the last test run here so that we can compare runtimes once my pull request (with lots of additional `\varcond` statements) is merged into this branch:

Job | Run time |
-- | -- |
optional-tests | 32m 35s
integration-tests (testProveRules, ubuntu-latest, 11) | 16m 5s
integration-tests (testRunAllFunProofs, ubuntu-latest, 11) | 1h 27m 29s
integration-tests (testRunAllInfProofs, ubuntu-latest, 11) | 56m 47s
unit-tests (ubuntu-latest, 11) | 21m 42s
unit-tests (ubuntu-latest, 17) | 20m 44s
**Overall** | **3h 55m 22s**

samysweb:
Looks like all tests are passing.
Comparison of runtimes:
Job | Run time old | Runtime after merge
-- | -- | --
optional-tests | 32m 35s | --
integration-tests (testProveRules, ubuntu-latest, 11) | 16m 5s | **11m 46s**
integration-tests (testRunAllFunProofs, ubuntu-latest, 11) | **1h 27m 29s** | 1h 46m 50s
integration-tests (testRunAllInfProofs, ubuntu-latest, 11) | 56m 47s | **45m 43s**
unit-tests (ubuntu-latest, 11) | 21m 42s | **17m 24s**
unit-tests (ubuntu-latest, 17) | **20m 44s** | 25m 18s

Left to do:
- [ ] Check once more that rules actually do the job
- [ ] Check that SMT translation is available (see above)
- [ ] Is slower run of `testRunAllFunProofs` due to changes in int rules?

samysweb:
Alternative to variable conditions: Create a new schema variable `IntegralVariable` and then use `#integralLoc` (code for Schema Variable class provided by Richard Bubel).

Drodt:
This PR is now up-to-date again. What is the status here? Anything that needs help?

PR #3031, opened 2023-02-07
Overhaul of the Configuration

wadoon:
The KeY settings are in a *hysterical* and *historic* shape. 

This PR tries to homogenize the settings to Java POJOs using the JavaBeans standard and mechanism. The measurement include:

* Introduction of Getter/Setter
* Removal of `SettingsListener` in favour of `PropertyChangeListener`
  (You can listen on individual configuration changes.)

**Most of the changes** results from dependencies to the `SettingsListener`, or the change of field-access to getter and setter.

A next PR step will change the settings language in KeY files to a custom type-safe backward 
compatible language.


key.ui/src/main/java/de/uka/ilkd/key/gui/sourceview/JavaDocument.java
```diff
@@ -251,15 +251,15 @@ private enum CommentState {
...
+    private final transient PropertyChangeListener listener = e -> updateStyles();
 
     /**
      * Creates a new JavaDocument and sets the syntax highlighting styles (as in eclipse default
      * settings).
      */
     public JavaDocument() {
         updateStyles();
-        ColorSettings.getInstance().addSettingsListener(listener);
+        ColorSettings.getInstance().addPropertyChangeListener(e -> updateStyles());
```
jwiesler, line 262: Shouldn't this be `listener`? Otherwise, `dispose` removes nothing.

key.ui/src/main/java/de/uka/ilkd/key/gui/MainWindow.java
```diff
@@ -1707,12 +1710,9 @@ public synchronized void autoModeStopped(ProofEvent e) {
...
         }
 
-        /**
-         * invoked when the strategy of a proof has been changed
-         */
         @Override
-        public synchronized void settingsChanged(EventObject e) {
-            if (proof.getSettings().getStrategySettings() == e.getSource()) {
+        public void propertyChange(PropertyChangeEvent evt) {
+            if (proof.getSettings().getStrategySettings() == evt.getSource()) {
```
jwiesler, line 1715: `proof` is accessed and `synchronized` was removed. Is this right? See the Github Annotation above.

key.core/src/main/java/de/uka/ilkd/key/settings/ChoiceSettings.java
```diff
@@ -20,57 +12,56 @@
...
-    @SuppressWarnings("unchecked")
-    public HashMap<String, Set<String>> getChoices() {
-        return (HashMap<String, Set<String>>) category2Choices.clone();
+    public Map<String, Set<String>> getChoices() {
+        return Collections.unmodifiableMap(category2Choices);
     }
 
-
     /**
      * returns a copy of the HashMap that maps categories to their currently selected choices.
```
jwiesler, line 59: +1 for using unmodifiable instead of cloning, however, the comments are wrong now

key.core/src/main/java/de/uka/ilkd/key/settings/ChoiceSettings.java
```diff
@@ -98,64 +94,46 @@ private static ImmutableSet<Choice> choiceMap2choiceSet(HashMap<String, String>
...
-                category2Choices = c2C;
-                fireSettingsChanged();
+            var tmp = new HashMap<>(category2Choices);
+            if (!remove) {
+                tmp.putAll(c2C);
+                setChoiceCategories(tmp);
             } else {
-                category2Choices.putAll(c2C);
-                ProofSettings.DEFAULT_SETTINGS.saveSettings();
+                setChoiceCategories(c2C);
```
jwiesler, line 111: Correct me if I'm wrong but either the comment or the code is wrong: `choiceNS` might contain more elements than `category2Choices`, so removing all from `category2Choices` if they are not contained in `choiceNS` is not equal to `choiceNS`.

E.g. `category2Choices = [a: 1, b: 2]`, `choiceNS = [a: 1, c: 3, d: 4]` should imo lead to `category2Choices = [a: 1]` and currently leads to `category2Choices = [a: 1, c: 3, d: 4]`

key.ui/src/main/java/de/uka/ilkd/key/gui/smt/TacletTranslationSelection.java
```diff
@@ -16,6 +17,120 @@
 import de.uka.ilkd.key.taclettranslation.assumptions.SupportedTaclets.TreeItem;
 import de.uka.ilkd.key.taclettranslation.assumptions.SupportedTaclets.TreeItem.SelectionMode;
 
+interface InfoListener {
+    void eventShowInfo(TreeItem item, TreeNode node);
+}
+
+
+public class TacletTranslationSelection {
```
jwiesler, line 25: What does this do? Maybe a file you added by mistake?

jwiesler (approved):

key.ui/src/main/java/de/uka/ilkd/key/gui/smt/TacletTranslationSelection.java
wadoon: I do not know. Git blame goes back to the 2000s. Maybe I just touched it to adapt this to the new interfaces. 

I delete it.

key.core/src/main/java/de/uka/ilkd/key/settings/ChoiceSettings.java
wadoon: I would expect the latter result you gave. And key `b` is only removed if the parameter is set to true. 

It updates the internal data structure by the found choices in the KeY files. 

BTW. This was refactored from this:

```
public void updateChoices(Namespace<Choice> choiceNS, boolean remove){
	Iterator<Choice> it = choiceNS.allElements().iterator();
	HashMap<String,Set<String>> c2C = new LinkedHashMap<String, Set<String>>();
	Choice c;
	Set<String> soc;
	while(it.hasNext()){
	    c=(Choice)it.next();
	    if(c2C.containsKey(c.category())){
		soc=c2C.get(c.category());
		soc.add(c.name().toString());
		c2C.put(c.category(),soc);
	    }else{
		soc = new LinkedHashSet<String>();
		soc.add(c.name().toString());
		c2C.put(c.category(),soc);
	    }
	}
	if(!c2C.equals(category2Choices)){
	    if(remove){
		category2Choices = c2C;
		fireSettingsChanged();
	    }else{
		category2Choices.putAll(c2C);
		ProofSettings.DEFAULT_SETTINGS.saveSettings();
	    }
	}
	for (final String s : getDefaultChoices().keySet()) {
	    if(category2Choices.containsKey(s)){
		if(!category2Choices.get(s).
		   contains(category2Default.get(s))){
		    category2Default.put(s,category2Choices.get(s).
					 iterator().next());
		    fireSettingsChanged();
		}
	    }else{
		category2Default.remove(s);
		fireSettingsChanged();
	    }
	}
    }

```

key.core/src/main/java/de/uka/ilkd/key/settings/ChoiceSettings.java
```diff
@@ -163,21 +153,19 @@ public void readSettings(Properties props) {
...
+     * the form &lt; key &gt; = &lt; value &gt; (,&lt;
      * value &gt;)* are allowed.
      *
-     * @param props the Properties object where to write the settings as (key, value) pair
+     ** @param props the Properties object where to write the
+     *        settings as (key, value) pair
      */
+    @Override
     public void writeSettings(Properties props) {
-        StringBuilder choiceSequence = new StringBuilder();
```
jwiesler, line 172: @wadoon you deleted an earlier change from me here by the way...

PR #3032, opened 2023-02-07
Overloaded Operators for KeY lang

wadoon:
This MR brings overloaded operators to KeY lang:

1. On function and predicate definitions you can additional add prefix, infix, and postfix notations.
   ```
   \functions {
     int add(int, int) \infix(+);
   }
   ```

   These information are stored inside the `Function` object. 

2. These information also impacts the pretty printing.
 
3. Operators are now a sequence of symbols `[-+*/^<>=|]`. Their first character determines the precedence. Following are legal expressions:

   ```
     a ++ b + c  = ( a ++ b ) + c
     a <|> b <= c 
     a +- b != a + -b
   ```
  This trick avoids an "operator precedence parser", which would also be nice.


mattulbrich:
> This MR brings overloaded operators to KeY lang:

Does this include overloading in the sense that several signatures can share the same symbol (+ for floats and for integers e.g.). That would really be important.

key/key.core/src/main/antlr4/KeYParser.g4
```diff
@@ -228,18 +230,47 @@ func_decl
...
+
+adt_constructor
+:
+    doc=DOC_COMMENT? funcpred_name                                      #adt_constructor_base
+  | doc=DOC_COMMENT? funcpred_name LPAREN sortId (COMMA sortId)* RPAREN #adt_constructor_recursion
+;
+
+functionMetaData
+:
+    INFIX LPAREN op=(PLUS|STAR|SLASH|MINUS|EXP|PERCENT|LESS|LESSEQUAL|GREATER|GREATEREQUAL|LGUILLEMETS) RPAREN
```
mattulbrich, line 259: This seams to deviate from the PR description

key/key.core/src/main/antlr4/KeYParser.g4
```diff
@@ -228,18 +230,47 @@ func_decl
...
+  functionMetaData?
   SEMI
 ;
 
+/**
+ \datatypes {
+    List = Nil | Cons(any, List);
+ }
+ */
+adt_decl
```
mattulbrich, line 242: This MR brings overloaded operators to KeY lang:

PR #3039, opened 2023-02-17
Enabling Java 17 for unit-tests

wadoon:
This PR enables the Java 17 for the `unit-tests` check in the Github workflow. 

A reflection test case in recoder's test suite is not compatible anymore, and was disabled. 
Impact on operation in KeY is unknown.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -20,7 +20,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest ] # currently buggy, windows-latest]
-        java: [11]
+        java: [11, 17]
     continue-on-error: true
     runs-on: ${{ matrix.os }}
     env:
--- a/recoder/src/test/java/recoder/testsuite/fixedbugs/FixedBugs.java
+++ b/recoder/src/test/java/recoder/testsuite/fixedbugs/FixedBugs.java
@@ -82,7 +82,8 @@ public class FixedBugs {
     @Test
     public void testBasicReflectionImport() {
         // make sure non-public fields can be read...
-        ReflectionImport.getClassFile("java.lang.String");
+        // weigl, 2023-03-11, disabled, not working under Java 17
+        // ReflectionImport.getClassFile("java.lang.String");
     }
 
     @Test


WolframPfeifer (approved):

WolframPfeifer:
> Impact on operation in KeY is unknown.
In my quick test with Java 17, I could not find a difference in the behaviour of KeY (GUI version). Also, the tests are successful. So I would hope that the test case tested some feature of recoder that is not used by KeY. I think we can risk merge and also migration to Java 17 in general.

wadoon:
> > Impact on operation in KeY is unknown.
> > In my quick test with Java 17, I could not find a difference in the behaviour of KeY (GUI version). Also, the tests are successful. So I would hope that the test case tested some feature of recoder that is not used by KeY. I think we can risk merge and also migration to Java 17 in general.

Note, that this release does not increment the "release version"! It just enable the unit-tests under Java 17 additionally. 


If you want Java 17 enabled, incl. its language feature, we have to increase the number in `/build.gradle`.

jwiesler:
> If you want Java 17 enabled, incl. its language feature
Is there anything against doing this? Java 18 would be great as well, value based classes and more...

WolframPfeifer:
> Note, that this release does not increment the "release version"!

Yes, I know :slightly_smiling_face:. My argument just was that the disabled test is probably irrelevant for KeY, as it seems to test some feature of Recoder that is not in use by KeY. I played around with KeY with Java 17 (enabled in `build.gradle`), to see if something breaks. Since this does not seem to be the case, I think we can (in addition to this PR) "risk" the upgrade. From my point of view, nothing speaks against it.

mattulbrich:
I think that entire bytecode support within recorder is not (or hardly) used within KeY.

I seem to remember that there is a way to load ".jar" files as classpaths. We should check that this works as well (or badly) as before.

The test case could probably also be fixed by not using java.lang.String, but a class that is under our jurisdiction. However, dropping it is perfectly fine from my point of view.

mattulbrich:
> Java 18 would be great as well, value based classes and more...

Let's wait for the LTS this year.

PR #3040, opened 2023-02-22
Register modifier when parsing JML spec

### LINKED ISSUE #1722: Jml Modifiers of spec cases are offset by one contract
This issue was created at [git.key-project.org](https://git.key-project.org/key/key/-/issues/1722) where the discussions are preserved.
    
----
    
## Description
The modifiers passed to JML spec cases in `TextualTranslator` are offset by one.

Example:
```
/*@ public normal_behaviour
  @ requires true;
  @ also
  @ exceptional_behaviour
  @ requires true;
  (at)*/
public void test() {}
```

Parsed contracts (same when leaving out `also`):
* `normal_behaviour` without modifiers
* `exceptional_behaviour` with `public` modifier

The modifiers can also be swapped, leading to both contracts having no modifiers.

I don't have enough insight into the class to make this change but `visitSpec_case` can be a starting point.

### Steps to reproduce 
Debug. Those modifiers are not currently in use as far as I can see it, so no clue what to write here.
However, when actually using those modifiers (like I want to), this is very important.

### Additional information
* Commit: Newest master
    
----
    
Information: 
* created_at: 2022-10-20T16:30:00.932Z
* updated_at: 2022-10-20T16:41:24.637Z
* milestone: 
* user_notes_count: 2

FliegendeWurst:
Fixes #1722. Previously, the modifier was read after construction of `TextualJMLSpecCase`.

### DIFF
--- a/key.core/src/main/antlr4/JmlParser.g4
+++ b/key.core/src/main/antlr4/JmlParser.g4
@@ -50,7 +50,7 @@ class_invariant: INVARIANT expression SEMI_TOPLEVEL;
 method_specification: (also_keyword)* spec_case ((also_keyword)+ spec_case)*;
 also_keyword: (ALSO | FOR_EXAMPLE | IMPLIES_THAT);
 spec_case:
-  (modifier)?
+  (modifiers)?
   behavior=(BEHAVIOR | NORMAL_BEHAVIOR | MODEL_BEHAVIOUR | EXCEPTIONAL_BEHAVIOUR
            | BREAK_BEHAVIOR | CONTINUE_BEHAVIOR | RETURN_BEHAVIOR )?
   spec_body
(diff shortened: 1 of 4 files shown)


WolframPfeifer (approved):

WolframPfeifer:
> Great. Could you extend the test to test for multiple contracts in a row? E.g.:
> 
> ```
> /*@ public abstract final normal_behaviour
>   @ requires true;
>   @ private static exceptional_behaviour
>   @ requires false;
>   @*/
> ```

Is this even allowed in (KeY-)JML? In my opinion, multiple contracts should be separated by "also". However, I know that this is a discussion topic in the JML group and that it has been handled differently in the past.

jwiesler:
I don't know, but the example in #1722 used it.

However, this PR should test multiple contracts before closing #1722, with `also` or without doesn't matter to me.

WolframPfeifer:
@jwiesler I agree that we should have more unit tests here, in particular some that test the "multiple contracts" situation.

@FliegendeWurst Could you please add one or two test cases with multiple contracts and modifiers?

PR #3042, opened 2023-02-22
Fix CVC4 solver socket handling of newlines

### LINKED ISSUE #1656: Solver communication with CVC4 not working
This issue was created at [git.key-project.org](https://git.key-project.org/key/key/-/issues/1656) where the discussions are preserved.
    
----
    
## Description

The CVC4Socket for SMT communication searches for "\n" + "unsat" (or "sat"/"unknown") in incoming messages, but in my case the message looks like that:
grafik
This leads to the result not being recognized and every solver process I start with CVC4 hence fails.

## Reproducible
  
Always.

### Steps to reproduce 

Start a solver process with CVC4 that returns a result before the timeout is reached.
KeY won't recognize the result.

### Additional information

Windows 10, CVC4 version 1.8 from https://cvc4.cs.stanford.edu/downloads/builds/win64-opt/.

* Commit: bcd46a9f
    
----
    
Information: 
* created_at: 2021-12-28T12:33:20.632Z
* updated_at: 2021-12-28T12:33:53.044Z
* milestone: 
* user_notes_count: 0

FliegendeWurst:
Fixes #1656


jwiesler:
Could this be abused by naming things "sat"? Is the returned message always only "sat"?

FliegendeWurst:
> Could this be abused by naming things "sat"? Is the returned message always only "sat"?

No, CVC4 only prints `... > ... > ... > success` for each added assertion.

wadoon (approved):

wadoon:
> Could this be abused by naming things "sat"? Is the returned message always only "sat"?

This is the protocol of sat solvers. On `(check-sat)` the answer is either symbol `sat`, `unsat` or `unknown`. If the SAT solver want to print out something, it needs to use quotes.

PR #3045, opened 2023-02-27
Fix #3035

### LINKED ISSUE #3035: KeY silently removes parts of unsupported \sum expressions to make \bsum's
## Description

KeY only supports \sum expressions of the form `\sum T x; n <= x && x < m; t` and converts them to bsum terms. (see KeY book chap 8.1.2.3)

When writing a specification with a \sum expression that is not supported by KeY (because it has 3 conditions) the additional parts of the sum-condition get silently removed to make it a valid (but different) bsum expression. 

For example `( \sum int x; 0 <= x && x < values.length && values[x]>=0; values[x]) ` gets treated the same as `( \sum int x; 0 <= x && x < values.length; values[x] )`

## Reproducible
  
always

### Steps to reproduce 

1. Load this example program:
```java
class Example20230214 {
    int[] values;

    /*@ normal_behaviour
      @
      @ requires true;
      @
      @ ensures ( \sum int x; 0 <= x && x < values.length;           values[x] ) == 
      @         ( \sum int x; 0 <= x && x < values.length && x%2==0; values[x] );
      @
      @ assignable \nothing;
      @*/
    void sumExample() {  }
}
```

2. Observe that both \sum expressions get translated to the same term on the sequent:  `javaCastInt(bsum{int x;}(0, self.values.length, self.values[x]))`

3.  The AutoPilot closes the proof (but imo shouldn't)

### Expected behaviour

Generate a raw `sum` term (similar to \sum expressions that are totally different from the bsum form) - or throw an error.

### Additional information

Reproduced on the current ~~master~~ main branch ( 87ceb1cd6f965beeb584b66d372ee5eb09cf4571 )

jwiesler:
I had a look at it, the problematic lines are [here](https://github.com/KeYProject/key/blob/main/key.core/src/main/java/de/uka/ilkd/key/speclang/njml/JmlTermFactory.java#L346). I don't understand why this is done and whether it is ok to remove this.

The upper bound method [here](https://github.com/KeYProject/key/blob/main/key.core/src/main/java/de/uka/ilkd/key/speclang/njml/JmlTermFactory.java#L327) has the same problem.

I removed both lines, let's see whether the tests pass.

### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/speclang/njml/JmlTermFactory.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/speclang/njml/JmlTermFactory.java
@@ -324,9 +324,6 @@ public final class JmlTermFactory {
      * @return lower bound term (or null)
      */
     private Term lowerBound(Term a, LogicVariable lv) {
-        if (a.arity() > 0 && a.sub(0).op() == Junctor.AND) {
-            a = a.sub(0);
-        }
         if (a.arity() == 2 && a.op() == Junctor.AND && a.sub(0).arity() == 2
                 && a.sub(0).sub(1).op() == lv && a.sub(0).op()
                         .equals(services.getTypeConverter().getIntegerLDT().getLessOrEquals())) {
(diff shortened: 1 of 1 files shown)


mattulbrich (approved):
Your fix seems quite appropriate. It does fix the issue at hand. However, we cannot process such input yet but that's beyond this fix. MR 502 on gitlab tried to fix it more generally. We should have a look at this, too.

PR #3049, opened 2023-02-28
Position information overhaul

jwiesler:
* Removes most manual passing of line, column and replaces it with `Location`
* [Fix issue dialog pointing to the wrong line (at least on Windows)](https://github.com/KeYProject/key/commit/9894f7a5a8ca9b8b0b962d9565cced69dc60c89f)
* Fix many positioning issues coming from 0-based/1-based columns/lines
* Fix position information of block contracts and the like

Blocking on #3034. There are some usages in the `PrettyPrinter` I won't fix since it's deleted there.

[Artiweb](https://keyproject.github.io/artiweb/3049/)


mattulbrich (approved):
Approved after discussion. Thanks!

mattulbrich:
We should use this PR to substantially test our reporting both regarding locations and positions and the issue dialog.

PR #3059, opened 2023-03-09
Issue write permission for artiweb workflow

jwiesler:
Should fix the broken workflow.
@wadoon


jwiesler:
Doesn't work this way

PR #3064, opened 2023-03-12
Code cleanup

jwiesler:
This MR does some automated code cleanup.
Lots of changes, I grouped the changes in commits and the review should happen on a per commit basis. 

Some of them might be debateable, please do so.
However, I think most of them really help with parsing of and reasoning about the code.


jwiesler:
This PR already did some good, it resulted in a [bug report](https://youtrack.jetbrains.com/issue/IDEA-315243/Java-Inspection-Minimum-switch-branches-deletes-switch-with-side-effects?clearDraft=true&description=%0A%0AIU-222.3739.54,%20JRE%2017.0.3%2B7-b469.37x64%20JetBrains%20s.r.o.,%20OS%20Windows%2010(amd64)%20v10.0%20,%20screens%201920.0x1080.0,%201920.0x1080.0) for IntelliJ.

key.core/src/main/java/de/uka/ilkd/key/proof/io/ProofSaver.java
```diff
@@ -1,20 +1,24 @@
...
-import de.uka.ilkd.key.proof.Proof;
-import de.uka.ilkd.key.proof.io.event.ProofSaverEvent;
-import de.uka.ilkd.key.proof.io.event.ProofSaverListener;
-import de.uka.ilkd.key.util.KeYConstants;
-
 /**
  * Saves a proof and provides useful methods for pretty printing terms or programs.
  */
 public class ProofSaver extends OutputStreamProofSaver {
+    private static final Logger LOGGER = LoggerFactory.getLogger(ProblemInitializer.class);
```
mattulbrich, line 21: What's our policy towards Loggers? Should that not be `OutputStreamProofSaver.class`?

key.core/src/main/java/de/uka/ilkd/key/util/MiscTools.java
```diff
@@ -59,10 +25,27 @@
...
+import java.net.URL;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.*;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.zip.ZipFile;
+
 /**
```
mattulbrich, line 49: Again: Should spotless not take care of these irrelant changes in imports?

key.core/src/main/java/de/uka/ilkd/key/util/MiscTools.java
```diff
@@ -781,7 +764,7 @@ public static URI extractURI(DataLocation loc) {
                 return new URI("urn", loc.toString(), null);
             }
         } catch (URISyntaxException | IOException e) {
-            e.printStackTrace();
+
```
mattulbrich, line 767: Please do not drop an exception w/o logging (even if it is only at a low visibility level). Perhaps move the IAE from below and use e as cause.

key.core/src/test/java/de/uka/ilkd/key/proof/runallproofs/proofcollection/ForkedTestFileRunner.java
```diff
@@ -174,7 +174,7 @@ public static void main(String[] args) throws IOException {
                     testResults.add(testFile.runKey());
                 } catch (Exception e) {
                     error = true;
-                    e.printStackTrace();
+                    LOGGER.warn("Run failed", e);
```
mattulbrich, line 177: A general question: Will ".warn" be usually printed to the console by default?

key.core.symbolic_execution/src/main/java/de/uka/ilkd/key/symbolic_execution/AbstractUpdateExtractor.java
```diff
@@ -560,7 +560,7 @@ protected Term createLocationPredicateAndTerm(
             argumentsList.add(param.createPreValueTerm());
             param.setValueTermIndexInStatePredicate(++argumentIndex);
         }
-        Term[] arguments = argumentsList.toArray(new Term[argumentsList.size()]);
+        Term[] arguments = argumentsList.toArray(new Term[0]);
```
mattulbrich, line 563: see comment on other commit: Why is this better?

mattulbrich (changes requested):
Thanks for the spring clean!

I have a few comments and questions.

mattulbrich:
Apparently, this PR reshuffles many import statements and puts `.*` rather than individual imports (or v.v.). Is this not something that should be taken care of automatically in spotless?

@WolframPfeifer @jwiesler @wadoon

key.core/src/test/java/de/uka/ilkd/key/proof/runallproofs/proofcollection/ForkedTestFileRunner.java
jwiesler: Our log level is >= Info for all projects I know of, so yes. Any printing using `System.err` (which includes `e.printStackTrace`) is not included in any log files generated by KeY so they are lost after closing the console, so this code definitely should go. The only question is whether all log levels are configured correctly.

key.core.symbolic_execution/src/main/java/de/uka/ilkd/key/symbolic_execution/AbstractUpdateExtractor.java
jwiesler: From the inspection description: 

>In older Java versions, using a pre-sized array was recommended, as the reflection call necessary to create an array of proper size was quite slow.
> However, since late updates of OpenJDK 6, this call was intrinsified, making the performance of the empty array version the same, and sometimes even better, compared to the pre-sized version. Also, passing a pre-sized array is dangerous for a concurrent or synchronized collection as a data race is possible between the size and toArray calls. This may result in extra nulls at the end of the array if the collection was concurrently shrunk during the operation.

key.core/src/main/java/de/uka/ilkd/key/rule/AuxiliaryContractBuilders.java
```diff
@@ -1611,7 +1611,7 @@ private StatementBlock constructAbruptTerminationIfCascade() {
                 new NotEquals(
                     new ExtList(new Expression[] { variables.exception, NullLiteral.NULL })),
                 KeYJavaASTFactory.throwClause(variables.exception)));
-            return new StatementBlock(ifCascade.toArray(new Statement[ifCascade.size()]));
+            return new StatementBlock(ifCascade.toArray(new Statement[0]));
```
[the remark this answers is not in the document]
jwiesler: toArray(Statement[]::new) calls toArray(new Statement[0]) internally, so I see no reason to use it. Apart from that, see the explanation in the comment below.

key.core/src/main/java/de/uka/ilkd/key/control/instantiation_model/TacletInstantiationModel.java
```diff
@@ -1,35 +1,18 @@
...
-import de.uka.ilkd.key.proof.Proof;
-import de.uka.ilkd.key.proof.SVInstantiationException;
-import de.uka.ilkd.key.proof.SVInstantiationParserException;
-import de.uka.ilkd.key.proof.SortMismatchException;
-import de.uka.ilkd.key.rule.IfFormulaInstSeq;
-import de.uka.ilkd.key.rule.IfFormulaInstantiation;
-import de.uka.ilkd.key.rule.MatchConditions;
-import de.uka.ilkd.key.rule.Taclet;
-import de.uka.ilkd.key.rule.TacletApp;
+import de.uka.ilkd.key.proof.*;
```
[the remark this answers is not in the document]
jwiesler: I don't know, this is caused by optimize imports of IntelliJ, I didn't modify any settings.

key.core/src/main/java/de/uka/ilkd/key/util/MiscTools.java
jwiesler: It does not, there isn't even a warning.

There is a huge amount of files with unused imports (I cleaned most of them up I think).

key.core/src/main/java/de/uka/ilkd/key/util/MiscTools.java
jwiesler: Woops, I think I meant to throw the exception from two lines below here. Fixed in a follow up commit.

key.core/src/main/java/de/uka/ilkd/key/proof/io/ProofSaver.java
jwiesler: I don't know whether we have one but I think this should definitely be the current class and not the parent class (this one here is wrong, don't know why the cli option missed it).

It's very confusing to get logger messages from `BaseClass` and not find any log statement declared in this class.

key.core/src/main/java/de/uka/ilkd/key/util/MiscTools.java
mattulbrich: Why does it not look at that? It can't or it is not configured to do?

key.core/src/main/java/de/uka/ilkd/key/proof/io/ProofSaver.java
mattulbrich: Perhaps we should make that the rule then. (-> KaKeY)

jwiesler: Exception: `SyntaxErrorReporter`, although this is a very odd place to log and I think it should go away.

key.core/src/main/java/de/uka/ilkd/key/util/MiscTools.java
jwiesler: I don't know whether it is configured to do that. [Here](https://github.com/diffplug/spotless/blob/main/plugin-maven/README.md#java) is an example how to configure it, which I think we should to keep the order consistent, whatever order we decide on (e.g. first current project, then external libraries, finally standard library). 
@WolframPfeifer

key.core/src/main/java/de/uka/ilkd/key/proof/io/ProofSaver.java
jwiesler: I think this is should be added to the logging doc page, but I'm currently waiting on some changes by @wadoon to land before writing that one.

wadoon:
> Apparently, this PR reshuffles many import statements and puts `.*` rather than individual imports (or v.v.). Is this not something that should be taken care of automatically in spotless?
> 
> @WolframPfeifer @jwiesler @wadoon

Depends on the formatter. It seems that the eclipse formatter (that we currently use) can only reorder the imports.

jwiesler:
> 

I didn't find any option that orders the imports?

jwiesler:
That was a question, I checked the file that loads the configuration and didn't find anything the like.

wadoon:
> That was a question, I checked the file that loads the configuration and didn't find anything the like.

[You can find some traces](https://github.com/diffplug/spotless/issues/522) but in the end, the simplest thing is to configure the import via the eclipse GUI, and export an XML file.

mattulbrich:
Since there is a (partial?) solution to the import problem, this should be cleared after a merge with main.

mattulbrich (approved):

PR #3063, opened 2023-03-10
Re-apply view filter when switching proof

FliegendeWurst:
Previously, changing the view filter would only update the currently selected proof. With this fix, the view filter is refreshed after selecting another proof.

PR #3070, opened 2023-03-15
Fix ProofTreeView NPE

jwiesler:
Small fix, I think this was mentioned in other comments as well.


FliegendeWurst:
I also fixed this issue using a different strategy: https://github.com/KeYProject/key/pull/3063/files#diff-34f34dd32e949cfb08ca430aa677a077a252127746cc523f8f5c11e352ce868dR441-R444

My fix always selects the first node if there is no selection.

jwiesler:
Yes but only on load. That might be good, however I have no idea whether that's enough. I'd just merge these three lines and keep your changes as well.

mattulbrich (approved):

PR #3076, opened 2023-03-19
JSON-formatted log files

wadoon:


jwiesler:
`"` in log messages is not escaped in the JSON format.

key.ui/src/main/java/de/uka/ilkd/key/logging/JsonLayout.java
```diff
@@ -87,7 +86,11 @@ public String doLayout(ILoggingEvent event) {
         printEntry(pw, "thread", event.getThreadName());
         printEntry(pw, "level", event.getLevel().levelStr);
         printEntry(pw, "loggerName", event.getLoggerName());
-        printEntry(pw, "message", event.getFormattedMessage().replace("\\","\\\\"));
+        printEntry(pw, "message", event.getFormattedMessage()
+                .replace("\\", "\\\\")
+                .replace("\"", "\\\"")
+                .replace("\n", "\\n")
```
jwiesler: This looks very suspicious. Either JSON supports line breaks in strings anyways (as far as i know this is the case) or this breaks as soon as any other line break than `\n` appears inside a log message.

jwiesler: Did you check this with any other line breaks?

jwiesler:
Why does this not use logback's JsonFormatter? 
https://logback.qos.ch/manual/encoders.html

wadoon:
> Why does this not use logback's JsonEncoder? https://logback.qos.ch/manual/encoders.html

Either I did not find it, or it requires a JSON library that I do not want to attach.

PR #3077, opened 2023-03-20
Remove sonarqube from readme

jwiesler:


wadoon (approved):

jwiesler:
Good idea, didn't know that existed! 
* Does `**/src` match `**/src/*`? Why even add this when you are using file type filters?
* I'd add the `lib` folder or `*.jar` as well since there are some library files.
* `*.proof` for taclet/other proofs
* The `.txt` file for `runAllProofs` (or didn't you change that recently to a java file?)

Maybe there are some "battle-tested" paths we can take from an already existing repo? I'm not sure those are all...

mattulbrich:
Good idea! Let's collect other files that might be mentioned in this inclusion list. Is there also a way to state an exclusion list like "everything but ..."?

PR #3078, opened 2023-03-20
fixing error reporting in AbstractBuilder

mattulbrich:
the original code was not correct, threw NPEs and forgot the original cause exception.

* `e.getMessage()` could produce an NPE
* using `e.getMessage()` as format string was probably not the desired functionality here.

---

There is a second small bugfix related to error reporting: If an undefined schema variable is used as a bound variable, an NPE was thrown. This fix makes sure that "No schema variable x has been defined" is reported.

### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/nparser/builder/ExpressionBuilder.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/nparser/builder/ExpressionBuilder.java
@@ -1066,7 +1066,15 @@ public class ExpressionBuilder extends DefaultBuilder {
         if (sort != null && id != null) {
             return bindVar(id, sort);
         }
-        return doLookup(new Name(ctx.id.getText()), schemaVariables(), variables());
+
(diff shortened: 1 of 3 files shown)


key.core/src/main/java/de/uka/ilkd/key/nparser/builder/ExpressionBuilder.java
```diff
@@ -1047,7 +1047,16 @@ public QuantifiableVariable visitOne_bound_variable(KeYParser.One_bound_variable
         if (sort != null && id != null) {
             return bindVar(id, sort);
         }
-        return doLookup(new Name(ctx.id.getText()), schemaVariables(), variables());
+
+        // TODO Does this not check schemaVariables() for the second time?
```
jwiesler, line 1051: Yes, I'd remove `schemaVariables()` below. I think the check above was added later.

jwiesler (changes requested):

jwiesler (approved):

PR #3083, opened 2023-03-21
Proof caching

FliegendeWurst:
The changes in this PR allow for closing a goal by "reference" to another proof.
Implemented features:
- manually close by reference
- automatically search for references when applying strategy
- copy proof steps from referenced proof into new proof
- automated test of new feature
- modify statistics dialog to have useful buttons ("Close", "Export statistics...", "Save proof...", "Save proof bundle...", "Copy referenced proof...")

image

Testing on real proofs:
- SumAndMax: first proven with 3/4 ensures clauses, then with all of them => 20 cached branches

Other changes:
- Font Awesome is updated to version 6 for a new symbol


mattulbrich:
The name suggests that this is a draft PR, hence it marked it as such.

FliegendeWurst:
Merge conflicts with `main` have been fixed. Some initial tests with Aunt Agatha / SumAndMax indicate great potential (144/105 2316/1425 steps each).

key.core/src/main/java/de/uka/ilkd/key/java/Services.java
```diff
@@ -327,6 +327,14 @@ public Counter getCounter(String name) {
         return c;
     }
 
+    /**
+     * Reset all counters associated with this service.
+     * Only use this method if the proof is empty!
```
wadoon, line 332: Do you think it would be useful to add a guard? I think this is dangerous.

key.core/src/main/java/de/uka/ilkd/key/macros/TryCloseMacro.java
```diff
@@ -143,6 +143,8 @@ public ProofMacroFinishedInfo applyTo(UserInterfaceControl uic, Proof proof,
             return null;
         }
 
+        proof.register(this, TryCloseMacro.class);
```
wadoon, line 146: I think this is not necessary. `#register` is for adding a resource for later use to the lifecycle of a proof.

wadoon (approved):
There is change coming with this PR not mention in the description. Font Awesome is updated to version 6. 

I could not determine whether the caching information is stored between consecutive KeY runs.

wadoon:
> Discussion points:
> 
> * is it useful to have a manual action to close by reference? (currently in the proof context menu)

I have mixed feelings about this. On one side it might help the user to forgot about some branches and to focus on the important work. On the other side, it might be dangerous. 


I am asking for `ci-tool`. What is the behavior on the command line? Could closing by cache be suddenly applied inside of `ci-tool`?

key.core/src/main/java/de/uka/ilkd/key/macros/TryCloseMacro.java
FliegendeWurst: It is required so that the caching code knows this macro is running. When a branch is "closed" by the automatic caching search, it is left open. The macro notices this and undos all its steps, which is not desirable.

FliegendeWurst:
> > Discussion points:
> > 
> > * is it useful to have a manual action to close by reference? (currently in the proof context menu)
> 
> I have mixed feelings about this. On one side it might help the user to forgot about some branches and to focus on the important work. On the other side, it might be dangerous.

It is not dangerous, it performs the same checks as the automatic search before marking the branch as cached. See the documentation PR https://github.com/KeYProject/key-docs/pull/5 for more details.

> I am asking for `ci-tool`. What is the behavior on the command line? Could closing by cache be suddenly applied inside of `ci-tool`?

It depends on how exactly `ci-tool` works. If it doesn't load any extensions, the caching extension will not be activated. If only one proof is loaded at a time, there is no other proof to reference. In any case the caching can be disabled by the user or programatically via `CachingSettingsProvider`.

key.core/src/main/java/de/uka/ilkd/key/java/Services.java
FliegendeWurst: I have added a guard.

WolframPfeifer:
@FliegendeWurst A follow-up on our discussion from today: I think that in case of a proof that is closed, but has open dependencies left **and** has cached goals, the original "dependencies-left" icon should be used. It kind of "includes" the new icon, as it also has parentheses:
image
Btw.: Is it on purpose that the checkmark is thicker in your new icon (center)?

WolframPfeifer:
Apart from the icon, there are some points to do left from our discussion:
* [ ] Slight changes of the statistics dialog: show number of cached goals plus a short explanatory tooltip
* [ ] Add a setting about what should happen when a proof that is used as cache is removed. Possibilities are:
    * Transfer the cache (might take some time)
    * Reopen the goals
   We also discussed an option to disable automatic transfer of cached goals for saving, but I don't think it makes much sense.
* [ ] Behavior of the statistics dialog: Show once when a proof is closed modulo cache, and maybe a second time if the cache has successfully been transferred into "real" proof steps (maybe we can relax this later, but for now I think it is a good idea to notify the user of the replay status).
* [ ] Use the same dialog for "Proof Statistics" and "Subtree Statistics"

I also found another bug: The statistics dialog displays again (even opens twice) if I load another (unrelated) proof obligation, but have a "closed by cache" proof in my list.

FliegendeWurst:
The new statistics dialog (point 1 and 4 done):

image

Unfortunately I was not able to add a help icon or a dotted underline to indicate that a tooltip is available.

FliegendeWurst:
More icons:

image
image
image
image
image

FliegendeWurst:
I have finished working on all of the points.

key.ui/src/main/java/de/uka/ilkd/key/gui/plugins/caching/DefaultReferenceSearchDialogListener.java
```diff
@@ -0,0 +1,49 @@
...
+
+import de.uka.ilkd.key.core.KeYMediator;
+import de.uka.ilkd.key.gui.MainWindow;
+import de.uka.ilkd.key.gui.actions.ShowProofStatistics;
+import de.uka.ilkd.key.proof.Proof;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DefaultReferenceSearchDialogListener implements ReferenceSearchDialogListener {
```
WolframPfeifer, line 13: Can you provide a short JavaDoc, please?

key.util/src/main/java/org/key_project/util/EqualsModProofIrrelevancyUtil.java
```diff
@@ -99,7 +99,10 @@ public static int hashCodeIterable(Iterable<? extends EqualsModProofIrrelevancy>
     public static boolean compareImmutableLists(
             ImmutableList<? extends EqualsModProofIrrelevancy> a,
             ImmutableList<? extends EqualsModProofIrrelevancy> b) {
-        if (a.size() != b.size()) {
+        if (a == b || (a == null && b.size() == 0) || (b == null && a.size() == 0)) {
```
WolframPfeifer, line 102: Doesn't this throw a NPE if b is null (as well as if a == null and b.size() > 0)? Also, it seems strange that empty lists and null lists should be equal ...

key.ui/src/main/java/de/uka/ilkd/key/gui/prooftree/ProofTreeView.java
```diff
@@ -53,10 +54,21 @@ public class ProofTreeView extends JPanel implements TabPanel {
         ColorSettings.define("[proofTree]gray", "", Color.DARK_GRAY);
     public static final ColorSettings.ColorProperty LIGHT_BLUE_COLOR =
         ColorSettings.define("[proofTree]lightBlue", "", new Color(230, 254, 255));
+    /**
+     * Other green color, currently used for cached goals.
+     */
+    public static final ColorSettings.ColorProperty OTHER_GREEN_COLOR =
```
WolframPfeifer, line 60: I think we should use the same green color. The difference is currently barely noticeable and brings no benefit in my opinion.

WolframPfeifer (changes requested):
I am finally done with the review. Overall, the code quality is quite high and nearly everything is nicely documented. Thanks for that! Unfortunately, I still have some complaints:
Sometimes, the statistics dialog still opens multiple times. E.g., with the following class:
```java
class Caching {
    int f;

    /*@ ensures f == 7;
      @*/
    void m() {
        f = 7;
        //@ assert true;
    }
}
```
This seems to be a problem with macros. For the "apply strategy" button, it works. However, when I close the proof by using a macro and this macro utilizes the cach, I get 3 instances of the dialog.

In addition, I think that nodes should be marked as automatic again after they are closed, since otherwise they are "suddenly" interactive when pruning into a proof that was closed using the cache.

The current icon (database + checkmark) is good, as I already mentioned. I also discussed it with @mi-ki. However, in the proof tree there should be a better symbol for cached goals. Maybe also use the same icon? That would be consistent. Also, I would favor "Closed goal (via cache)" over the current "Cached goal".

In the proof management dialog, the same icons as in the task tree should be used.

It would be a good idea to make the context menu entry "Proof caching -> Close by reference to another proof" not only available for open goals, but also for other nodes (then, the cache search should be done for all subgoals).

Btw.: We recently used the caching in the case study of @gewitternacht, where it was really helpful, since multiple additional ensures clauses had to be added to a contract and the caching saved a lot of work. It would be really happy to have it in the release!

key.core/src/main/java/de/uka/ilkd/key/proof/Proof.java
```diff
@@ -311,6 +320,9 @@ public String header() {
         return problemHeader;
     }
 
+    public void setProblemHeader(String problemHeader) {
```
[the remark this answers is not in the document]
FliegendeWurst: It will be used for the database, but something like `saveWithProblemHeader` would be enough.

key.core/src/main/java/de/uka/ilkd/key/settings/PathConfig.java
```diff
@@ -38,6 +38,10 @@ public final class PathConfig {
      * In which file to store the proof-independent settings.
      */
     private static String proofIndependentSettings;
+    /**
+     * In which file to store the proof cache index.
+     */
+    private static File cacheIndex;
```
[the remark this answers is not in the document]
FliegendeWurst: Yes, I will remove them from this PR.

key.core/src/main/java/de/uka/ilkd/key/control/AbstractUserInterfaceControl.java
```diff
@@ -102,10 +102,14 @@ protected void fireTaskProgress(int position) {
...
-            proverTaskListener.toArray(new ProverTaskListener[0]);
-        for (ProverTaskListener l : listener) {
-            l.taskFinished(info);
+        try {
+            ProverTaskListener[] listener =
+                proverTaskListener.toArray(new ProverTaskListener[0]);
+            for (ProverTaskListener l : listener) {
+                l.taskFinished(info);
+            }
+        } catch (Exception e) {
```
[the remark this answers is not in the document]
FliegendeWurst: It is an error for the listener to throw an exception, since the other listeners will not get the event in that case.

key.util/src/main/java/org/key_project/util/EqualsModProofIrrelevancyUtil.java
FliegendeWurst: If a == null and b == null the first check `a == b` succeeds. Therefore at least one of a, b will be non-null when checking the rest of the condition.

If a is null and b.size() > 0, the next check in the function returns false.

I'm not sure this change is required at this point, I will check again.

WolframPfeifer: Ah, you are right, I was confused here ;-) We can leave it as is.

WolframPfeifer (changes requested):
Thanks for the quick response. However, now the button in the status bar is always disabled (at least during my test).

In addition, there are some problems in interplay with pruning. For example, I get a NPE when I use the "Close by reference" context menu entry on a node that has goal closed by caching and other open goals:
```
Exception in thread "AWT-EventQueue-0" java.lang.NullPointerException: Cannot invoke "de.uka.ilkd.key.proof.Goal.node()" because "goalToClose" is null
	at de.uka.ilkd.key.proof.Proof.closeGoal(Proof.java:532)
	at de.uka.ilkd.key.gui.plugins.caching.CachingExtension$CloseByReference.actionPerformed(CachingExtension.java:309)
```

And sometimes, when I prune a proof, run the macro again (close via cache), transferring the nodes from cache freezes the KeY GUI (waiting/running cursor).

Once I even had a goal marked as interactive after pruning, which then was marked as green (closed) "Open goal" after an additional TryClose. I was not able to reproduce this behavior, though.

Finally, I think we should keep the key_checks_incremental.xml.

FliegendeWurst:
`key_checks_incremental.xml` is generated by `scripts/tools/checkstyle/runIncrementalCheckstyle.sh`. It is safe to remove from the repository.

I will fix the other issues soon.

WolframPfeifer:
Copying the referenced proof steps via context menu crashes with the following exception (and thus freezes the GUI):
```
Exception in thread "AWT-EventQueue-0" java.lang.NullPointerException: Cannot invoke "de.uka.ilkd.key.proof.Goal.proof()" because "newNode" is null
	at de.uka.ilkd.key.proof.replay.CopyingProofReplayer.copy(CopyingProofReplayer.java:39)
	at de.uka.ilkd.key.gui.plugins.caching.CachingExtension$CopyReferencedProof.actionPerformed(CachingExtension.java:359)
   ...
```

FliegendeWurst:
@WolframPfeifer all checks have passed, I think this is ready to merge.

WolframPfeifer:
> I fixed the exception, the root cause was a misleading method name: Proof::getGoal only returns _open_ goals...

This is for historical reasons: Before pruning in closed goals was implemented, the terminology was that a "goal" is always open ("closed goals" were only there for rendering, since the underlying data structures were removed for memory reasons directly after the goal was closed). Nowadays, it is a good idea to make the name clearer, thanks!

WolframPfeifer (approved):
Everything I tested seems to work. I will merge it, thanks for implementing the feature.

To be sure it is production ready, we need to test the feature a bit more before the release.

PR #3094, opened 2023-03-31
set an import order for Java code

wadoon:
After discussion in the KaKeY, we finally set an import order for Java source files.

The only relevant non-automatic change is in `build.gralde`


WolframPfeifer (approved):
Nice! Thank you!

PR #3096, opened 2023-04-02
JUnit XML files for optional-tests

wadoon:
JUnit XML files for optional-tests were not uploaded to the task guard.


jwiesler (approved):
Looks good. 

Is there any info relayed to Artiweb about which tests are optional and which are not such that this can be displayed?

PR #3100, opened 2023-04-03
Fix type of `Long.MIN_VALUE` and `Long.MAX_VALUE`

ChristianHein:
`Long.MIN_VALUE` and `Long.MAX_VALUE` should be of type `long` like it is in the JDK sources.

This change shouldn't impact any previously working proofs since these values get translated to the correct numbers in JavaDL as sort `int` regardless.

Theoretically this change could introduce new type issues when parsing previously working Java code, but this seems unlikely since the `int` types would probably have been widened to `long` before anyway.


jwiesler (approved):

mattulbrich:
Nice catch, should not even have been accepted in the first place. Recoder does not seem to make all necessary type checks.

jwiesler:
I wonder why this compiled in the first place: assigning a `long` literal to an `int` requires a narrowing conversion which must not be implicit.

PR #3105, opened 2023-04-05
AutoPerf: Replace RuleAppCost with long

jwiesler:
Removes RuleAppCost making lots of the code easier to reason about (and a bit faster).

Blocking on https://github.com/KeYProject/key/pull/3124


jwiesler:
The queue based automatic is extremely fragile: It depends *heavily* on the ordering of rule apps with the same cost given by the current queue implementation.

Changing the used queue or forcing an explicit ordering on rule apps with the same cost makes two proofs (in our test suite) non closeable.
And I do not mean it takes more steps, the automatic ends with **no further** applicable rules!

mattulbrich:
From a quick inspection, it seems you replaced `RuleCost`s by `long`s. I assume that class had been introduced for a reason at its time. Richard @unp1, do you know about them?

How much performance do you buy us be removing the `RuleCost` data structure? (It seems like (auto)-boxing /-unboxing ...)

jwiesler:
I replaced them by `long` since all costs are either instances of `NumberRuleAppCost` or `TopRuleAppCost`. There are no other implementations of the interface.
`NumberRuleAppCost` is either a `LongRuleAppCost` or `IntRuleAppCost` and uses an internal locked cache to construct instances.

Using `long` does not introduce any precision changes or posibilities of overflow that did not exist before and is simply a lot cleaner than a class that is essentially a `Long`.

@wadoon even added a comment (a few years ago) to the class that the whole mechanism should be replaced with just using `float`, however this could change the resulting costs due to floating point precision which this change does not.

The performance gains aren't large, however, I think this is cleaner. Many cost functions could only handle `NumberRuleAppCost` anyways and explicitly panic if any other cost is used so why even use this interface?

mattulbrich:
However, `TopRuleAppCost` does not belong to a long value. Is it free of implications if that is replaced by an actual number?

jwiesler:
Of course not, but I think it's safe to assume that if we ever hit `long::MAX_VALUE` as a cost before we also would have gotten overflows.

Most costs are well below `1_000_000`. There isn't even any multiplication of costs, only addition.

wadoon:
> @wadoon even added a comment (a few years ago) to the class that the whole mechanism should be replaced with just using `float`, however this could change the resulting costs due to floating point precision which this change does not.

My idea was that float/double has an explicit `inf` value, as well as `NaN`, which have the properties we emulate with the adt `RuleCost`.

PR #3107, opened 2023-04-06
Lost rules: Seqperm, slice all taclet proofs

jwiesler:
* Fix some error reporting around `TacletProofObligationInput`
* Recover some lost (proven) rules involving `seqPerm`
* Slice all taclet proofs


key.core/src/main/resources/de/uka/ilkd/key/proof/rules/seqPerm.key
```diff
@@ -608,4 +608,52 @@
         \displayname "seqPermTrans"
     };
 
+    // Proven by ulbrich in isabell
```
mattulbrich, line 611: Small comment. It's "Isabelle/HOL".

mattulbrich (changes requested):
There is an open question by the author:

> Edit 2: Slicing also deletes statistics like autoModeTime. Is there a way to keep those?

jwiesler:
The automode time is now being used for the newly created sliced proof.

mattulbrich (approved):

PR #3109, opened 2023-04-07
Depromote optional-tests

wadoon:
Optional tests are currently the tests of symbolic execution and proof references, required by KeY4clipse. 

**Before:** Optional tests were always executed on pushes on main and pull-requests.

**With this PR:** Optional tests are only executed every 10 days on the main branch or on manually demand.


mattulbrich (approved):
Since the optional tests do not tell us a lot and always fail, I'd prefer to demote them.

PR #3115, opened 2023-04-13
Artiweb workflow fix

jwiesler:
Fix failure on empty file


jwiesler:
@wadoon could you check why the opttest workflow is failling? It reports a syntax error in line 15, however I see nothing.

wadoon (approved):

PR #3117, opened 2023-04-15
Critical changes to the KeY parser/grammer for more performance

wadoon:
This PR brings restructuring of the KeY's grammar `KeYParser.g4`  for more performance.

Some of the changes are critical as they change the grammar, and need to be discuss.

The **current situation** is as follows. All given benchmarks given on our biggest Key-file `javaRules.key`. *Parse time* should be treated carefully, as JIT and caching effects have critical. 

Screenshot from 2023-04-15 03-36-14

You see a lots of ambiguities and sometimes high maximal lookahead (max. 105).

The problem lies more in the *primitives* and some stupid rules. 

The following is possible, depending on the severity on changes. 

Screenshot from 2023-04-15 04-03-41

Screenshot from 2023-04-15 04-01-50

### For discussing the changes, please refer to the comments on the source code below. 

/cc @jwiesler @mattulbrich @unp1

### DIFF
--- a/key.core/src/main/antlr4/KeYParser.g4
+++ b/key.core/src/main/antlr4/KeYParser.g4
@@ -99,7 +99,7 @@ one_sort_decl
 
 simple_ident_dots
 :
-  simple_ident (DOT simple_ident)* | INT_LITERAL
+  simple_ident (DOT simple_ident)*
 ;
 
 simple_ident_dots_comma_list
@@ -329,7 +329,8 @@ literals:
   | string_literal
 ;
 
-term: parallel_term;
+
+term: equivalence_term;
 //labeled_term: a=parallel_term (LGUILLEMETS labels=label RGUILLEMETS)?;
 parallel_term: a=elementary_update_term (PARALLEL b=elementary_update_term)*;
 elementary_update_term: a=equivalence_term (ASSIGN b=equivalence_term)?;
@@ -348,7 +349,7 @@ comparison_term: a=weak_arith_term ((LESS|LESSEQUAL|GREATER|GREATEREQUAL) b=weak
 weak_arith_term: a=strong_arith_term_1 (op+=(PLUS|MINUS) b+=strong_arith_term_1)*;
 strong_arith_term_1: a=strong_arith_term_2 (STAR b+=strong_arith_term_2)*;
 strong_arith_term_2: a=atom_prefix ((PERCENT|SLASH) b=strong_arith_term_2)?;
-update_term: (LBRACE u=term RBRACE) (atom_prefix | unary_formula);
+update_term: (LBRACE u=parallel_term RBRACE) (atom_prefix | unary_formula);
 substitution_term:
  LBRACE SUBST  bv=one_bound_variable SEMI
      replacement=comparison_term RBRACE
@@ -360,7 +361,7 @@ atom_prefix:
     update_term
   | substitution_term
   | locset_term
-  | cast_term
+  //| cast_term
   | unary_minus_term
   | bracket_term
 ;
@@ -381,8 +382,15 @@ primitive_term:
   | ifThenElseTerm
   | ifExThenElseTerm
   | abbreviation
-  | accessterm
-  | literals
+  | prim_conflicting;
+
+prim_conflicting:
+   accessterm  |
+     boolean_literal
+     | char_literal
+     | integer
+     | floatnum
+     | string_literal
 ;
 
 /*
@@ -442,8 +450,16 @@ term
  */
 accessterm
 :
-  (sortId DOUBLECOLON)?
-  firstName=simple_ident
+  // OLD
+  //(sortId DOUBLECOLON)?
+  //firstName=simple_ident
+
+  //Faster version
+  simple_ident_dots
+  ( EMPTYBRACKETS*
+    DOUBLECOLON
+    simple_ident
+  )?
   call?
   ( attribute )*
 ;
@@ -522,8 +538,9 @@ argument_list
     RPAREN
 ;
 
+integer_with_minux: MINUS? integer;
 integer:
-  (MINUS)? (INT_LITERAL | HEX_LITERAL | BIN_LITERAL)
+  (INT_LITERAL | HEX_LITERAL | BIN_LITERAL)
 ;
 
 floatnum: // called floatnum because "float" collide with the Java language
@@ -665,7 +682,9 @@ varexpId: // weigl, 2021-03-12: This will be later just an arbitrary identifier.
 
 varexp_argument
 :
-    sortId //also covers possible varId
+    //weigl: Ambguity between term (which can also contain simple_ident_dots and sortId)
+    //       suggestion add an explicit keyword to request the sort by name or manually resolve later in builder
+    SORT sortId //also covers possible varId
   | TYPEOF LPAREN y=varId RPAREN
   | CONTAINERTYPE LPAREN y=varId RPAREN
   | DEPENDINGON LPAREN y=varId RPAREN
@@ -691,8 +710,7 @@ option
 option_list
 :
   LPAREN
-    ( (option (COMMA option)*)
-      | option_expr)
+    (option_expr (COMMA option_expr)*)
   RPAREN
 ;
 
--- a/key.core/src/main/antlr4/KeYParser.g4
+++ b/key.core/src/main/antlr4/KeYParser.g4
@@ -304,7 +304,7 @@ id_declaration
 
 funcpred_name
 :
-  (sortId DOUBLECOLON)? name=simple_ident_dots
+  (sortId DOUBLECOLON)? (name=simple_ident_dots|num=INT_LITERAL)
 ;
 
 
@@ -330,7 +330,7 @@ literals:
 ;
 
 
-term: equivalence_term;
+term: parallel_term; // weigl: should normally be equivalence_term
 //labeled_term: a=parallel_term (LGUILLEMETS labels=label RGUILLEMETS)?;
 parallel_term: a=elementary_update_term (PARALLEL b=elementary_update_term)*;
 elementary_update_term: a=equivalence_term (ASSIGN b=equivalence_term)?;
@@ -361,7 +361,7 @@ atom_prefix:
     update_term
   | substitution_term
   | locset_term
-  //| cast_term
+  | cast_term
   | unary_minus_term
   | bracket_term
 ;
@@ -382,16 +382,9 @@ primitive_term:
   | ifThenElseTerm
   | ifExThenElseTerm
   | abbreviation
-  | prim_conflicting;
-
-prim_conflicting:
-   accessterm  |
-     boolean_literal
-     | char_literal
-     | integer
-     | floatnum
-     | string_literal
-;
+  | accessterm
+  | literals
+  ;
 
 /*
 weigl, 2021-03-12:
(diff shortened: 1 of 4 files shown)


key.core/src/main/antlr4/KeYParser.g4
```diff
@@ -99,7 +99,7 @@ one_sort_decl
 
 simple_ident_dots
 :
-  simple_ident (DOT simple_ident)* | INT_LITERAL
```
wadoon, line 102: For some reason, a number (1,2,3, etc) can also be treated as a dotted identifier (e.g., `java.lang.Object`). 

Hence, `2522` would be a valid name of a sort (`sordId` derives to `simple_ident_dots`). 

This lead to an ambiguity: `23423` can be an `integer`, or and `accessterm`.

wadoon, line 102: For some reason, a number (1,2,3, etc) can also be treated as a dotted identifier (e.g., `java.lang.Object`). 

Hence, `2522` would be a valid name of a sort (`sordId` derives to `simple_ident_dots`). 

This lead to an ambiguity: `23423` can be an `integer`, or and `accessterm`.

key.core/src/main/antlr4/KeYParser.g4
```diff
@@ -329,7 +329,8 @@ literals:
   | string_literal
 ;
 
-term: parallel_term;
```
wadoon, line 332: The entry point to the `term` was `parallel_term`, which is a sequence of terms seperated by `||`. 

This mean, that `a || b` would be a valid term, although the `||` is only allowed inside UPDATES. 

Here I changed it: Now the top-level term is an `equivalence_term` (`a <-> b <-> c`), and `update_term` points `parallel_term`. 

I think, this is also not quite right. The term `{ 2+2 } x` should be rejected by the grammar, but is currently allowed. The assignment in `elementary_update_term` should be mandatory, but I think it is required by matcher expression.

key.core/src/main/antlr4/KeYParser.g4
```diff
@@ -360,7 +361,7 @@ atom_prefix:
     update_term
   | substitution_term
   | locset_term
-  | cast_term
+  //| cast_term
```
wadoon, line 364: We should discuss, if the `cast_term` is needed. 

`cast_term` allows you to write Java-like cast expression `(int) x` in KeY. Drawback is the required lookahead: The parser needs to check the whole expression inside the parenthesis, until it recognise the missing operator after the right paren.

On the other side, we also use `<sortId>::<expr>` (in `accessterm`) for casting.

key.core/src/main/antlr4/KeYParser.g4
```diff
@@ -381,8 +382,15 @@ primitive_term:
   | ifThenElseTerm
   | ifExThenElseTerm
   | abbreviation
-  | accessterm
-  | literals
+  | prim_conflicting;
```
wadoon, line 385: This was a refactoring to locate the ambiguity between `accessterm` and `literals`. 

The answer is in `INT_LITERAL` in the rule `simple_ident_dots`.

key.core/src/main/antlr4/KeYParser.g4
```diff
@@ -442,8 +450,16 @@ term
  */
 accessterm
 :
-  (sortId DOUBLECOLON)?
-  firstName=simple_ident
+  // OLD
```
wadoon, line 453: Rewrite of `acessterm`. It is more uglier, but requires no lookahead. 

I am not sure whether we should do it.

key.core/src/main/antlr4/KeYParser.g4
wadoon: Will be reverted before merge.

key.core/src/main/antlr4/KeYParser.g4
```diff
@@ -522,8 +538,9 @@ argument_list
     RPAREN
 ;
 
+integer_with_minux: MINUS? integer;
 integer:
-  (MINUS)? (INT_LITERAL | HEX_LITERAL | BIN_LITERAL)
```
wadoon, line 526: The minus is a prefix operator and already handled by `unary_minus_term`.

On the other side, there might be some rules, that want allow negative integers, e.g., in the non-expression part of the grammar.

key.core/src/main/antlr4/KeYParser.g4
```diff
@@ -665,7 +682,9 @@ varexpId: // weigl, 2021-03-12: This will be later just an arbitrary identifier.
 
 varexp_argument
 :
-    sortId //also covers possible varId
```
wadoon, line 668: Ambiguity: `x` might be a name of a sort, or a variable. 

Either we resolve this programmatically during the  taclet building or by a new identifier, e.g., `\varcond(\new(\sort(x)))`.

key.core/src/main/antlr4/KeYParser.g4
```diff
@@ -691,8 +710,7 @@ option
 option_list
 :
   LPAREN
-    ( (option (COMMA option)*)
-      | option_expr)
+    (option_expr (COMMA option_expr)*)
```
wadoon, line 713: An `option` (`integer:on`) is derivable directly or via `option_expr`. 

This resolves the ambiguity by saying that a list of `option_expr` is allowed, therefore this covers the old behavior of:

```
\rules(programRules:Java, runtimeExceptions:allow) {
```


A list of `option_expr` is treated as a conjunction.

mattulbrich:
In the description @wadoon wrote: 

> Some of the changes are critical as they change the grammar, and need to be discuss[ed].

As long as the language remains the same, this should be no problem.
Or do you suggest breaking changes to the language? Can all existing files in the codebase be read with the new parser?
If there are implications on the language, can you point them out?

key.core/src/main/antlr4/KeYParser.g4
mattulbrich: I think that 0,1,2,...,9 are used as function names and function names are `simple_ident_dots`. This might be solved by moving the `INT_LITERAL` to `funcpred_name` if that was the reason.

mattulbrich: I find the fact that `1` can be an int constant (resolving to `Z(1(#))`) and at the same time the name of the function used inside the constants a little unsatisfying, I must admit. (But ... is it worth changing that?)

@unp1

key.core/src/main/antlr4/KeYParser.g4
mattulbrich: It will (probably) be rejected later during AST resolution. If you can make the grammar clearer, sure!
Updates can be: Identifiers, assignments, parallel updates, empty. I believe, I also saw cases of strangely nested updates, that should be checked.

key.core/src/main/antlr4/KeYParser.g4
mattulbrich: Do you mean `<sortId>::cast(<expr>)`?

That would be breaking change that needs to be discussed.

key.core/src/main/antlr4/KeYParser.g4
mattulbrich: Yes.

key.core/src/main/antlr4/KeYParser.g4
mattulbrich: I think the parentheses `\sort(x)` are not in the grammar

Is every word in sortId also in term? Then we could drop this here and have this decided programmatically (as it probably is right now)

key.core/src/main/antlr4/KeYParser.g4
mattulbrich: You could probably also change `(option (COMMA option)*)` to ` (option (COMMA option)+)` to obtain the desired effect, right?

mattulbrich (changes requested):
Some points are breaking languages changes and need discussion.

If we renovate the language anyway, perhaps should can touch a few other places?

key.core/src/main/antlr4/KeYParser.g4
jwiesler: Couldn't they just be called `ZN` or `_N`?

mattulbrich: If you mean `Z0, Z1, ...` then I would say this would be clearer; but a change at the very roots.

wadoon: On the other side, we have a special lexer thing, that lifts numbers before parens to identifier. But I needed to add `INT_LITERAL` to `funcpred_name` to parse some files again.

https://github.com/KeYProject/key/blob/a9ec83b0c80fa10bdf24dc9b94713c19d45d18d5/key.core/src/main/antlr4/KeYLexer.g4

```
   public void emit(Token token) {
       int MAX_K = 10;
       if (token.getType() == INT_LITERAL) {//rewrite INT_LITERALs to identifier when preceeded by an '('
           for (int k = 1; k <= MAX_K; k++) {
               int codePoint = _input.LA(k);
               if (Character.isWhitespace(codePoint)) continue;
               if (codePoint == '(') ((org.antlr.v4.runtime.CommonToken) token).setType(IDENT);
               break;
           }
       }
       if(token.getType() == PROOF) {
          tokenBackStorage = super.emitEOF();
          //will later be overwritten the EOF token
       }
       super.emit(token);
    }
```

key.core/src/main/antlr4/KeYParser.g4
wadoon: Yeah, the `updateRules.key` is the problem. Here there are rules with problematic terms:

```
    applyOnParallel {
        \schemaVar \update u, u2, u3;

        \find({u}(u2 || u3))

        \replacewith({u}u2 || {u}u3)

        \heuristics(update_apply_on_update)
    };
```

Not sure how to solve this.

key.core/src/main/antlr4/KeYParser.g4
wadoon: Yes. But at least `cast_term` is not ambigue.

key.core/src/main/antlr4/KeYParser.g4
wadoon: No. You still need to lookbehind the `option` to recognize the if there is a `COMMA`.

mattulbrich: really? I'd say this resolves ambiguity.

wadoon:
I pushed a new version, which preserves the language (but has some of the old drawbacks). 

Parsing of key's standard lib take ca. 800 - 900 ms for 30886 lines (on this branch). For comparison, Javaparser requires 400ms to read JavaRedux (2290 lines, w/o JML).

key.core/src/main/antlr4/KeYParser.g4
wadoon: I currently see no solution. We are back on the old track.

key.core/src/main/antlr4/KeYParser.g4
mattulbrich: I'd like to keep the syntax.

mattulbrich (approved):

mattulbrich:
During meeting on 7 Jul: We plan to refactor this grammar in near future for developers' convenience.

Let's fix #3130 then.

PR #3129, opened 2023-04-20
Fix double highlight in sequent view

FliegendeWurst:
If the new sequent has a different height than the previous view, repaint is called twice. To fix overlapping highlights, remove all highlights when painting the sequent.


wadoon (approved):
Why is `printSequent` synchronized? I would guess it should only be called EDT.

PR #3133, opened 2023-04-28
Fix NPE while extracting recoder error position and use existing information better

jwiesler:


WolframPfeifer (changes requested):
Thanks for the quick fix!
However, could you please remove "var" from the code and make the types explicit? As far as I know, we agreed in a discussion that it makes the code more difficult to read. From my point of view, it is ok in tests and debug code, but not in the "core"/main classes.

wadoon:
> Thanks for the quick fix! However, could you please remove "var" from the code and make the types explicit? As far as I know, we agreed in a discussion that it makes the code more difficult to read. From my point of view, it is ok in tests and debug code, but not in the "core"/main classes.

I am not aware that we agreed on a consens.

wadoon (approved):

WolframPfeifer:
> > Thanks for the quick fix! However, could you please remove "var" from the code and make the types explicit? As far as I know, we agreed in a discussion that it makes the code more difficult to read. From my point of view, it is ok in tests and debug code, but not in the "core"/main classes.
> 
> I am not aware that we agreed on a consens.

I thought that we had, but maybe we should discuss it in our next meeting then...

wadoon:
> I thought that we had, but maybe we should discuss it in our next meeting then...

For the discussion: For example, if the rhs is a constructor invocation, the type is clear. 

```java 
var x = new HashMap<Int, new HashMap<List<Int>, Function<String, String>>>;
```

jwiesler:
@WolframPfeifer removed var where necessary.

mattulbrich:
Julian @jwiesler, I added to this branch and created a PR in your repo.

jwiesler:
@mattulbrich as a member you can just push to my branch in case you didn't know.

WolframPfeifer (approved):

PR #3140, opened 2023-05-15
Fix highlighting of selected terms

FliegendeWurst:
After #3129, this was broken. Clearing all highlights also removes the assocation to the stored tags (`defaultHighlight` / `additionalJavaHighlight`), so they have to be added again.

### DIFF
--- a/key.ui/src/main/java/de/uka/ilkd/key/gui/nodeviews/SequentView.java
+++ b/key.ui/src/main/java/de/uka/ilkd/key/gui/nodeviews/SequentView.java
@@ -98,13 +98,13 @@ public abstract class SequentView extends JEditorPane {
     public boolean refreshHighlightning = true;
 
     // the default tag of the highlight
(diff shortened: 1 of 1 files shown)


WolframPfeifer:
@FliegendeWurst I do not quite understand the problem, on current main everything seems fine for me (highlighting of instantiated terms and active statements seems to work). Also, I am not able to observe a change in behavior compared to this PR. What exactly is the bug here? Can you list the steps to reproduce it?

FliegendeWurst:
I think the issue only occurs in closed proofs.

WolframPfeifer (approved):
I did the review and after some experiments could find out what the actual bug was. For documentation, I added a bug description to the initial comment. The proposed fix works nicely.

@FliegendeWurst Can you please add at least a short description of the bug in the future? This helps with the review :wink:. Nonetheless, thanks for the fix!

PR #3141, opened 2023-05-15
Proof reordering (WIP)

FliegendeWurst:
Experiment to logically group steps in proof.

Two main features:
- "Reorder" proof steps such that adjacent proof steps modify the same formula (no interweaving of two simplification chains)
- "Regroup" proof steps in the proof tree view (UI only)

TODOs
- [x] Handle pruning correctly after regrouping
- [ ] Tests
- [ ] Documentation

The order of proof steps is determined as follows:

1. Steps with no inputs (no find / assumes)
2. Steps ordered by the dep. graph data
3. (optional) Step with no input that splits (e.g. `sign_case_distinction`)

To order the steps based on the dependency graph a queue of "available" graph nodes is managed. This queue is repeatedly iterated through to check for edges (steps) that are applicable. If they aren't the node is added to the end of the queue again. (More details will follow soon™)


WolframPfeifer:
The branch `pfeifer/proofReorderingMainMerge` contains my attempts to update the changes. I don't update the PR since it does not work at the moment, but I think there are only slight adaptations needed to fix it.

Something that already does not work on this PR without main merge: When trying to add a group in the settings, the dialog appears "behind" and input is blocked by the settings window, thus unusable. Probably, the parent of the dialog just needs to be set correctly.

PR #3152, opened 2023-05-31
Test parsing exceptions in JML and Java parsing 

mattulbrich:
JML parsing and error reporting is not yet ideal, we should fix these case.
They can also serve as valuable test cases for future replacement frameworks.

This also fixes a bug in the position treatement of Recoder2KeY.

From the javadoc:

This test case is used to ensure that errors in JML (and perhaps also in Java)
are reported with a reasonable error message and the right position pointing
into the file.
To add a test case, locate the "exceptional" directory in the resources
(below the directory for this package here) and add your single Java file
that contains an error that should be presented to the user (like syntax
error, unresolved names, ...)
The first lines of the Java file may contain meta data on what to expect
from the exception.

- introducing exceptional JML tests
- more test cases for exceptional jml cases
- fixing an NPE bug in the error reporting of Recoder2KeY
- spotless and documentation

### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/java/Recoder2KeY.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/java/Recoder2KeY.java
@@ -1255,7 +1255,7 @@ public class Recoder2KeY implements JavaReader {
         }
 
         Pair<String, Position> pos = extractPositionInfo(cause.toString());
-        if(pos != null) {
+        if (pos != null) {
             reportErrorWithPositionInFile(message, cause, pos.second, pos.first);
         } else {
             reportErrorWithPositionInFile(message, cause, null, null);
--- a/key.core/src/main/java/de/uka/ilkd/key/java/Recoder2KeY.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/java/Recoder2KeY.java
@@ -1254,8 +1254,12 @@ public class Recoder2KeY implements JavaReader {
             throw (PosConvertException) cause;
         }
 
-        var pos = extractPositionInfo(cause.toString());
-        reportErrorWithPositionInFile(message, cause, pos.second, pos.first);
+        Pair<String, Position> pos = extractPositionInfo(cause.toString());
+        if(pos != null) {
+            reportErrorWithPositionInFile(message, cause, pos.second, pos.first);
+        } else {
+            reportErrorWithPositionInFile(message, cause, null, null);
+        }
     }
 
     public static void reportErrorWithPositionInFile(String message, Throwable cause,
--- a/key.core/src/test/java/de/uka/ilkd/key/speclang/njml/JMLParserExceptionTest.java
+++ b/key.core/src/test/java/de/uka/ilkd/key/speclang/njml/JMLParserExceptionTest.java
@@ -72,7 +72,6 @@ public class JMLParserExceptionTest {
 
     // This method does not depend on anything can also be called from other test cases.
     public static void parseAndInterpret(Path file) throws Exception {
-
         List<String> lines = Files.readAllLines(file);
         Properties props = new Properties();
         for (String line : lines) {
@@ -102,7 +101,7 @@ public class JMLParserExceptionTest {
 
         } catch (Throwable e) {
             if ("true".equals(props.getProperty("verbose"))) {
-                LOGGER.info("Exception raised while parsing " + file.getFileName(), e);
+                LOGGER.info("Exception raised while parsing {}", file.getFileName(), e);
             }
 
             try {
@@ -144,8 +143,7 @@ public class JMLParserExceptionTest {
                 }
             } catch (AssertionFailedError assertionFailedError) {
                 // in case of a failed assertion log the stacktrace
-                LOGGER.debug("Original stacktrace leading to failed junit assertion in "
-                    + file.getFileName(), e);
+                LOGGER.debug("Original stacktrace leading to failed junit assertion in {}", file.getFileName(), e);
                 throw assertionFailedError;
             }
         }
--- a/key.core/src/test/java/de/uka/ilkd/key/speclang/njml/JMLParserExceptionTest.java
+++ b/key.core/src/test/java/de/uka/ilkd/key/speclang/njml/JMLParserExceptionTest.java
@@ -105,7 +105,8 @@ public class JMLParserExceptionTest {
 
     public static final boolean IGNORE_BROKEN = true;
 
-    private final static Pattern PROP_LINE = Pattern.compile("//\\s*(\\p{Alnum}+)\\s*[:=]\\s*(.*?)\\s*");
+    private final static Pattern PROP_LINE =
+        Pattern.compile("//\\s*(\\p{Alnum}+)\\s*[:=]\\s*(.*?)\\s*");
 
     public static Stream<Arguments> getFiles() throws URISyntaxException, IOException {
         URL fileURL = JMLParserExceptionTest.class.getResource("exceptional");
--- a/key.core/src/test/java/de/uka/ilkd/key/speclang/njml/JMLParserExceptionTest.java
+++ b/key.core/src/test/java/de/uka/ilkd/key/speclang/njml/JMLParserExceptionTest.java
@@ -1,0 +1,99 @@
+package de.uka.ilkd.key.speclang.njml;
+
+import de.uka.ilkd.key.control.DefaultUserInterfaceControl;
+import de.uka.ilkd.key.control.UserInterfaceControl;
+import de.uka.ilkd.key.proof.io.AbstractProblemLoader;
+import org.junit.jupiter.api.Assumptions;
+import org.junit.jupiter.params.ParameterizedTest;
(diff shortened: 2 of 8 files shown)


key.core/src/main/java/de/uka/ilkd/key/java/Recoder2KeY.java
```diff
@@ -1254,8 +1254,12 @@ public static void reportError(String message, Throwable t) {
             throw (PosConvertException) cause;
         }
 
-        var pos = extractPositionInfo(cause.toString());
-        reportErrorWithPositionInFile(message, cause, pos.second, pos.first);
+        Pair<String, Position> pos = extractPositionInfo(cause.toString());
```
wadoon, line 1257: Here, it would be useful to avoid `Pair` and use one of the location classes available in KeY.

key.core/src/test/java/de/uka/ilkd/key/speclang/njml/JMLParserExceptionTest.java
```diff
@@ -0,0 +1,199 @@
...
+
+    public static final boolean IGNORE_BROKEN = true;
+
+    private final static Pattern PROP_LINE =
+        Pattern.compile("//\\s*(\\p{Alnum}+)\\s*[:=]\\s*(.*?)\\s*");
+
+    public static Stream<Arguments> getFiles() throws URISyntaxException, IOException {
+        URL fileURL = JMLParserExceptionTest.class.getResource("exceptional");
+        assert fileURL != null : "Directory 'exceptional' not found";
+        assert fileURL.getProtocol().equals("file") : "Test resources must be in file system";
```
wadoon, line 114: Either use `getResource` but then do not expect a file system, or just use the file system directly. Test cases are executed in the corresponding project folder.

key.core/src/test/java/de/uka/ilkd/key/speclang/njml/JMLParserExceptionTest.java
```diff
@@ -0,0 +1,199 @@
...
+        Path dir = Paths.get(fileURL.toURI());
+        return Files.list(dir).map(it -> Arguments.of(it, it.getFileName()));
```
wadoon, line 116: Maybe, `Files#walk` is more useful, as it allows to create sub folders.

key.core/src/test/java/de/uka/ilkd/key/speclang/njml/JMLParserExceptionTest.java
```diff
@@ -0,0 +1,199 @@
...
+import org.junit.jupiter.api.Assumptions;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.opentest4j.AssertionFailedError;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
+ * This test case is used to ensure that errors in JML (and perhaps also in Java)
```
wadoon, line 32: Maybe a `README.md` inside the resource folder is easier to read and could also be reused on key-docs.

wadoon (changes requested):

key.core/src/test/java/de/uka/ilkd/key/speclang/njml/JMLParserExceptionTest.java
```diff
@@ -0,0 +1,199 @@
...
+                    Location actLoc = ExceptionTools.getLocation(e);
+                    assertNotNull(actLoc, "Exception location must not be null");
+                    assertEquals(file.toUri().toURL(), actLoc.getFileURL(),
+                        "Exception location must point to file under test");
+                    assertEquals(loc, actLoc.getPosition().toString());
+                }
+            } catch (AssertionFailedError assertionFailedError) {
+                // in case of a failed assertion print the stacktrace
+                System.err.println("Original stacktrace:");
+                e.printStackTrace();
```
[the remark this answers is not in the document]
mattulbrich: done.

key.core/src/test/java/de/uka/ilkd/key/speclang/njml/JMLParserExceptionTest.java
```diff
@@ -0,0 +1,199 @@
...
+                control, false, new Properties());
+            pl.setLoadSingleJavaFile(true);
+            pl.load();
+            // No exception encountered
+            assertEquals("true", props.get("noException"),
+                "Unless 'noException: true' has been set, an exception is expected");
+
+        } catch (Throwable e) {
+            if ("true".equals(props.getProperty("verbose"))) {
+                e.printStackTrace();
```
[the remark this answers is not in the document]
mattulbrich: done

key.core/src/test/java/de/uka/ilkd/key/speclang/njml/JMLParserExceptionTest.java
mattulbrich: I'd rather not rely on a particular layout of the data (i.e. the maven-layout with a resoureces folder). If there is such a directory in the classpath: great. If it is in a JAR file. ... does not work currently (and I do not want to implement it).
I'd argue that this is at least as flexible as your proposal, and error reporting is clear.

key.core/src/test/java/de/uka/ilkd/key/speclang/njml/JMLParserExceptionTest.java
mattulbrich: ok. good idea.

key.core/src/test/java/de/uka/ilkd/key/speclang/njml/JMLParserExceptionTest.java
mattulbrich: ok.

key.core/src/test/java/de/uka/ilkd/key/speclang/njml/JMLParserExceptionTest.java
```diff
@@ -0,0 +1,199 @@
...
+        assert fileURL != null : "Directory 'exceptional' not found";
+        assert fileURL.getProtocol().equals("file") : "Test resources must be in file system";
+        Path dir = Paths.get(fileURL.toURI());
+        return Files.list(dir).map(it -> Arguments.of(it, it.getFileName()));
+    }
+
+    @ParameterizedTest(name = "case {1}")
+    @MethodSource("getFiles")
+    public void testParseAndInterpret(Path file, Path localFilename) throws Exception {
+        List<String> lines = Files.readAllLines(file);
```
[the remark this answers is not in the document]
mattulbrich: I've converted it to a static method that can be invoked from elsewhere.

key.core/src/main/java/de/uka/ilkd/key/java/Recoder2KeY.java
mattulbrich: I'd like to delegate this to @jwiesler who changed this from `Position` to `Pair`.
This is merely a bugfix to avoid the NPE which would arise on almost all input with a syntax error.

key.core/src/test/java/de/uka/ilkd/key/speclang/njml/JMLParserExceptionTest.java
```diff
@@ -0,0 +1,153 @@
...
+            AbstractProblemLoader pl = new SingleThreadProblemLoader(file.toFile(), null, null,
+                null, AbstractProfile.getDefaultProfile(), false,
+                control, false, new Properties());
+            pl.setLoadSingleJavaFile(true);
+            pl.load();
+            // No exception encountered
+            assertEquals("true", props.get("noException"),
+                "Unless 'noException: true' has been set, an exception is expected");
+
+        } catch (Throwable e) {
```
wadoon, line 103: Maybe you should not catch `Throwable`. The logic gets quite complicated. 

JUnit 5 provides `assertThrows` to specify that a code block should throw an exception. https://howtodoinjava.com/junit5/expected-exception-example/

This only misses a handler for the exceptional case. The source code is [here](https://github.com/junit-team/junit5/blob/main/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertThrows.java#L49) and could be extended. 

You could also exclude the assertion case from `Throwable` by adding:

```java
try{...}
catch(AssertionFailure e){throw e;}
catch(Throwable e){ ... }
```

wadoon (approved):

key.core/src/test/java/de/uka/ilkd/key/speclang/njml/JMLParserExceptionTest.java
mattulbrich: right!

wadoon:
Interesting: 

```
 Check warning on line 1259 in key.core/src/main/java/de/uka/ilkd/key/java/Recoder2KeY.java

Codecov
/ codecov/patch
key.core/src/main/java/de/uka/ilkd/key/java/Recoder2KeY.java#L1259
Added line #L1259 was not covered by tests
```

PR #3154, opened 2023-06-05
Fix proof view filter not applied on load

### LINKED ISSUE #3151: Inconsistency of Proof Tree Settings Upon Loading a Proof
## Description

The proof tree settings that are displayed in the menu sometimes do not reflect the actual proof tree settings shown in the displayed proof tree. It seems that this occurs "only" upon loading a proof, until the specific settings option is first changed.

For example, when loading a proof, the option "Hide Intermediate Proofsteps" is ticked, even though it is not active. Unchecking the option the first time (correctly) results in (still) showing intermediate proof steps, then checking the option (again) results in hiding intermediate proof steps.

The observed settings inconsistency seems to occur upon each proof loading in the GUI, have not tried it for proof selection via the command line.

## Reproducible
  
Always.

### Steps to reproduce 

1. Start KeY.
2. Select "Sum and Max" from the examples dialogue.
3. Select "JML normal_behavior operation contract 0" and start proof.
4. Apply the "impRight" rule to the outermost implication.
5. Observe the proof tree and the checkmarks from its settings menu.

### Expected behavior
Hidden intermediate proof steps, as indicated by the checkmark in the proof-tree settings menu.

### Actual behavior
Unhidden intermediate proof steps, even though the checkmark indicated otherwise.
See also the attached screenshots.

### Additional information

**Checkmarks upon loading the proof:**
Checkmarks upon loading the proof

**Selection of some proof rule:**
Selection of some proof rule

**The resulting proof tree is inconsistent:**
Resulting proof tree is inconsistent

**See also the wrong checkmark:**
See also the wrong checkmark

---

* Commit: 58af713f232a9377f0eb5998d60ffcb65dfbf6f7 (current main branch)

FliegendeWurst:
Fixes #3151

### DIFF
--- a/key.ui/src/main/java/de/uka/ilkd/key/gui/prooftree/GUIProofTreeModel.java
+++ b/key.ui/src/main/java/de/uka/ilkd/key/gui/prooftree/GUIProofTreeModel.java
@@ -247,24 +247,12 @@ public class GUIProofTreeModel implements TreeModel, java.io.Serializable {
 
     /**
      * Set filters active or inactive and update tree if necessary.
-     *
-     * @param filter
-     * @param active
-     */
-    public void setFilter(ProofTreeViewFilter filter, boolean active) {
-        if (active != filter.isActive()) {
-            setFilterImmediately(filter, active);
-        }
-    }
-
-    /**
-     * Set filters active or inactive and update tree if necessary.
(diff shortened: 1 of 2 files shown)


mattulbrich (approved):
I stumbled over the same bug #3151 earlier today. This fixes it.

PR #3161, opened 2023-06-08
ADTs for KeY

wadoon:
Extension of the KeY grammar to allow the definition of ADTs. 

The grammar for ADTs follows the typical defintions in KeY. ADTs are defined within a `\datatypes { ... }` block, and are translated onto the existing infrastructure (Sorts, Functions, Taclets). The grammar is as follows:

```
\datatypes {
   <name> := <constructor> | ... | <constructor>;
   ...
 }
```

where constructor is a function definition `<cname>( <sort> <argName>, ...)`. 

Currently, three taclets are generated for each datatype definiton:

1. Axiom taclets that adds the induction as an axiom on the sequence 

$$
\frac{
(\forall a \in ADT.~ \phi) \leftrightarrow  \phi[a/c_1] \wedge \ldots \wedge \phi[a/c_n] 
\Longrightarrow
}{
\forall a \in ADT.~ \phi[a]
}
$$

2. Induction taclets for proving a sentence for all ADTs.

$$
\frac{
\begin{matrix}
   \forall a \in ADT; \phi \Longrightarrow \\
   \Longrightarrow \phi[a/c_1]  \\
   \Longrightarrow \ldots      \\
   \Longrightarrow phi[a/c_n]
\end{matrix}
}{
\Longrightarrow \forall a \in ADT.~ \phi[a]
}
$$

   A direct consequence of the induction formula in (1).

4. A case distinction taclet
  
   This is a direct consequence of the disjunction $x = c_1 \vee\ldots\vee x = c_n$ resulting from the construction principles of ADTs. This allows us to make a case distinction about the various constructors. Note, the constructor parameters are need to existential bounded. 

$$
\frac{
\begin{matrix}
   x = c_1      \Longrightarrow \\
   \ldots    \Longrightarrow    \\
   x = c_n   \Longrightarrow 
\end{matrix}
}{
     x : ADT ~ \text{(anywhere)}
}
$$

### Example

**Input**
```
\datatypes {
  MyList = MNil | MCons(any head, MyList tail);
}
```

**Generated Taclets**

```
MyList_Axiom {
    \schemaVar \formula phi;
    \schemaVar \variables MyList x;
    
    \find(\forall x; phi)
    \varcond(\notFreeIn(x, phi))
    \add(
             \forall x; phi
         <->   {\subst x;MNil}phi
             & \forall tail; \forall head; ({\subst x;tail}phi -> {\subst x;MCons(head, tail)}phi)
        ==>
      )
    \displayname "Axiom for MyList"
  }

MyList_Ind {
       \schemaVar \formula phi;
       
       "MNil":
           \add(==> {\subst x;MNil}phi);
       "MCons(anyhead,MyListtail)":
           \add(
               ==>
                \forall tail; \forall head; ({\subst x;tail}phi -> {\subst x;MCons(head, tail)}phi)
             );
       "Use case of MyList":
           \add(\forall x; phi ==>)
       \displayname "Induction for MyList"
     }

MyList_ctor_split {
          \schemaVar \term MyList var;
          \schemaVar \skolemTerm any head;
          \schemaVar \skolemTerm MyList tail;
          
          \find(var)
          \sameUpdateLevel
          "#var = MNil":
              \add(var = MNil ==>);
          "#var = MCons":
              \add(var = MCons(head, tail) ==>)
          \displayname "case distinction of MyList"
        }
```

### Current Problems / TODO 

* [x] Matching/Application of taclets not working 
* [x] Test cases


mattulbrich:
Any ideas how to deal with automation? Will there be default rule sets? Is everything manual?

wadoon:
> Any ideas how to deal with automation? Will there be default rule sets? Is everything manual?

First of all, I need to figure out, why I can not apply the taclets. 

Automation is a heuristic/ruleset assignment, for this experience is required.

wadoon:
> Any ideas how to deal with automation? Will there be default rule sets? Is everything manual?

First of all, I need to figure out, why I can not apply the taclets. 

Automation is a heuristic/ruleset assignment, for this experience is required.

WolframPfeifer:
> > Any ideas how to deal with automation? Will there be default rule sets? Is everything manual?
> 
> Automation is a heuristic/ruleset assignment, for this experience is required.
> 
> But first of all, I need to figure out, why I can not apply the taclets.

As a first step towards automation, we could try to assign one of the heuristics `userTaclets1..3`. Maybe this is enough already. However, even with the strategy option "High prior." it still has quite low priority, as for example class axiom rules are applied first. But maybe it helps for experimenting...

wadoon:
@mattulbrich For reviewing:

* The critical parts (taclet generation) are in `TacletPBuilder` (`createAxiomTaclet`, `createInductionTaclet`, `createConstructorSplit`). The rest is typical stuff, sorts and functions declaration. There is a fix in `KeYUserProblemFile` for avoiding double declaration. 

* Test case(s) missing. Would your ADT example fit here?

mattulbrich:
Thanks for getting this started. I have a few questions:
* In the inference rules above, should there not be sequent(s) above and below the line? What's the meaning if not? 
* I would expect that the constructors also take arguments that we probably have to quantify.
* Why is the first rule not a replacement?
* The first premise of the second rule was totally unexpected to me. Why does the original formula change sides?
* What's the quantifiers in 4.?

The induction scheme hints at these types being generated. How about them being free? There is already a keyword `\unique` in KeY to indicate that functions are free injections. Perhaps that should/could be re-used?

mattulbrich:
>     * Test case(s) missing. Would your ADT example fit?

The list example is nice handled using `\seq`. The tree example could benefit from a tree adt and functions on them. It would be more a JML-integration test however.

>     * **Unclear** I have a `\free` attribute. Is this only the `unique` attribute for functions?

I saw this after my comment above. I'd say yes.

wadoon:
> * In the inference rules above, should there not be sequent(s) above and below the line? What's the meaning if not?

Yes. My meaning is the same as `\find( f(x) )` (where f(x) is not in the succ or antec of a sequent). When I omit the sequent separation symbol `==>`, I want to say "anywhere" on the sequence, incl. non-toplevel formula. 

> * I would expect that the constructors also take arguments that we probably have to quantify.

Yes. And this is the tricky part to get the quantification/binding right. For a list, the axiom is 

$$
\forall x : List.~  \phi(x)   ~~\Leftrightarrow~~\phi(nil) \wedge \forall tail : List, val : any.~\phi(tail) \rightarrow \phi( cons(head, tail) )
$$

Anything needs to be derive from this axiom. For the splitting, the universal quantification becomes an existential quantification. 

> * Why is the first rule not a replacement?

It mimics the introduction of an axiom. This rule is not used in practise, it is more the base for the verification of the second and third rule.

> * The first premise of the second rule was totally unexpected to me. Why does the original formula change sides?

This is the "Use case". You find a similar structure in the KeY taclets with `int_induction`.

```
\rules {
    int_induction {
        "Base Case":
            \add( ==> {\subst nv; 0}(b));
        "Step Case":
            \add( ==> \forall nv; ((geq(nv, 0) & b) -> {\subst nv; (nv + 1)}b));
        "Use Case":
            \add(\forall nv; (geq(nv, 0) -> b) ==>)
    };
}

```


> * What's the quantifiers in 4.?

I guess existential. If you have a list `l`, it is either $l = nil$ or $l = cons(h,t)$ for some skolem constant $h$ and $t$.

> The induction scheme hints at these types being generated. How about them being free? There is already a keyword `\unique` in KeY to indicate that functions are free injections. Perhaps that should/could be re-used?

"\free" just survived from the CASL specification. I am not sure, if we have free ADTs here.

wadoon:
> > Test case(s) missing. Would your ADT example fit?
> 
> The list example is nice handled using `\seq`. The tree example could benefit from a tree adt and functions on them. It would be more a JML-integration test however.

Maybe I find something in the work with CASL. I did not find any possibility to prove a taclet, that is not written in a KeY file (i.e., dynamical generated).

mattulbrich:
Is this meant to be code-reviewed? Have all theoretical and design issues been discussed?

wadoon:
> Is this meant to be code-reviewed? Have all theoretical and design issues been discussed?

Unclear. I would have nothing to add to features. Minor is Heuristics of the Taclets. Major is missing tests. 

Do you need this feature urgently?

wadoon:
@mattulbrich @WolframPfeifer 

We have an error in saving/loading the proofs. I received on the stdout following exception:

```
Caused by: de.uka.ilkd.key.util.parsing.BuildingException: Malformed origin string: "User_Interaction @ node 0 (Induction for List)"
(Well-formed origins have one of the following formats: "spec_type @ file <file name> @ line <line number>")
spec_type @ node <node number> (<rule name>)")
spec_type (implicit)")
```
The proof files were created with KeY. 

I was able to point this to spaces inside of `\displayname` of Taclets. 
In my opinion the `\displayname` is for the human and should allowed to contain space characters. Maybe, it would be better to use the name of the taclet which can not contain a space if it was parsed from a file. And it would also not be ambigue. 

For this PR I have fixed it manually, but I would not ship a feature which breaks loading of a proofs if induction on ADTs used.

wadoon:
I avoid spaces for know. Issue is created #3286.

mattulbrich:
image

Would it perhaps be advantageous to change the rule to
`#var = Red | #var = Green | #var = Blue ==>` such that the automation can decide whether it wants to split or not?

mattulbrich:
Should the following not be provable:
```
\datatypes { \free Color = Red | Green | Blue; }
\problem { Red != Green }
```

mattulbrich (approved):

WolframPfeifer:
> Should the following not be provable:
> 
> ```
> \datatypes { \free Color = Red | Green | Blue; }
> \problem { Red != Green }
> ```
> 
> I would have assumed that the construtors of free datatypes are unique functions

I would also think so. In the example linked in the [documentation](https://keyproject.github.io/key-docs/user/ADTs/), the section "Algebraic Data Types (before KeY-2.14.0)" contains also these \unique modifiers for the functions ...

I vote for removing the "\free" keyword and making the data types by default generated (i.e., for each element of the type there is a sequence of constructors that creates it) and free (i.e., for each element of the type there is a unique sequence of constructors that creates it). If necessary, it would still be possible to create non-free/non-generated ADTs by manually defining the functions and axioms, but I think that is not the main use case.

By the way, I think in Isabelle data types are free and generated; you can prove the following theorems:
```
datatype Color = Red | Green | Blue
theorem "Red ≠ Green" by simp
theorem "∀ c::Color. c ≠ Red ∨ c ≠ Green ∨ c ≠ Blue" using Color.exhaust by blast
```

(I hope that I did not confuse the terms "free" and "generated" here, I find them quite confusing and non-intuitive ...)

PR #3166, opened 2023-06-10
fix StringUtil.trim

### LINKED ISSUE #3165: StringUtil.trim deletes too much
## Description

When specifying `\javaSource ".";`, an exception may be thrown depending on the current working directory.

The underlying problem is that the string trimming function is broken.

## Reproducible
  
* *Original:* sometimes. For the original problem to occur, several factors must come together.
* *Trim function:* always.

### Steps to reproduce 

Apply the following patch to add test cases for the trim function that currently fail.

```diff
--- a/key.util/src/test/java/org/key_project/util/testcase/java/StringUtilTest.java
+++ b/key.util/src/test/java/org/key_project/util/testcase/java/StringUtilTest.java
@@ -288,6 +288,7 @@ public class StringUtilTest {
         assertEquals("", trim(
             "   \n\t\fa234231hsdafhvnyxcksdaökfhsdaöfhsahövcln231847231 42310897423187sdfsdafbc\n",
             it -> true));
+        assertEquals(".", trim("\".\"", '"'));
+        assertEquals(".", trim(".", '"'));
+        assertEquals(".", trim(".", it -> false));
         assertEquals("", trim("", 'c'));
     }
 }
```

They should not fail.

* Commit: f66ad2af25d64b601065c177e12159b435c26982

wadoon:
Just a little index-of-by-one thing. 

Fixes  #3165

### DIFF
--- a/key.util/src/test/java/org/key_project/util/testcase/java/StringUtilTest.java
+++ b/key.util/src/test/java/org/key_project/util/testcase/java/StringUtilTest.java
@@ -289,5 +289,9 @@ public class StringUtilTest {
             "   \n\t\fa234231hsdafhvnyxcksdaökfhsdaöfhsahövcln231847231 42310897423187sdfsdafbc\n",
             it -> true));
         assertEquals("", trim("", 'c'));
+
+        assertEquals(".", trim("\".\"", '"'));
+        assertEquals(".", trim(".", '"'));
+        assertEquals(".", trim(".", it -> false));
     }
 }
--- a/key.util/src/main/java/org/key_project/util/java/StringUtil.java
+++ b/key.util/src/main/java/org/key_project/util/java/StringUtil.java
@@ -400,7 +400,7 @@ public final class StringUtil {
         while (first <= last && predicate.test(value[last])) {
             --last;
         }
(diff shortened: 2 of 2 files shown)


key.util/src/test/java/org/key_project/util/testcase/java/StringUtilTest.java
```diff
@@ -136,8 +135,8 @@ public void testEqualIgnoreWhiteSpace() {
         assertFalse(StringUtil.equalIgnoreWhiteSpace("A B C D", "A B C"));
         assertTrue(StringUtil.equalIgnoreWhiteSpace("  A B C", "A B C\t\n"));
         assertTrue(StringUtil.equalIgnoreWhiteSpace(
-            "{result=self.doubleValue(_value)@ExistingContractTest; }",
-            "{\n  result=self.doubleValue(_value)@ExistingContractTest;\n}"));
+                "{result=self.doubleValue(_value)@ExistingContractTest; }",
+                "{\n  result=self.doubleValue(_value)@ExistingContractTest;\n}"));
```
FliegendeWurst, line 139: Unnecessary formatting changes.

mattulbrich (approved):

mattulbrich:
I added test cases, yet the coverage decreased ...

Issue #3170, opened 2023-06-13
CITATION.cff


wadoon:
Github supports to point out a specific citation for a project. This appears in the right sidebar as "Cite this repository". 

The description of the format of `CITATION.cff` is given here: 

https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files

It would be beneficial to point to the second KeY book. It seems that there is no nice translation tool for this. So someone needs to rewrite this entry

```bibtex
@book{DBLP:series/lncs/10001,
  editor       = {Wolfgang Ahrendt and
                  Bernhard Beckert and
                  Richard Bubel and
                  Reiner H{\"{a}}hnle and
                  Peter H. Schmitt and
                  Mattias Ulbrich},
  title        = {Deductive Software Verification - The KeY Book - From Theory to Practice},
  series       = {Lecture Notes in Computer Science},
  volume       = {10001},
  publisher    = {Springer},
  year         = {2016},
  url          = {https://doi.org/10.1007/978-3-319-49812-6},
  doi          = {10.1007/978-3-319-49812-6},
  isbn         = {978-3-319-49811-9},
  timestamp    = {Tue, 14 May 2019 10:00:55 +0200},
  biburl       = {https://dblp.org/rec/series/lncs/10001.bib},
  bibsource    = {dblp computer science bibliography, https://dblp.org}
}
```

to this format:

```yaml
cff-version: 1.2.0
message: "If you use this software, please cite it as below."
authors:
- family-names: "Lisa"
  given-names: "Mona"
  orcid: "https://orcid.org/0000-0000-0000-0000"
- family-names: "Bot"
  given-names: "Hew"
  orcid: "https://orcid.org/0000-0000-0000-0000"
title: "My Research Software"
version: 2.0.4
doi: 10.5281/zenodo.1234
date-released: 2017-12-18
url: "https://github.com/github-linguist/linguist"
preferred-citation:
  type: article
  authors:
  - family-names: "Lisa"
    given-names: "Mona"
    orcid: "https://orcid.org/0000-0000-0000-0000"
  - family-names: "Bot"
    given-names: "Hew"
    orcid: "https://orcid.org/0000-0000-0000-0000"
  doi: "10.0000/00000"
  journal: "Journal Title"
  month: 9
  start: 1 # First page number
  end: 10 # Last page number
  title: "My awesome research software"
  issue: 1
  volume: 1
  year: 2021
```

[...]

PR #3171, opened 2023-06-14
added CITATION.bib

mattulbrich:
In answer to #3170: Adding a bibtex file works as well. See also 

https://github.com/mattulbrich/key

image


wadoon (approved):

wadoon:
But a `CITATION.cff` is more beautiful.

PR #3172, opened 2023-06-15
Translation of bitwise operators

ChristianHein:
- Adds missing taclets for intermediate JavaDL functions `unsignedshiftrightJlong`, `xorJlong`, `orJlong`, and `andJlong`. Without these, one cannot use the Java operators `>>>`, `^`, `|`, and `&` with parameters of type `long`.
- Fixes a bug in the taclets which pull a bitwise negation out of a modality, making them unusable. This was caused by a trivial mismatch between the program variables referenced in the taclets' `\find` and `\replacewith` statements. This bug was not present in 208396f69e.


FliegendeWurst:
You will need to update `taclets.old.txt` as described in `TestTacletEquality.java`.

PR #3178, opened 2023-06-22
Add context menu to proofs in task tree

FliegendeWurst:
This PR adds a context menu to proof entries in the task tree. Entries are equal to the proof menu in the menu bar.

### DIFF
--- a/key.ui/src/main/java/de/uka/ilkd/key/gui/MainWindow.java
+++ b/key.ui/src/main/java/de/uka/ilkd/key/gui/MainWindow.java
@@ -831,7 +831,7 @@ public final class MainWindow extends JFrame {
         JMenuBar menuBar = new JMenuBar();
         menuBar.add(createFileMenu());
         menuBar.add(createViewMenu());
-        menuBar.add(createProofMenu());
+        menuBar.add(createProofMenu(null));
         menuBar.add(createOptionsMenu());
         KeYGuiExtensionFacade.addExtensionsToMainMenu(this, menuBar);
         menuBar.add(Box.createHorizontalGlue());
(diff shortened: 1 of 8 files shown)


FliegendeWurst:
Fixed.

mattulbrich (changes requested):
I may be picky, but now I observe that when I right-click on a proof and then cancel the popup menu, the selected item in the proof selection panel and the actually selected proof do not coincide. Missing selection listener?

FliegendeWurst:
> I may be picky, but now I observe that when I right-click on a proof and then cancel the popup menu, the selected item in the proof selection panel and the actually selected proof do not coincide. Missing selection listener?

Most applications do not restore the selection after closing the context menu. I'll fix it anyways.

key.ui/src/main/java/de/uka/ilkd/key/gui/TaskTree.java
```diff
@@ -73,10 +75,77 @@ public TaskTree(KeYMediator mediator) {
         delegateView.setShowsRootHandles(false);
         delegateView.setRootVisible(false);
         delegateView.putClientProperty("JTree.lineStyle", "Horizontal");
+
+        // create a context menu on demand
+        delegateView.addMouseListener(this);
```
WolframPfeifer, line 80: There already is another mouse listener (see field "mouseListener"). Is it necessary to have two? Can they be merged? In addition, I think it is cleaner to use inner classes instead of having TaskTree implement the MouseListener/PopupMenuListener interfaces.

key.ui/src/main/java/de/uka/ilkd/key/gui/MainWindow.java
```diff
@@ -915,40 +915,50 @@ private JMenu createSelectionMenu() {
         return goalSelection;
     }
 
-    private JMenu createProofMenu() {
+    /**
+     * Create the proof menu.
+     *
+     * @param proof a specific proof that the menu should work on, may be null
```
WolframPfeifer, line 921: JavaDoc parameter name is already out of sync...

WolframPfeifer (changes requested):
There are situations where the currently selected proof is different than the one used for the context menu. This seems to happen always when I open the menu without selecting (by left click) the proof first.
grafik

I think it would be reasonable (and should be easy) to select the correct proof first before showing it's context menu. Or are there downsides?

WolframPfeifer:
Under Ubuntu, it looks fine for me. I will see if it works on my Windows machine tomorrow.

key.ui/src/main/java/de/uka/ilkd/key/gui/TaskTree.java
FliegendeWurst: This is done.

key.ui/src/main/java/de/uka/ilkd/key/gui/MainWindow.java
FliegendeWurst: Also done

mattulbrich (approved):

key.ui/src/main/java/de/uka/ilkd/key/gui/TaskTree.java
```diff
@@ -253,10 +249,13 @@ public void mousePressed(MouseEvent e) {
 
         private void checkPopup(MouseEvent e) {
```
WolframPfeifer, line 250:
```suggestion0
        @Override
        public void mouseReleased(MouseEvent e) {
            checkPopup(e);
        }

        /**
         * Checks whether the popup menu should be shown and does so if necessary.
         * <br>
         * <b>Important:</b><br>
         * For the platform specific popup trigger to work, we need to check the popup in pressed, released, and clicked
         * event. For example, on Windows the e.isPopupTrigger() information is only available in the released event.
         *
         * @param e the mouse event that may create the popup
         */
         private void checkPopup(MouseEvent e) {
```
Without this code, the popup menu does not work on Windows (isPopupTrigger() call only returns true in mouseReleased!). Can you add this please?

WolframPfeifer (changes requested):
I had a problem where the popup menu does not appear on Windows (my code suggestion fixes that). Apart from that, the code works fine. I think that the new behavior (selecting the proof when opening the popup) is consistent to user expectations.

WolframPfeifer (approved):

PR #3179, opened 2023-06-22
Make settings dialog look nicer

### LINKED ISSUE #1645: Expand Entries in Settings Dialog (Left Column)
This issue was created at [git.key-project.org](https://git.key-project.org/key/key/-/issues/1645) where the discussions are preserved.
    
----
    
 

## Description

Expanding or collapsing entries in the left column of the settings dialog requires a double click even if entry is selected.

Expected Behaviour: Clicking on a selected entry should expand/collapse the entry

## Reproducible
  
always (macOS 15.1)

### Steps to reproduce 

1. Start KeY
2. Open Settings Dialog (Options > Show Settings)
3. Select "Appearance and Behaviour" (one click)
3. Will not collapse / expand afterwards by just one click, requires double click

Expected Behaviour: Clicking on a selected entry should expand/collapse the entry and not require a double click

### Additional information

---

* Commit: 

57a19c0ce6fb08e9dcf183eb4116afbc72d85b05

    
----
    
Information: 
* created_at: 2021-12-18T14:30:29.261Z
* updated_at: 2021-12-18T14:30:29.261Z
* milestone: 
* user_notes_count: 0

FliegendeWurst:
Changes done:
- automatic width calculation based on largest settings panel
- less padding next to the help icon
- fix for #1645

Feel free to comment on the new look:
image

### DIFF
--- a/key.ui/src/main/java/de/uka/ilkd/key/gui/settings/SettingsManager.java
+++ b/key.ui/src/main/java/de/uka/ilkd/key/gui/settings/SettingsManager.java
@@ -120,7 +120,6 @@ public class SettingsManager {
         dialog.setSettingsProvider(settingsProviders);
         dialog.setDefaultCloseOperation(WindowConstants.DISPOSE_ON_CLOSE);
         dialog.setIconImage(IconFactory.keyLogo());
-        dialog.setSize(800, 600);
         dialog.setLocationByPlatform(true);
         return dialog;
     }
--- a/key.ui/src/main/java/de/uka/ilkd/key/gui/actions/MinimizeInteraction.java
+++ b/key.ui/src/main/java/de/uka/ilkd/key/gui/actions/MinimizeInteraction.java
@@ -7,15 +7,14 @@ import java.util.EventObject;
 import de.uka.ilkd.key.gui.MainWindow;
 import de.uka.ilkd.key.settings.ProofIndependentSettings;
 
-/*
- * Is this a legacy option? Finding instantiations seems to be done by the prover, even if this
- * option is disabled. (Kai Wallisch, December 2013)
+/**
+ * Option to specify all instantiations manually (does not apply to the automatic solver).
  */
 public class MinimizeInteraction extends KeYMenuCheckBox {
     public static final String NAME = "Minimize Interaction";
     public static final String TOOL_TIP =
-        "If ticked and automated strategy (play button) is used, the prover tries to minimize user interaction, "
-            + "e.g., if the prover can find instantiations by itself, it will not ask the user to provide them.";
+        "If not ticked, applying a taclet manually will require you to instantiate " +
+            "all schema variables.";
 
     /**
      *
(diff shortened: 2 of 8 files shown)


mattulbrich:
Mmmmh. This looks pretty identical to 2.10 to me? Can you point out the differences please? (right is 2.10, left is HEAD of this branch)

image

mattulbrich:
And on my system, the SMT settings are broken UI-wise (why are the horizontal schroll bars needed?

image

FliegendeWurst:
The only difference is the look of the icons in the selection menu on the left. As for the SMT settings, I did not touch them in this PR. But I'll push a change to increase the width of the settings dialog.

key.ui/src/main/java/de/uka/ilkd/key/gui/settings/SettingsManager.java
```diff
@@ -120,7 +120,6 @@ private SettingsDialog createDialog(MainWindow mainWindow) {
         dialog.setSettingsProvider(settingsProviders);
         dialog.setDefaultCloseOperation(WindowConstants.DISPOSE_ON_CLOSE);
         dialog.setIconImage(IconFactory.keyLogo());
-        dialog.setSize(800, 600);
         dialog.setLocationByPlatform(true);
```
WolframPfeifer, line 123:
```suggestion
```
I think we should delete this. It can lead to non-centered dialog (occurred on my windows machine), which might even be on another monitor than the main window.

WolframPfeifer (changes requested):
I like the idea of streamlining the icons. Size and position of the dialog are still not optimal. On my Windows machine, while the tree is fine, the SMT tab (and some others) by default is still a bit to narrow:
grafik
@mattulbrich How does it look on your machine now?

And I think we should center the dialog, as the OS specific location may be strange (see my suggestion). Btw., there are other dialogs with the same problem, e.g. the ProgressDialog.

key.ui/src/main/java/de/uka/ilkd/key/gui/settings/SettingsManager.java
FliegendeWurst: I think it would be best to remove all setLocationByPlatform calls at once in another PR since that will require more testing.

WolframPfeifer: Ok, let's move this to a separate PR, as it also affects some other dialogs.

WolframPfeifer:
Under Ubuntu, settings layout looks fine for me. I will see if it works on my Windows machine tomorrow.

WolframPfeifer:
On my Windows machine, it looks mostly fine now. However, since the info text for the PRINCESS solver is quite long, the dialog looks like this now:
grafik

I think it is a general problem, maybe we should use a multiline textbox for the (potentially) long text in the info field ...

FliegendeWurst:
I have changed the info to a text area and fixed the width calculation to also consider the divider.

WolframPfeifer (approved):

WolframPfeifer:
> I have changed the info to a text area and fixed the width calculation to also consider the divider.

This solves the problem also on my Windows. Thanks!

Issue #1708, opened 2022-12-23
Partial class invariants have non-local naming scheme


wadoon:
This issue was created at [git.key-project.org](https://git.key-project.org/key/key/-/issues/1708) where the discussions are preserved.
    
----
    
Title. This is very bad because it breaks proofs in unrelated files (numbering changes, proof that uses partial invariant axiom doesn't find it any more).

## Reproducible
Load the proof for `B::test` in this archive [key-test.7z](/uploads/ceb1062ed4e7f728de87e27078cc4758/key-test.7z). See that it succeeds (it uses the partial invariant of `B` only for demonstration purposes). Comment out the second class invariant of `A` (the other class). Load the proof for `B::test` again. See that it fails to load.

## Proposed solution
Make the numbering per class.

Dream solution: Make them nameable so they can be reordered without breaking proofs.
    
----
    
Information: 
* created_at: 2022-04-10T18:13:29.665Z
* updated_at: 2022-04-10T20:16:10.317Z
* closed_at: None  (closed_by: )
* milestone: 
* user_notes_count: 2

PR #3184, opened 2023-06-26
towards fixing #1708

mattulbrich:
Numbering of JML invariants made different for each class to allow for reloading of proofs if invariants are changed in other classes.

@gewitternacht, perhaps you can give this a try?

### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/speclang/jml/translation/JMLSpecFactory.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/speclang/jml/translation/JMLSpecFactory.java
@@ -65,7 +65,7 @@ public class JMLSpecFactory {
     protected final de.uka.ilkd.key.logic.TermBuilder tb;
     protected final de.uka.ilkd.key.java.Services services;
     protected final ContractFactory cf;
-    private int invCounter;
+    private final Map<KeYJavaType, Integer> invCounter = new IdentityHashMap<>();
     /**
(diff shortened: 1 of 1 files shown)


WolframPfeifer:
I made two small changes:

1. Use fully qualified name to avoid possible conflicts between classes with equal name in different packages.
2. Change order: first the (qualified) name, then the number. This leads to a better grouping of the rules in the "Info" tab (see screenshots).

| main | this PR|
|---------|-----------|
|screenshot_main_partial_inv_axioms|screenshot_fix_partial_inv_axioms|

mattulbrich:
> I made two small changes:
> 
>     1. Use fully qualified name to avoid possible conflicts between classes with equal name in different packages.

Thanks for this. I wonder if the taclet names are now not unbearably long?

WolframPfeifer:
> > I made two small changes:
> > 
> > 1. Use fully qualified name to avoid possible conflicts between classes with equal name in different packages.
> 
> Thanks for this. I wonder if the taclet names are now not unbearably long?

What does "unberable" mean here? Yes, they are a bit longer now, but from my perspective it is necessary to use fully qualified names to be robust against newly added classes. If we need shorter names, we could maybe shorten the prefix "Partial_inv_axiom_for_JML_class_invariant_in_" a bit. Also, we could set shorter displaynames. However, I do not see where long names are a problem right now ...

wadoon (approved):

Issue #3185, opened 2023-06-27
Free invariants and assignable clauses


flo2702:
## Please describe your proposal in a ONE sentence

There should be an `invariant_free` keyword and an `assignable_free` keyword to go with the existing `requires_free` and `ensures_free`.

## Underlying problem

While we can express free pre- and postconditions (as well as `assume` statements) in KeY, expressing free invariants and assignable clauses requires manually desugaring them into free pre- and postconditions, which is onerous and violates modularity.

## Additional context

Free invariants can be implemented by adding a new JML function `\invariant_free_for(obj)` to get the free invariant, while `\invariant_for(obj)` continutes to only get the non-free invariant.
In proofs, a new symbol `self.<inv_free>` can be added to every free precondition and to every free postcondition. This can be expanded into the free invariant, just like `self.<inv>` can be expanded into the non-free invariant.
While this is in my opinion the most straightforward way to implement this, it also clutters up the sequent a bit, so I'm open to alternatives.

For free assignable clauses, @mattulbrich and me previously (https://git.key-project.org/key/key/-/merge_requests/546#63b309e61e235f6869864a89f960c27a2e5de96f) discussed what their semantics should be and came to the following conclusion: A method with clauses `assignable f; assignable_free g;` must prove that it only modifies `{f,g}`. Its callers may assume that it only modifies `f`. Thus, a method with clauses `assignable \nothing; assignable_free g;` may modify `g` but its callers may assume it is pure. The default value if no explicit `assignable_free` clause is given should be `\strictly_nothing`.

PR #3187, opened 2023-06-28
 Invariant free

flo2702:
This branch implements `invariant_free` clauses and the `\invariant_free_for` keyword. See #3185

# Usage example

All contracts in the following listing are provable.

```
final class A { 
    int f;
    int g;
    //@ invariant f > 0;
    //@ invariant_free g > 0;
    
    /*@ normal_behavior
      @ ensures true;
      @*/
    A() {
        f = 1;
        g = 0;
    }
    
    /*@ normal_behavior
      @ ensures \result > 0;
      @*/
    int foo() {
        return g;
    }
}

final class B {
    /*@ normal_behavior
      @ ensures \result > 0;
      @*/
    int foo() {
        return new A().g;
    }
    
    /*@ normal_behavior
      @ requires \invariant_free_for(a);
      @ ensures \result > 0;
      @*/
    int bar(A a) {
        return a.g;
    }
}
```

# Review

* `\invariant_free_for(obj)` gets the free invariant, while `\invariant_for(obj)` only gets the non-free invariant.
* In proofs, a new symbol `self.<inv_free>` is added to every free precondition and to every free postcondition. This can be expanded into the free invariant, just like `self.<inv>` can be expanded into the non-free invariant. If the free invariant is `true`, this symbol is not added to avoid cluttering the sequent and for backward compatibility.


mattulbrich (approved):

jwiesler:
I've already used this, but I find it very weird that I have to require the free invariant in method contracts. Why should I need to do that? This holds for any object of the type (independant of heap) and I also never have to (and never actually can) show the free invariant.

jwiesler:
Furthermore, what this was used for was to prove some program with non free invariants, then make the invariants free.

However, this transformation is not sound if invariant_free **always** holds. Especially, when proving functions in the class itself, this can lead to massive problems: Free invariant says `i = 0` but function changes value to `i = 1` and back to `i = 0`. In this case, it should only hold when the normal invariant holds as well.

flo2702:
> I've already used this, but I find it very weird that I have to require the free invariant in method contracts. Why should I need to do that? This holds for any object of the type (independant of heap) and I also never have to (and never actually can) show the free invariant.

Unless I'm misunderstanding you, you don't have to require the free invariant. In the example above, the contract `A::foo` is provable because the free invariant is assumed as part of the free precondition.

> However, this transformation is not sound if invariant_free always holds. Especially, when proving functions in the class itself, this can lead to massive problems: Free invariant says i = 0 but function changes value to i = 1 and back to i = 0. In this case, it should only hold when the normal invariant holds as well.

Again, I'm not sure if I'm understanding you, but the free invariant acts like a free precondition (i.e., the callee can assume it holds in its prestate + a free postcondition, i.e., the caller can assume it holds after the callee returns). This is the same behavior as for non-free invariants, except that the free invariant never has to be shown.

PR #3188, opened 2023-06-28
Assignable free

flo2702:
This branch implements `assignable_free` clauses, which akin to `ensures_free` allow you to specify frame conditions that the caller can assume but the callee does not have to prove. See #3185.

`assignable_free` clauses are implemented for functional method contracts, functional block/loop contracts, and loop invariants.

# Semantics

Consider the following example.

```
/*@ public normal_behavior
  @ ensures field == 21; */
void foo() {
  field = 21;
  bar();
}

/*@ public normal_behavior
  @ ensures true; 
  @ assignable \nothing;
  @ assignable_free field; */
void bar() {
  field = 42;
}
```

Both contracts are provable: The proof obligation for `bar()` must show that only `field` is modified, but then when `bar()` is called from `foo()`, `foo()` assumes that `bar()` modifies `\nothing`.

In general, the caller assumes that the callee only modifies the locations mentioned in its `assignable` clause(s), while the callee must prove that it only modifies the locations in **the union of its `assignable` and `assignable_free` clauses**.


key.core/src/main/antlr4/JmlLexer.g4
```diff
@@ -78,7 +78,7 @@ fragment Pfree: '_free'?;       //suffix
 ACCESSIBLE: 'accessible' Pred -> pushMode(expr);
 ASSERT: 'assert' Pred  -> pushMode(expr);
 ASSUME: 'assume' Pred -> pushMode(expr);
-ASSIGNABLE: 'assignable' Pred -> pushMode(expr);
+ASSIGNABLE: ('assignable' Pfree) -> pushMode(expr);
```
mattulbrich, line 81: why the parentheses?

mattulbrich (approved):

mattulbrich:
Rather a large addition to the code. Thanks for your efforts.

PR #3192, opened 2023-07-01
Fix slicing bug with dependency contracts

FliegendeWurst:
This PR fixes a very rare bug of the slicing replay mechanism that @WolframPfeifer reported to me.


WolframPfeifer (approved):
It is obscure what the "step" field is used for. Could this be necessary for permissions, where you have multiple heaps (@mattulbrich)? However, as far as I see, this PR does nothing that could break things and fixes the original problem where slicing failed.

PR #3193, opened 2023-07-02
Allow selection of sub-nodes in OSS steps

FliegendeWurst:
This change allows the user to select the individual steps done inside a One Step Simplification. The selection model is extended to include a currently shown sequent and rule app.

### DIFF
--- a/key.ui/src/main/java/de/uka/ilkd/key/gui/nodeviews/InnerNodeView.java
+++ b/key.ui/src/main/java/de/uka/ilkd/key/gui/nodeviews/InnerNodeView.java
@@ -1,6 +1,7 @@
 package de.uka.ilkd.key.gui.nodeviews;
 
 import java.awt.*;
+import java.awt.geom.Rectangle2D;
 import javax.swing.*;
 import javax.swing.border.CompoundBorder;
 import javax.swing.border.EmptyBorder;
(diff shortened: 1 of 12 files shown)


WolframPfeifer:
That is a long requested feature. In a first quick test, it seems to work very nicely, thank you very much! I have two related feature requests:
1. When "Show taclet info" is activated, it would be nice to have the concrete taclet shown in the info area instead of "One Step Simplification".
2. I noticed during my quick test that the SequentFormula the rule is applied to is not always visible after switching nodes. Would it be possible to scroll to the "active" formula by default? This would be the formula matched by the "find" part of the taclet, for NoPosTaclets we can still try to find a formula matched by "assumes". If none of these exists, we could default to the current behavior (no scrolling).

key.ui/src/main/java/de/uka/ilkd/key/gui/nodeviews/InnerNodeView.java
```diff
@@ -149,6 +157,28 @@ private Range highlightPos(PosInOccurrence pos, HighlightPainter light)
...
+                    Range rFormula = posTable.rangeForPath(pathTop);
+                    setCaretPosition(rFormula.start());
+                    Rectangle2D rect = modelToView2D(rFormula.start() + 2);
+                    Rectangle visible = getVisibleRect();
+                    if (rect != null && visible != null
+                            && !visible.contains(rect.getMinX(), rect.getMinY())) {
+                        MainWindow.getInstance().scrollTo((int) rect.getMinY());
+                    }
+                } catch (BadLocationException e) {
+                    // ignore, should never happen
```
WolframPfeifer, line 178: I also would hope so, but maybe we should log at least ...

WolframPfeifer (approved):
Apart from my comment about the exception, this PR looks fine to me. However, I still would like to have another opinion on this (@mattulbrich).

mattulbrich:
Nice catch indeded using that the rule app contains the position in occurrence which contains a reference to the sequent!
Thanks!

key.ui/src/main/java/de/uka/ilkd/key/gui/nodeviews/InnerNodeView.java
FliegendeWurst: Done

PR #3195, opened 2023-07-06
Set statement: Check assignee for validity

jwiesler:
Replaces set statements as copy assignments with proper set statements that do not have an Java-AST for the assigned value.

Pretty much copy-pastes what JML asserts do.


key.core/src/main/java/de/uka/ilkd/key/speclang/jml/translation/JMLSpecFactory.java
```diff
@@ -1347,19 +1344,76 @@ public void translateJmlAssertCondition(final JmlAssert jmlAssert, final IProgra
...
+            var variable = (LocationVariable) assignee.op();
+            if (variable.isGhost()) {
+                return null;
+            } else {
+                return variable + " is not a ghost variable";
+            }
+        } else if (services.getTypeConverter().getHeapLDT().isSelectOp(assignee.op())) {
+            var field = assignee.subs().last();
+            var op = field.op();
+            var split = HeapLDT.trySplitFieldName(op);
```
mattulbrich, line 1377: type? pls make explicit

key.core/src/main/java/de/uka/ilkd/key/speclang/jml/translation/JMLSpecFactory.java
```diff
@@ -1347,19 +1344,76 @@ public void translateJmlAssertCondition(final JmlAssert jmlAssert, final IProgra
...
+            if (split != null) {
+                var attribute =
```
mattulbrich, line 1379: type? pls make explicit

key.core/src/main/java/de/uka/ilkd/key/speclang/jml/translation/JMLSpecFactory.java
```diff
@@ -1347,19 +1344,76 @@ public void translateJmlAssertCondition(final JmlAssert jmlAssert, final IProgra
...
+    /**
+     * Translates a set statement.
+     *
+     * @param statement the set statement
+     * @param pm the enclosing method
+     */
+    public void translateSetStatement(final SetStatement statement, final IProgramMethod pm)
+            throws SLTranslationException {
+        final var pv = createProgramVariablesForStatement(statement, pm);
+        var setStatementContext = statement.takeParserContext();
```
mattulbrich, line 1405: type? pls make explicit

key.core/src/main/java/de/uka/ilkd/key/speclang/jml/translation/JMLSpecFactory.java
```diff
@@ -1347,19 +1344,76 @@ public void translateJmlAssertCondition(final JmlAssert jmlAssert, final IProgra
...
+        var io = new JmlIO(services).context(Context.inMethod(pm, tb)).selfVar(pv.selfVar)
+                .parameters(pv.paramVars)
+                .resultVariable(pv.resultVar).exceptionVariable(pv.excVar).atPres(pv.atPres)
+                .atBefore(pv.atBefores);
+        var assignee = io.translateTerm(setStatementContext.assignee);
```
mattulbrich, line 1410: type? pls make explicit

key.core/src/test/java/de/uka/ilkd/key/speclang/SetStatementTest.java
```diff
@@ -0,0 +1,116 @@
...
+
+    private String parseAndCheck(String statementText) {
+        JMLSpecFactory jsf = new JMLSpecFactory(services);
+        ImmutableList<TextualJMLConstruct> constructs =
+            new PreParser().parseMethodLevel(statementText, null, Position.newOneBased(1, 1));
+        assertEquals(constructs.size(), 1);
+        assertInstanceOf(TextualJMLSetStatement.class, constructs.head());
+        var statement = (TextualJMLSetStatement) constructs.head();
+        var context = statement.getAssignment();
+        var assignee = jmlIO.translateTerm(context.assignee);
```
mattulbrich, line 113: type? pls make explicit

key.core/src/test/java/de/uka/ilkd/key/speclang/SetStatementTest.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    private String parseAndCheck(String statementText) {
+        JMLSpecFactory jsf = new JMLSpecFactory(services);
+        ImmutableList<TextualJMLConstruct> constructs =
+            new PreParser().parseMethodLevel(statementText, null, Position.newOneBased(1, 1));
+        assertEquals(constructs.size(), 1);
+        assertInstanceOf(TextualJMLSetStatement.class, constructs.head());
+        var statement = (TextualJMLSetStatement) constructs.head();
+        var context = statement.getAssignment();
```
mattulbrich, line 112: type? pls make explicit. Ist this a term?

mattulbrich (changes requested):

mattulbrich:
This PR seems to replace set statements as copy assignments with proper set statements that do not have an Java-AST for the assigned value.

If that is so, that would be great and expressions can be supported in set statements like quantifiers etc.
We can then probably remove all the special java expression constructs for JML-artifacts, right?

jwiesler:
> We can then probably remove all the special java expression constructs for JML-artifacts, right?

The Recoder removal branch does this except for set statements.

key.core/src/main/java/de/uka/ilkd/key/java/Recoder2KeYConverter.java
```diff
@@ -1117,8 +1118,12 @@ public VariableSpecification convert(
 
             final ProgramElementName name =
                 VariableNamer.parseName(makeAdmissibleName(recoderVarSpec.getName()));
+            var isGhost = containsModifier(recoderVarSpec.getParent(),
```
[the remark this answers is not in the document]
jwiesler: Yes and they IMO follow what the "guide" page says: The type is reflected in both the variable name ("is") and the function ("contains").

key.core/src/test/java/de/uka/ilkd/key/speclang/SetStatementTest.java
jwiesler: It's Term and I think that's obvious from the context.

key.core/src/main/java/de/uka/ilkd/key/speclang/jml/translation/JMLSpecFactory.java
jwiesler: I'm reluctant in adding these types since they are the generated types coming from the JmlParser and their names convey almost no meaning.

key.core/src/main/java/de/uka/ilkd/key/java/Recoder2KeYConverter.java
```diff
@@ -1220,6 +1225,18 @@ public FieldSpecification convert(recoder.java.declaration.FieldSpecification re
         return varSpec;
     }
 
+    private static boolean containsModifier(recoder.java.declaration.JavaDeclaration fs,
+            Class<? extends DeclarationSpecifier> cls) {
+        var specifiers = fs.getDeclarationSpecifiers();
```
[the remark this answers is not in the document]
jwiesler: Same here, the type is `ASTList<DeclarationSpecifier>` and I think this follows the style page as well which allows eliding `List<T>` if it is named properly.

wadoon:
I have removed the `var` types and updated the branch to the current KeY version.

mattulbrich:
This does not fully reach its potential: The JavaCC parser still runs over the expression which is set; this is unlike in the case for JML assertions where all expressions are supported. Hence

    //@ set b = (\forall int k; true);

still raises an exception for a boolean ghost variable b.

mattulbrich:
I reverted commit c539948718b55c86456478f7b599f68cbd1ede20 since that has nothing to do with the problem here and introduces a GUI runtime exception due to pretty printing. The issue can be separated and be resolved properly elsewhere.

mattulbrich:
(this is WIP. forced updates to the branch are not unlikely in the near future.)

PR #3196, opened 2023-07-08
Repairs on KeYParser

### LINKED ISSUE #3130: KeY parsing pretty-printed terms can change operator precedence of `/` and `%`
## Description

I have a term of the form `a / b % c` in my sequent (with pretty syntax on). Here, the division takes precedence over modulo (and is thus equivalent to `(a / b) % c`). The issue: If I use this term in the taclet instantiation window (e.g. cut rule) then the term gets parsed differently as `a / (b % c)`. This unexpectedly changes the operator precedence.

It's easy to accidentally cause this if part of your workflow is to click-drag or copy-paste terms from the sequent into the cut rule window, for example.

## Reproducible

Always.

### Steps to reproduce 

1. Load a problem like below:
```
\problem {
    (10 / 5) % 2 = 0
}
```
(Notice the parentheses around the division. If you omitted them the term would display as `10 / (5 % 2) = 0` in the sequent.)

2. Make sure pretty syntax is on. You should see ` ⟹ 10 / 5 % 2 = 0`.
3. Check that the division is taking precedence over the modulo (hover your mouse over the subterms).
4. Apply the cut rule and enter the same fomula `10 / 5 % 2 = 0`.

**Expected behavior:**  
The new formula on the sequent is also `10 / 5 % 2 = 0` (or, alternatively, `(10 / 5) % 2 = 0`).

**Actual behavior:**  
The new formula on the sequent is `10 / (5 % 2) = 0`.

### Additional information

Here's how variations of the above term are parsed in KeY:

`10 / 5 % 2` → `10 / (5 % 2)`  
`(10 / 5) % 2` → `10 / 5 % 2`  
`10 / (5 % 2)` → `10 / (5 % 2)`  

`10 % 5 / 2` → `10 % (5 / 2)`  
`(10 % 5) / 2` → `10 % 5 / 2`  
`10 % (5 / 2)` → `10 % (5 / 2)`  

Related: #1560 #1192 #726 

---

* Commit: 5ecc1aa794

wadoon:
* fix #3130
* enable UTF operators
* test for precedences


key.core/src/main/java/de/uka/ilkd/key/nparser/builder/ExpressionBuilder.java
```diff
@@ -375,16 +394,47 @@ public Object visitStrong_arith_term_1(KeYParser.Strong_arith_term_1Context ctx)
...
+        if (ldt == null) {
+            // falling back to integer ldt (for instance for untyped schema variables)
+            ldt = services.getTypeConverter().getIntegerLDT();
+        }
+
+        assert ctx.op.size() == ctx.b.size();
+
+        for (int i = 0; i < termL.size(); i++) {
+            var opName = ctx.op.get(i).getType() == KeYLexer.PERCENT ? "mod" : "div";
+            Function op = ldt.getFunctionFor(opName, services);
```
mattulbrich, line 430: Is division hence also possible for floats and doubles thanks to this?

mattulbrich (approved):

key.core/src/main/java/de/uka/ilkd/key/nparser/builder/ExpressionBuilder.java
wadoon: It does a lookup after the corresponding LDT, and request the function of it. 

It only depends, whether the LDT answers for the given sort.

PR #3198, opened 2023-07-09
Reduce clutter in toolbar

FliegendeWurst:
image
View menu:
image

This reduces clutter in the main toolbar by:
- moving the action history next to the other undo buttons
- moving the layout controls to a menu (View -> Layout)
- marking the exploration extension as experimental (as documented)
- disabling the functionality to reorder toolbars (only ever used by accident)

As a bonus, the bug that extension panels sometimes fail to show up should also be fixed.

### DIFF
--- a/key.ui/src/main/java/de/uka/ilkd/key/gui/docking/DockingLayout.java
+++ b/key.ui/src/main/java/de/uka/ilkd/key/gui/docking/DockingLayout.java
@@ -8,6 +8,7 @@ import java.io.File;
 import java.io.IOException;
 import java.util.*;
 import java.util.List;
+import javax.annotation.Nonnull;
 import javax.swing.*;
 
 import de.uka.ilkd.key.core.KeYMediator;
@@ -40,10 +41,10 @@ public final class DockingLayout implements KeYGuiExtension, KeYGuiExtension.Sta
         KeYGuiExtension.MainMenu {
     private static final Logger LOGGER = LoggerFactory.getLogger(DockingLayout.class);
 
-    public static final float SIZE_ICON_DOCK = 12f;
-    public static final File LAYOUT_FILE = new File(PathConfig.getKeyConfigDir(), "layout.xml");
-    public static final String[] LAYOUT_NAMES = new String[] { "Default", "Slot 1", "Slot 2" };
-    public static final int[] LAYOUT_KEYS =
+    private static final float SIZE_ICON_DOCK = 12f;
+    private static final File LAYOUT_FILE = new File(PathConfig.getKeyConfigDir(), "layout.xml");
+    private static final String[] LAYOUT_NAMES = new String[] { "Default", "Slot 1", "Slot 2" };
+    private static final int[] LAYOUT_KEYS =
         new int[] { KeyEvent.VK_F10, KeyEvent.VK_F11, KeyEvent.VK_F12 };
 
     private MainWindow window;
@@ -126,13 +127,19 @@ public final class DockingLayout implements KeYGuiExtension, KeYGuiExtension.Sta
         DockingHelper.restoreMissingPanels(window);
     }
 
+    @Nonnull
     @Override
-    public List<Action> getMainMenuActions(MainWindow mainWindow) {
+    public List<Action> getMainMenuActions(@Nonnull MainWindow mainWindow) {
         List<Action> actions = new ArrayList<>();
         int keypos = 0;
         for (String layout : LAYOUT_NAMES) {
             Integer key = keypos < LAYOUT_KEYS.length ? LAYOUT_KEYS[keypos] : null;
             actions.add(new LoadLayoutAction(mainWindow, layout, key));
+            keypos++;
+        }
+        keypos = 0;
+        for (String layout : LAYOUT_NAMES) {
+            Integer key = keypos < LAYOUT_KEYS.length ? LAYOUT_KEYS[keypos] : null;
             actions.add(new SaveLayoutAction(mainWindow, layout, key));
             keypos++;
         }
@@ -149,7 +156,7 @@ final class SaveLayoutAction extends MainWindowAction {
     SaveLayoutAction(MainWindow mainWindow, String name, Integer key) {
         super(mainWindow);
         this.layoutName = name;
-        setName("Save as " + name);
+        setName("Save " + name);
         setIcon(IconFactory.saveFile(MainWindow.TOOLBAR_ICON_SIZE));
         setMenuPath("View.Layout");
         if (key != null) {
@@ -162,7 +169,7 @@ final class SaveLayoutAction extends MainWindowAction {
     @Override
     public void actionPerformed(ActionEvent e) {
         mainWindow.getDockControl().save(layoutName);
-        mainWindow.setStatusLine("Save layout as " + layoutName);
+        mainWindow.setStatusLine("Layout saved to " + layoutName);
     }
 }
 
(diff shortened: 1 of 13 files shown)


wadoon:
> This reduces clutter in the main toolbar. Other options to consider:
> * moving the layout controls to a menu

Yes the layout toolbar can be moved to a menu.

FliegendeWurst:
@WolframPfeifer I believe the bug that extension panels sometimes fail to show up is fixed with the latest commit.

key.ui/src/main/java/de/uka/ilkd/key/gui/extension/impl/KeYGuiExtensionFacade.java
```diff
@@ -221,6 +223,7 @@ public static List<KeYGuiExtension.Toolbar> getToolbarExtensions() {
      */
     public static List<JToolBar> createToolbars(MainWindow mainWindow) {
         return getToolbarExtensions().stream().map(it -> it.getToolbar(mainWindow))
+                .peek(x -> x.setFloatable(false))
```
WolframPfeifer, line 226: I am ok with setting all toolbars to non-floatable, as I have never used the possibility to rearrange them. I think it is indeed a good idea to remove this. However, since this makes the layout more restrictive than earlier, we should ask others for feedback.
@KeYProject/reviewers What do you think? Has anyone ever used this feature productively?

WolframPfeifer:
> @WolframPfeifer I believe the bug that extension panels sometimes fail to show up is fixed with the latest commit.

I tested it by switching to an older branch without proof slicing and then back to this one, and the slicing panel appeared as expected. Thanks!

key.ui/src/main/java/de/uka/ilkd/key/gui/docking/DockingLayout.java
```diff
@@ -135,65 +136,112 @@ private void setLayout(String layout) {
...
-    public List<Action> getMainMenuActions(MainWindow mainWindow) {
         ensureActions(mainWindow);
+
         return actions;
     }
 }
 
 
-class SaveLayoutAction extends MainWindowAction {
+final class SaveLayoutAction extends MainWindowAction {
```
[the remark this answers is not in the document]
FliegendeWurst: Previously they were added to the menu. The only thing special about them is that they load/save a specific layout. I can remove them and instead add shortcuts for the regular load/save actions.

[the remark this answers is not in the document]
WolframPfeifer: For reference, here are screenshots:
| current main | this PR |
|------|------------------|
|image|image|

Personally, I think that the old interface (left) is clearer. For example, in the right one, it is unclear if the layout is already changed when changing the selection of the radio button or only after selecting and clicking "Load Layout". This was already a problem of the toolbar version. In addition, the old version can be used with keyboard shortcuts.

Either way, if you remove the toolbar buttons, the pair of classes can be removed (or merged with the other pair).

FliegendeWurst:
Current state:

image

key.ui/src/main/java/de/uka/ilkd/key/gui/extension/impl/KeYGuiExtensionFacade.java
wadoon: This a feature, that is not very popular amongst Swing applications. I have no obstacles, and we can revert this easily.

key.ui/src/main/java/de/uka/ilkd/key/gui/docking/DockingLayout.java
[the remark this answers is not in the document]
wadoon: The intention of the left one was to have the shortcut available. Some sort of Quick-Save/Quick-Load known from games. I am not sure, whether this was popular. 

I think this could be useful, if more and more views are added, and we also gain different perspective, e.g., one for classic logic, and one for code-centric.

Currently, I would tend to the left one, but without the additional menu level of Load/Save.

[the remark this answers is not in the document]
FliegendeWurst: I have updated the code to flatten the menu (see my other comment for a screenshot).

key.ui/src/main/java/de/uka/ilkd/key/gui/extension/impl/KeYGuiExtensionFacade.java
```diff
@@ -82,6 +81,14 @@ public static void addExtensionsToMainMenu(MainWindow mainWindow, JMenuBar menuB
...
     }
 
+    /*
+     * public static Optional<Action> getMainMenuExtensions(String name) {
+     * Spliterator<KeYGuiExtension.MainMenu> iter =
+     * ServiceLoader.load(KeYGuiExtension.MainMenu.class).spliterator(); return
+     * StreamSupport.stream(iter, false) .flatMap(it -> it.getMainMenuActions(mainWindow).stream())
+     * .filter(Objects::nonNull) .filter(it -> it.getValue(Action.NAME).equals(name)) .findAny(); }
+     */
+
```
WolframPfeifer:
```suggestion
```
Let's remove this code instead of having it commented out (this doesn't really help anybody).

key.ui/src/main/java/de/uka/ilkd/key/gui/docking/DockingLayout.java
```diff
@@ -140,102 +126,17 @@ private void setLayout(String layout) {
...
-                    }
-                }
-                mainWindow.getDockControl().save(layout);
-            }
+        int keypos = 0;
+        for (String layout : LAYOUT_NAMES) {
+            Integer key = keypos < LAYOUT_KEYS.length ? LAYOUT_KEYS[keypos] : null;
+            actions.add(new LoadLayoutAction(mainWindow, layout, key));
+            actions.add(new SaveLayoutAction(mainWindow, layout, key));
+            keypos++;
```
WolframPfeifer, line 137:
```suggestion
        for (String layout : LAYOUT_NAMES) {
            Integer key = keypos < LAYOUT_KEYS.length ? LAYOUT_KEYS[keypos] : null;
            actions.add(new LoadLayoutAction(mainWindow, layout, key));
            keypos++;
        }
        for (String layout : LAYOUT_NAMES) {
            Integer key = keypos < LAYOUT_KEYS.length ? LAYOUT_KEYS[keypos] : null;
            actions.add(new SaveLayoutAction(mainWindow, layout, key));
            keypos++;
```

Grouping the menu entries by slot makes it hard to read in my opinion:
image

My suggestion groups them by action:
image

key.ui/src/main/java/de/uka/ilkd/key/gui/docking/DockingLayout.java
```diff
@@ -245,7 +146,7 @@ final class SaveLayoutAction extends MainWindowAction {
     private static final long serialVersionUID = -2646217961498111734L;
     private final String layoutName;
 
-    public SaveLayoutAction(MainWindow mainWindow, String name, Integer key) {
+    SaveLayoutAction(MainWindow mainWindow, String name, Integer key) {
         super(mainWindow);
         this.layoutName = name;
         setName("Save as " + name);
```
WolframPfeifer, line 152:
```suggestion
        setName("Save " + name);
```
"Save as" does not really make sense. It should be "Save to", but to have a cleaner menu, I would just remove the preposition.

WolframPfeifer (changes requested):
I have a few final requests (see my suggestions). I would like to have the status line messages harmonized to `Layout saved to ...` and `Layout loaded from ...`.

Otherwise, I think the PR is fine and really increases usability of KeY.

FliegendeWurst:
I have integrated all of your suggestions.

WolframPfeifer (approved):
Thanks!

PR #3201, opened 2023-07-14
Proof Management Command Line Tool (+ small GUI extension)

WolframPfeifer:
This PR finally brings the proof management tool to the main branch (development started at the HacKeYthon in 2018!). The description of the features can be found in the Readme (also added below for convenience).

# Impact on key.core/key.ui
* For the dependency check, it is necessary to know under which modality a contract was applied in a proof. Therefore, this information is saved in the proofs from now on. Note that for legacy proofs without saved modality, the dependency check assumes that the contracts have been applied under diamond modality (sound, but may lead to false positives).
* For performance reasons, there is a flag "deleteIntermediateTree" which can be set to false to allow reuse of the intermediate representation built during proof replay (this avoids double work in the proof management). For calls from KeY, the flag is always set to true, which keeps the old behavior.
* The CommandLine class has been extended to support subcommands (pm check, pm merge, ...) with their own sets of options and arguments.
* The check command can be called via a GUI extension to make it more visible to the users.

# Proof Management (copied from the Readme)

This project contains a proof management tool for KeY proofs. It can be used as a stand-alone program via its console
interface or as a GUI extension for KeY.

## Motivation
When verifying Java source code with JML specifications, it is standard that multiple methods with multiple
contracts must be proven. However, due to memory and time constraints, the corresponding proofs are often not conducted
during a single run of KeY. Furthermore, it is frequently the case that the source has to be re-loaded multiple times in
KeY between two proofs, since changes had to be made, for example, to the specification of the methods.

With this workflow it is crucial to have some form of proof management, since it is very difficult for the user to
manually track all changes and detect the proofs that have to be re-conducted after a change in the source code.
KeY comes with a built-in proof management that partially solves the problem. However, this internal proof management
has a few shortcomings:
1) It only considers the proofs currently loaded in the GUI. This is often not really helpful, since usually it is not a
   good idea to have too many proofs loaded at the same time (memory restrictions).
2) It only considers proofs loaded in the same environment (via "File" -> "Proof Management"), which is an option that
   is not obvious to many users. Furthermore, if the source code changes, a new environment is (and has to be, for 
   soundness reasons) created.
3) The current proof management does not check for compatible settings (overflows, different integer semantics, runtime
   exceptions, ...).
4) Some dependencies are not checked by KeY, for example cycles involving model methods or dependency contracts.

The intention of this project is to provide a tool that can check a proof bundle (a directory or zip file that contains
source code and a collection of proofs) for consistency.

## Features
The proof management tool contains the following features:
* **MissingProofsChecker:** Detects contracts (in the user-provided source code) that have no proof. Note that only
  unproven contracts from inside the `src` folder are reported (not from the cp or bcp folders, see below).
* **SettingsChecker:** Detects incompatible taclet settings (called `Choices` internally in KeY) between the proofs of a
  bundle. An example of this would be that one proof is conducted using mathematical integer semantics and another one
  with Java integers.
* **ReplayChecker:** Detects proofs that can not be replayed, for example because they were conducted with an old KeY
  version or because the proof file was edited manually and contains syntax errors.
* **DependencyChecker:** Detects proofs that contain unproven dependencies (i.e., rely on contracts that are unproven).
  In addition, the dependency checker reports illegal cyclic dependencies between proofs, for example if two mutually
  recursive methods contain wrong `measured_by` clauses and are proven separately with KeY (and thus the bundle contains two proofs that are both correctly closed per se).
* **HTML report:** Besides the plain text messages on the console, an HTML report can be generated that contains the
  console messages in a filterable format, lists contracts, proofs, and shows their corresponding status.

### Stand-alone Program (CLI)
```
Usage: pm <command> [<options>...] [<args>...]
  available commands:
    check: Checks a single proof bundle for consistency.
    merge: Merges multiple proof bundles.
```

#### check
The `check` command performs the selected consistency checks and is able to give a console or HTML report:
```
pm check [--missing] [--settings] [--replay] [--dependency] [--report <out_path>] <bundle_path>
```
The available options correspond to the features described in the section above.
`<bundle_path>` is the path of the proof bundle to check and can either denote a directory or a zip file.

The directory structure of the bundle has to conform that described in
[classpath documentation](https://key-project.org/docs/user/Classpath/).
An example of a (zipped) proof bundle would be:
```
bundle.zproof
  - src            // java classes (.java files only)
    - A.java
    - mypackage
      - B.java
      - C.java
  - classpath      // optional: classpath (may contain .jar files and subdirectories
                   //           with .java and/or .class files)
    - someLibrary.java
    - otherLibrary
      - Lib.java
      - Util.class
  - bootclasspath  // optional: bootclasspath (system classes from the Java class library),
                   //           replaces the files shipped with KeY
    - java
      - lang
        - Math.java
        - String.java
        - System.java
        ...
      - math
      ...
```

#### merge
The `merge` command merges multiple bundles in zip or directory format into a single one.
```
pm merge [--force] [--check "<check_args>"] <bundle1> <bundle2> ... <output>
```
Merging fails when two or more of the input bundles contain files with the same name (at the same path), but different
content. This can be the case if for example different versions of the Java code were used for the proofs. To merge
nonetheless, the `--force` flag can be set. In this case, the resulting bundle will contain the versions from the first
given bundle.

As a shortcut, the command has a `--check` option which performs the check command on the resulting bundle after the
merge and forwards the given arguments to that command.

### KeY GUI Extension
The project adds the `Proof Management` menu to KeY. It currently contains only a single entry which opens a dialog to
configure the checkers to run and the path for the HTML report to generate. After a successful run, the HTML report is
opened in the system's default browser.

## Known Problems / Missing features
Some features are currently not implemented:
* It is not checked that user-defined rules (taclets included in user-provided .key files) are proven.
* For the SettingsChecker, taclet settings have to be exactly identical at the moment, which could probably be relaxed a
  bit in the future.
* An `--auto` flag could be added for the `check` command, which tries to proof contracts without explicit proofs
  automatically. There could also be an option to save these automatically found proofs into the bundle.
* There could be a `bundle` command that creates a proof bundle by creating the required file structure based on the
  information provided in a `.key` file.


keyext.proofmanagement/src/main/java/org/key_project/proofmanagement/check/KeYFassade.java
```diff
@@ -0,0 +1,472 @@
...
+import org.key_project.proofmanagement.io.Logger;
+import org.key_project.proofmanagement.io.ProofBundleHandler;
+import org.key_project.util.reflection.ClassLoaderUtil;
+
+/**
+ * This class provides static methods to access the prover (KeY).
+ *
+ * @author Wolfram Pfeifer
+ */
+public final class KeYFassade {
```
flo2702, line 59: Fassade → Facade

keyext.proofmanagement/src/main/java/org/key_project/proofmanagement/check/Node.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.SortedSet;
+import java.util.TreeSet;
+
+/**
+ * Represents a vertex/node of a tree.
+ *
+ * @param <T> the type of the content contained by the node
+ * @author Wolfram Pfeifer
+ */
+public class Node<T extends Comparable<T>> implements Comparable<Node<T>> {
```
flo2702, line 12: Can this go into `key.util`?

keyext.proofmanagement/src/main/java/org/key_project/proofmanagement/check/dependency/ContractAppCollector.java
```diff
@@ -0,0 +1,216 @@
...
+        ImmutableSet<Contract> contracts = specRepo.splitContract(c);
+
+        // load information about the modality under which the contract was applied
+        Modality modality = Modality.getModality(biApp.getModality());
+        DependencyGraph.EdgeType edgeType;
+        if (modality == null) {
+            // in default case (e.g. legacy proofs without saved modality information)
+            // we assume diamond modality but print a warning
+            logger.print(LogLevel.WARNING, "No saved modality information was found!" +
+                " Assuming \"diamond\" (incomplete for box contracts)!");
```
flo2702, line 199: Shouldn't we assume box/termination-insensitive to be sound?

key.ui/src/main/java/de/uka/ilkd/key/util/CommandLine.java
```diff
@@ -189,6 +185,42 @@ protected void print(PrintStream ps, int descriptionCol) {
     public CommandLine() {
     }
 
+    private String usedSubCommand = "";
+
+    private final Map<String, CommandLine> subcommands = new HashMap<>();
+
+    public CommandLine addSubCommand(String name) {
```
flo2702, line 192: Can you add some documentation for this, esp. what the return value is?

key.ui/src/main/java/de/uka/ilkd/key/util/CommandLine.java
```diff
@@ -189,6 +185,42 @@ protected void print(PrintStream ps, int descriptionCol) {
...
+        }
+        if (subcommands.containsKey(name)) {
+            throw new IllegalArgumentException(name + " has already been registered");
+        }
+        CommandLine subCommand = new CommandLine();
+        subcommands.put(name, subCommand);
+        return subCommand;
+    }
+
+    public CommandLine getSubCommandLine(String name) {
```
flo2702, line 205: `getSubCommandLine` → `getSubCommand`

keyext.proofmanagement/src/main/java/org/key_project/proofmanagement/check/Node.java
WolframPfeifer: The class is not really general purpose: For a tree, in general the nodes are not needed to extend comparable. I tried to remove it, but for correct sorting in interplay with string template, it is needed (maybe it could be done differently, but I did not find an easy way). In addition, the compare method ignores the parent and child nodes at the moment. So I would suggest keeping the class here.

keyext.proofmanagement/src/main/java/org/key_project/proofmanagement/check/dependency/ContractAppCollector.java
WolframPfeifer: No. Inside a box modality, it is always sound to apply any (matching) contract. This means that such a contract application is ignored for the dependency check (not included into any strongly connected component). However, since potentially the contract could have been applied from a diamond modality (we do not know), we need to be conservative and assume that the modality was a diamond. This way the contract application (edge) can be included in the strongly connected components and needs to have a termination measure to be legal. This might mark some contract applications as illegal and generate false positives (cycles marked as illegal even though they are not), which then could be fixed easily by adding the modality information to the proof manually (with a text editor).

I think that most of the time the contracts and modalities in use should be termination sensitive, so it should work out of the box in the majority of cases.

keyext.proofmanagement/src/main/java/org/key_project/proofmanagement/check/KeYFassade.java
WolframPfeifer: Fixed!

key.ui/src/main/java/de/uka/ilkd/key/util/CommandLine.java
WolframPfeifer: Done.

key.ui/src/main/java/de/uka/ilkd/key/util/CommandLine.java
WolframPfeifer: I changed the name as suggested and added documentation.

flo2702 (approved):

PR #3099, opened 2023-04-03
A grammar for configuration

wadoon:

PR #3202, opened 2023-07-14
Abbreviations Manager

wadoon:
Abbreviations are built-in to the KeY system, but this construct is not very popular. This may be to their volatility: you are not able to store, and reuse them on new proof attempts.
This PR brings a new widget: 

image

The widget offers following features:

* view current defined abbreviations
* load and store abbreviation mappings
  * A text file where each line `<label>::==<term>` describes an abbreviation.
* transfer (import) abbreviations of currently open proofs
* Context menu:
  * change the abbreviation label
  * delete an abbreviation
  * disable an abbreviate


jwiesler:
Thanks for abbreviation delete!

What would be great imo is if I try to create an abbreviation and that name is already taken: Show an option to remove the old and use this as the new label term.

jwiesler:
Can't this be stored in the proof file itself? I think this would be a better place to store.

wadoon:
> Can't this be stored in the proof file itself? I think this would be a better place to store.

My current story for load/store is that you attempt to proof something, notice a specification (or KeY) failure, fix it and restart the attempt. Then you can rescue the abbreviations. 

Proof attempts are rather not stored as proof files. But maybe we should consider storing the abbreviations in the settings. I would wait for #3031 for a type-safe lists in the settings.

jwiesler:
#3031 is already merged?

Settings would be a good idea as well.

wadoon:
> #3031 is already merged?
> 
> Settings would be a good idea as well.

Sorry, mean #3099

FliegendeWurst:
I am confused by the save abbreviations dialog. It suggests to save this data as a `.proof`:
```
p::==i_arr_5
d::==i_arr_2
l::==i_arr_8
```

FliegendeWurst:
How does the load abbreviations functionality work? I selected my previously saved file, but no abbreviations were restored.

FliegendeWurst:
The abbreviations transfer does not work for program variables. I think they would need to be compared by name, not by identity.

Issue #3200, opened 2023-07-10
Automode sometimes not entered when running strategy


FliegendeWurst:
## Description

When quickly starting the automated proof search after loading a proof, the "automode" is not activated. This means that every proof step updates the UI, slowing everything down.

## Reproducible

Somewhat hard to reproduce.

### Steps to reproduce 

1. Load one of the examples (I used SumAndMax)
2. Quickly press on the start button in the toolbar

If you do it right, you will get this sequence of automode stops/starts:
<details>

```
stopping interface
current stack
java.base/java.lang.Thread.getStackTrace(Thread.java:1610)
de.uka.ilkd.key.core.KeYMediator.stopInterface(KeYMediator.java:592)
de.uka.ilkd.key.proof.io.ProblemLoader.runAsynchronously(ProblemLoader.java:137)
de.uka.ilkd.key.gui.WindowUserInterfaceControl.loadProblem(WindowUserInterfaceControl.java:102)
de.uka.ilkd.key.gui.WindowUserInterfaceControl.loadProblem(WindowUserInterfaceControl.java:107)
de.uka.ilkd.key.gui.MainWindow.loadProblem(MainWindow.java:1358)
de.uka.ilkd.key.gui.actions.OpenExampleAction.actionPerformed(OpenExampleAction.java:33)
de.uka.ilkd.key.gui.MainWindow.openExamples(MainWindow.java:1354)
de.uka.ilkd.key.gui.WindowUserInterfaceControl.openExamples(WindowUserInterfaceControl.java:284)
de.uka.ilkd.key.core.Main.loadCommandLineFiles(Main.java:224)
de.uka.ilkd.key.core.Main.main(Main.java:185)

stopping interface
current stack
java.base/java.lang.Thread.getStackTrace(Thread.java:1610)
de.uka.ilkd.key.core.KeYMediator.stopInterface(KeYMediator.java:592)
de.uka.ilkd.key.gui.WindowUserInterfaceControl.loadingStarted(WindowUserInterfaceControl.java:466)
de.uka.ilkd.key.proof.io.AbstractProblemLoader.load(AbstractProblemLoader.java:273)
de.uka.ilkd.key.proof.io.ProblemLoader.doWork(ProblemLoader.java:64)
de.uka.ilkd.key.proof.io.ProblemLoader$1.doInBackground(ProblemLoader.java:117)
de.uka.ilkd.key.proof.io.ProblemLoader$1.doInBackground(ProblemLoader.java:110)

[...]

PR #3203, opened 2023-07-15
Fix automode freezing the UI

FliegendeWurst:
See #3200 for details. This PR fixes all of #3200.

### DIFF
--- a/key.ui/src/main/java/de/uka/ilkd/key/core/Watchdog.java
+++ b/key.ui/src/main/java/de/uka/ilkd/key/core/Watchdog.java
@@ -38,7 +38,7 @@ public final class Watchdog {
     private static void run() {
         while (true) {
             try {
-                Thread.sleep(10000);
+                Thread.sleep(20000);
             } catch (InterruptedException e) {
                 return;
             }
@@ -66,8 +66,7 @@ public final class Watchdog {
              * Common-Cleaner TIMED_WAITING
              */
 
-            for (int i = 0; i < threads.length; i++) {
-                var thread = threads[i];
+            for (Thread thread : threads) {
                 if (thread == null || IGNORED_THREADS.contains(thread.getName())) {
                     continue;
                 }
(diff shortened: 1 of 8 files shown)


key.ui/src/main/java/de/uka/ilkd/key/ui/MediatorProofControl.java
```diff
@@ -95,6 +95,9 @@ public void stopAutoMode() {
      */
     @Override
     public void waitWhileAutoMode() {
+        if (SwingUtilities.isEventDispatchThread()) {
+            return; // do not block the UI thread
```
jwiesler, line 99: This is really unexpected. Maybe this should throw an exception instead of simply returning? Blocking the event thread is a programming error.

key.ui/src/main/java/de/uka/ilkd/key/core/Watchdog.java
```diff
@@ -0,0 +1,116 @@
...
+import java.util.Set;
+
+import de.uka.ilkd.key.util.ThreadUtilities;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Watchdog to monitor the state of the KeY system. If all worker and UI threads
+ * are waiting / blocked, a deadlock is reported.
```
jwiesler, line 13: What is the purpose of this? Debugging? Is there any resolution when an error is detected?

FliegendeWurst: The purpose of this is to catch further bugs. Currently, the only resolution is some logging. It is not easy to notify the user since the UI thread will also be blocked.

key.ui/src/main/java/de/uka/ilkd/key/ui/MediatorProofControl.java
FliegendeWurst: I think the only usage on the event thread is in `AutoModeAction`. I will change it there.

key.ui/src/main/java/de/uka/ilkd/key/core/Watchdog.java
jwiesler: Ok, that's what I thought. Then it would be great if you could add this to the doc comment :)

mattulbrich (approved):

Issue #834, opened 2022-12-23
Local in-method variables are not correctly handled


wadoon:
This issue was created at [git.key-project.org](https://git.key-project.org/key/key/-/issues/834) where the discussions are preserved.
    
----
    
* Mantis: [MT-1284](http://i12www.ira.uka.de/~klebanov/mantis/view.php?id=1284)
* Submitted on: 2013-03-27 by (at)most
* Updated: 2016-01-14

### Description
I am not sure how widespread this problems is, but it certainly applies to loops. The scenario is:
 
- a (final) local variable declared before the loop
- loop invariant that refers to this local variable
- an early proof split that makes KeY create different instances of this local
     variable on the different branches (localVar, localVar_0), etc. 
 
On the branches where the _X renaming takes place the local variable is not correctly applied to the invariant. 

### Steps to reproduce

Load the attached file, make a meaningful cut (paramToCut = 0), try to prove, one of the "Invariant Initially Valid" branches is not closeable.

## Files
  
`Loop.java`:

```java
public class Loop {
  

  int i;

  //@ requires param == false;
  //@ ensures true;
  void method(boolean param, int paramToCut) {
    final boolean test = param;
   
    i = 0;
    /*(at)
        loop_invariant this.i>=0 && this.i<=10;
        loop_invariant !test;
        decreasing 10 - this.i;
        assignable this.i;
      (at)*/
    while(i<10) {
      i++;
    }
  }

}

```

## Notes

_(at)most at 2013-04-08_
> Anyone?

_(at)grahl at 2014-05-12_
> I do not understand why local variables would need to be named differently between branches...

_(at)grahl at 2015-01-20_
> Removing the 2.4 target marker as nobody is assigned.

## Associated Bugs

* Relates to MT-1375 with 

## History

* (at)most -- (`NEW_BUG`) _2013-03-27_

* (at)most -- (`FILE_ADDED`) _2013-03-27_

* (at)most -- (`BUGNOTE_ADDED`) _2013-04-08_

* (at)grahl -- (`NORMAL_TYPE`) _2013-06-09_

* (at)grahl -- (`NORMAL_TYPE`) _2013-06-09_

* (at)grahl -- (`NORMAL_TYPE`) _2013-06-27_

* (at)grahl -- (`BUG_ADD_RELATIONSHIP`) _2013-10-24_

* (at)grahl -- (`NORMAL_TYPE`) _2013-11-25_

[...]

PR #3205, opened 2023-07-17
Fix for bug with effectless updates

WolframPfeifer:
This draft PR fixes a bug where updates are dropped even though they would have an effect. This happens since the DropEffectlessElementariesCondition performs the check using a LinkedHashSet, which relies on equals of LocationVariable. However, this is not implemented and defaults to equals of Object.

At the moment, I do not know why this bug only appears after pruning in the proof. Directly after opening/loading the instances of LocationVariable are the same, thus the check succeeds and the update is not dropped ...

Since the change might have a big impact, let's see if the tests run successfully.

### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/logic/op/LocationVariable.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/logic/op/LocationVariable.java
@@ -86,4 +86,29 @@ public final class LocationVariable extends ProgramVariable implements Updateabl
             argSorts(), name().toString(), arity(),
             whereToBind(), isRigid());
     }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (!(obj instanceof LocationVariable)) {
+            return false;
+        }
+        LocationVariable that = (LocationVariable) obj;
+        return Objects.equals(getKeYJavaType(), that.getKeYJavaType())
+            && isStatic() == that.isStatic()
+            && isModel() == that.isModel()
+            && isGhost() == that.isGhost()
+            && isFinal() == that.isFinal()
+            && sort().equals(that.sort())
+            && Objects.equals(argSorts(), that.argSorts())
+            && name().toString().equals(that.name().toString())
+            && arity() == that.arity()
+            && Objects.equals(whereToBind(), that.whereToBind())
+            && isRigid() == that.isRigid();
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(getKeYJavaType(), isStatic(), isModel(), isGhost(), isFinal(), sort(),
+            argSorts(), name().toString(), arity(), whereToBind(), isRigid());
+    }
 }


unp1:
> This draft PR fixes a bug where updates are dropped even though they would have an effect. This happens since the DropEffectlessElementariesCondition performs the check using a LinkedHashSet, which relies on equals of LocationVariable. However, this is not implemented and defaults to equals of Object.

That is strange. As LocationVariable is an Operator and operators are only considered equal if they are the same object, the defaulting to java.lang.Object's equals-method should be safe.

unp1:
Thought a bit about it: The current fix would consider that two Locationvariable's are equal even if they are not the same Object. This hints that when pruning somehow the pruned back sequent got changed which could indicate some serious problem: 
either 

- somewhere the immutability of terms, programs or sequents got violated or 
- in case this happens only for a variable that is renamed before then the effectless sth. rule is applied (if that is the case, I might have fixed the problem coincidentally with PR #3224 (commit: [6b6f356b8](https://github.com/KeYProject/key/pull/3224/commits/6b6f356b869400b92b5aee3a0c6d5fff9d622c7b)

FliegendeWurst:
> Do you have an example (and short description) where I can reproduce the problem?

There is a simple example in #834.

mattulbrich:
@unp1, I think that #834 is indeed an example for the bug.

To me it seems that the invariant of the loop there refers to a local variable "test". However, there seem to be two different variables named test on two branches. I assume that the SpecificationRepository can only refer to one of them, thus the check might overlook it. (That's only a guess).

Still, the equal method for location variables does not fit the design principles chosen in KeY.

key.core/src/main/java/de/uka/ilkd/key/logic/op/LocationVariable.java
```diff
@@ -86,4 +86,29 @@ public int hashCodeModProofIrrelevancy() {
             argSorts(), name().toString(), arity(),
             whereToBind(), isRigid());
     }
+
+    @Override
+    public boolean equals(Object obj) {
```
mattulbrich, line 91: I think that KeY has been such that references to a program variable only refer to the same variable if they use the same location variable operator object.

If that is really to be changed, I really think the entire code base must be inspected to change this fundamental decision.

mattulbrich (changes requested):

mattulbrich:
Richard @unp1 wrote:
> Thought a bit about it:  [...] Either
>     * somewhere the immutability of terms, programs or sequents got violated or

Exactly! This is what I assume is behind #3206 related to this problem. There in JmlAssert, there is a field which violates immutability. In #834, I assume that thanks to the specification repository, immutability is not necessarily guaranteed.(Or is it?)

FliegendeWurst:
> In https://github.com/KeYProject/key/issues/834, I assume that thanks to the specification repository, immutability is not necessarily guaranteed.(Or is it?)

The issue in #834 is that the LocationVariable is created twice by applying `methodBodyExpand` in different branches. Perhaps a better solution would be to store the location variables after expanding a method body and then try to reuse these objects when expanding the same method body in another branch.

mattulbrich:
> Perhaps a better solution would be to store the location variables after expanding a method body and then try to reuse these objects when expanding the same method body in another branch.

It would probable fix the issue in #834. There might also be several instances of the same modality on one sequent. (e.g. for info flow proof obligations). Would it be safe to reuse the existing program variable for both declarations. (It might be sound for valid Java programs. Is it?)

unp1:
Hi, 
 I also co-assigned myself to this bug and will look into it. The specification repository should not be a problem w.r.t. immutability. When variables are renamed the renaming should be (and I think is) done on the loop invariant itself and then be registered in addition. Such that the old mapping origLoop -> origLoppInv and loopWithRenamedVariables -> loopInvWithRenamedVariables are registered. 
 
 When pruning, the original loop is back and the original loop invariant should be found. This would break with overwriting the equals method.

mattulbrich:
> The specification repository should not be a problem w.r.t. immutability. When variables are renamed the renaming should be (and I think is) done on the loop invariant itself and then be registered in addition. 

Thanks for the clarification, Richard @unp1. I wonder: If there are two branches with two different loop invariants for the same (object-identical) loop (b/c of symbolic execution), how does the repository know which loop invariant goes with which branch, or is the repository branch-local?

unp1:
> Thanks for the clarification, Richard @unp1. I wonder: If there are two branches with two different loop invariants for the same (object-identical) loop (b/c of symbolic execution), how does the repository know which loop invariant goes with which branch, or is the repository branch-local?

That is a nice corner case and I have to look up if we had thought about it. At the moment I think it does not occur often as loops often have the same loop invariant and loops and loop invariants use local variables and method parameters. For these different variables are introduced on each branch and thus the loops are no longer equal on both branches.

mattulbrich:
Ah! So: It might be a not-so-bad solution to clone the method body prior
to method-body-statement execution? Then there are two different loop
objects (albeit holding the same content) that can have different loop
invariants.

I'd also like to clone JMLAssert statements on similar occasions (but
they have their condition locally)

unp1:
Yes. Also at the moment I am wondering what happens in case we do a cut right before applying the loop invariant rule. Maybe we need to add some unique branch name as part of the KeY (or have the spec repository branch local).

PR #3207, opened 2023-07-18
Remove remains of Webstart (+ other maintenance)

FliegendeWurst:
Java Webstart was removed in Java 11, which is required by KeY.


wadoon (approved):

WolframPfeifer (approved):
Thanks for the maintenance!

PR #3213, opened 2023-07-21
fixate bash as shell for dlsmt.sh

wadoon:
The release tests had an error on non-linux environments s.t. `dlsmt.sh` breaks.

1. We fixed the shell to `bash` which repairs windows
2. We removed the shebang which fixes macos

Covert change: removal of `jacocoTestReport.xml` from the artifacts for unit-tests, which should decreases the huge size of `test-results`.

### DIFF
--- a/.github/dlsmt.sh
+++ b/.github/dlsmt.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/sh
+#!/usr/bin/bash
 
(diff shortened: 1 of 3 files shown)


wadoon:
[We have positive test cases](https://github.com/KeYProject/key/actions/runs/5630989857) on windows and macos!

Issue #3214, opened 2023-07-22
Glitches to be fixed for 2.12


mattulbrich:
## Description

Issues identified during UI testing for the release 2.12.

Please add your points below:
* [x] Increasing the font size affects the proof tree only after a restart of KeY. But the distances are decreased. (s. A)
* [x] The Proof Statistics Window looks strange with the chaged font sizes. (s B)
* [x] What does "Copy referenced proof" mean. There should be a tooltip.
* [x] When removing the .key directory, recent files are no longer stored (s. C)

### Images:
A: 
![image](https://github.com/KeYProject/key/assets/17046629/6f63593e-99bb-4629-8530-8e095c2ccfed)

B:
![image](https://github.com/KeYProject/key/assets/17046629/651810d7-f3e3-4c9a-bc1b-b663ff49bf51)
![image](https://github.com/KeYProject/key/assets/17046629/92176f07-04b1-45a1-a875-a9f29daaea8f)

C:
```
Could not write recent files listjava.io.FileNotFoundException: /home/mattias/.key/recentFiles.props (No such file or directory)
	at java.base/java.io.FileInputStream.open0(Native Method)
	at java.base/java.io.FileInputStream.open(FileInputStream.java:219)
	at java.base/java.io.FileInputStream.<init>(FileInputStream.java:157)
	at de.uka.ilkd.key.gui.RecentFileMenu.store(RecentFileMenu.java:247)
	at de.uka.ilkd.key.gui.RecentFileMenu.save(RecentFileMenu.java:260)
```

FliegendeWurst:
I fixed issue 4, but I can't reproduce the broken layout of the proof tree. Which font setting did you change?

WolframPfeifer:
The statistics dialog (B) is better now as it has less superfluous space, but the button font does not match the rest of the text:
![image](https://github.com/KeYProject/key/assets/94171076/5064adb0-3d07-4edf-b340-de863d2d5dcd)

I think this is a general problem: Why do we have different font settings (for sequent/proof tree, and the rest)? Can we streamline this and have only a single global font factor setting?

And, by the way, I think that decreasing the font factor does not really work at the moment, as it affects only some places, but not in others, for example not in the tabs (proof tree, goals, etc.) and also not for the settings dialog itself. Increasing, on the other hand, has an immediate effect on all of them. This can lead to very strange situations, e.g. in this example after increasing to 1.5 and decreasing to 0.5 afterwards:
[...]

Issue #1652, opened 2022-12-23
Collected Runtime Exceptions


wadoon:
This issue was created at [git.key-project.org](https://git.key-project.org/key/key/-/issues/1652) where the discussions are preserved.
    
----
    
There tend to be a lot of random runtime exceptions which should not ever occur.

Instead of having to report each of them individually (plus preparing a minimal example and a description), I'd at least like to collect stacktraces here such that we can gather an overview of the "glitches" that are still there.

Conditions for listing here:
* Stacktrace was shown on console
* Exception does not harm operation in a conceivable way.
* It looks like a missing local check or something else which could be fixed locally.

If you fix one of these conditions here, please add a comment, so that others do not re-investigate.

Stacktraces need not be complete, the calls through the swing framework might as well be dropped if no relevant information is lost.
    
----
    
Information: 
* created_at: 2021-12-22T10:59:05.915Z
* updated_at: 2022-08-01T19:25:55.777Z
* closed_at: None  (closed_by: )
* milestone: 
* user_notes_count: 34

mattulbrich:
Below I copy the reported exceptions from gitlab

```
Exception in thread "AWT-EventQueue-0" java.lang.IllegalArgumentException: bad position: 4104
	at java.desktop/javax.swing.text.JTextComponent.setCaretPosition(JTextComponent.java:1689)
	at de.uka.ilkd.key.gui.nodeviews.SequentViewSearchBar.setExtraHighlight(SequentViewSearchBar.java:240)
	at de.uka.ilkd.key.gui.nodeviews.SequentViewSearchBar.searchNext(SequentViewSearchBar.java:149)
	at de.uka.ilkd.key.gui.SearchBar$4.actionPerformed(SearchBar.java:87)
	at java.desktop/javax.swing.JComponent$ActionStandin.actionPerformed(JComponent.java:3443)
	at java.desktop/javax.swing.SwingUtilities.notifyAction(SwingUtilities.java:1810)
```

mattulbrich:
```
Exception in thread "AWT-EventQueue-0" java.lang.NullPointerException
	at de.uka.ilkd.key.java.JavaInfo.getAllAttributes(JavaInfo.java:1144)
	at de.uka.ilkd.key.java.JavaInfo.getCanonicalFieldProgramVariable(JavaInfo.java:1108)
	at de.uka.ilkd.key.pp.FieldPrinter.isCanonicField(FieldPrinter.java:80)
[...]

Issue #1648, opened 2022-12-23
Settings Dialog: Appearance -> Font Factor for Sequent and Proof Tree Requires Restart to Take Effect


wadoon:
This issue was created at [git.key-project.org](https://git.key-project.org/key/key/-/issues/1648) where the discussions are preserved.
    
----
    
 

## Description

Using the settings dialog to change the font size of the sequent and proof tree, requires a restart of KeY to take effect.
This is independent of using "Apply" or "OK" button.

Changing the global factor takes effect immediately after clicking "Apply" or "OK"

Further the global factor is a factor, so named correctly.

The entry "Tree and sequent font factor" is actually a font size and should be renamed to "Tree and sequent font size" (btw. should we allow other font sizes as well?)

## Reproducible
  
always (macos 12.1)

### Steps to reproduce 

> Describe the steps needed to reproduce the issue.

0. Load a problem
1. Open the Options > Settings
2. Select "Appearance & Behaviour"
3. Change "Tree and sequent font factor" 
  -  Press "Apply" -> Nothing happens
  -  Press "OK" -> Nothing happens (Except as expected the dialog disappearing, but no font size changes in the sequent or proof tree)
4. Restart -> Font size has changed of sequent and tree

> What is your expected behavior and what was the actual behavior?

### Additional information

nothing more to add

---

* Commit: 57a19c0ce6fb08e9dcf183eb4116afbc72d85b05 

    
----
    
Information: 
* created_at: 2021-12-18T15:02:59.093Z
* updated_at: 2021-12-20T15:09:03.282Z
* closed_at: None  (closed_by: )
* milestone: v2.12.0
* user_notes_count: 2

FliegendeWurst:
Fixed in #3216

Issue #1526, opened 2022-12-23
Abandoning while proving raises exception


wadoon:
This issue was created at [git.key-project.org](https://git.key-project.org/key/key/-/issues/1526) where the discussions are preserved.
    
----
    
 

## Description

When pressing `ctrl-w` while in automatic proof mode (that is sometimes possible), concurrency-related exception
may be raised

## Reproducible
  
random

### Steps to reproduce 

Run automatic proof and abondon it while running. 

I managed to get into this by accidentally pressing STOP twice (thus restarting) and then abandoning.
The system had some lack at the time, it may be that the order of events was hence 'unusual'.

---

* Commit: 9a28833079d3cbcadeedb948cd0d4507c12ba6f7

* Component: /label ~GUI

    
----
    
Information: 
* created_at: 2019-11-08T10:50:43.572Z
* updated_at: 2019-11-08T10:50:43.572Z
* closed_at: None  (closed_by: )
* milestone: 
* user_notes_count: 0

PR #3216, opened 2023-07-22
Fix various UI bugs

### LINKED ISSUE #3215: Loading a PO from the project manager makes KeY freeze
## Description

Loading a PO from the project manager makes KeY freeze.

## Reproducible
  
always on my machine

### Steps to reproduce 

1. Open the example selector window
2. Select "Remove Duplicates"
3. Select the first method PO
4. Run auto to close
5. Open the proof manager
6. Select the second method PO
7. KeY freezes

### Additional information

git bisect reveals that c07a00fba07e8513966c2e958d797a5d6023d424 is the culprit.

I wonder if this removal has to do with the problem:
```diff
     @Override
     public void progressStopped(Object sender) {
-        mainWindow.getMediator().startInterface(true);
+        // no need to call startInterface(), will be done by ProblemLoader once loading is done
     }
```

---

* Commit: 0ac6d7f2f89621db5bcd223c76470fdcc857ac4d

### LINKED ISSUE #3228: FileNotFoundException because of missing proof settings
## Description

KeY throws a FileNotFoundException if the proof-settings file is missing

## Reproducible
  
always

### Steps to reproduce 

1. Delete or rename your `~/.key` directory
2. Start KeY
3. Observe the exception "No proof-settings could be loaded, using defaultsjava.io.FileNotFoundException: ~/.key/proof-settings.props (No such file or directory)" This exception is thrown whenever you start KeY until you create a proof-settings file by loading a proof obligation.

---

* Commit: 0ccf763d02de8e5b63d293d42ee214d079637bde

FliegendeWurst:
I recommend reviewing this PR commit-by-commit. Most of the bugs were found using the new UI testing facility (https://github.com/KeYProject/key/pull/3216/commits/3d73e59bd6ba8a8a15fb36f93aa193d3d7bd65d3).

Fixes #3228, #3227, #3215, #1648, https://github.com/KeYProject/key/issues/1652#issuecomment-1613075565, #1526, #1117 and task 1+4 of #3214


mattulbrich:
#3214 could be of interest here too.

key.ui/src/main/java/org/key_project/util/java/SwingUtil.java
```diff
@@ -149,4 +150,46 @@ public static JScrollPane createScrollPane(JTable table) {
...
+     *
+     * @param tree the tree to search
+     * @param tp the tree path
+     * @return a tree path or null
+     */
+    public static TreePath findPathInNewTree(JTree tree, TreePath tp) {
+        TreePath root = tree.getPathForRow(0);
+        Object last = root.getLastPathComponent();
+        int pathIndex = 0;
+        if (!last.toString().equals(tp.getPathComponent(pathIndex).toString())) {
```
jwiesler, line 166: Why does this have to be the toString repr? Filtering shouldn't change the nodes so why is reference equality not enough?

jwiesler (approved):

key.ui/src/main/java/org/key_project/util/java/SwingUtil.java
FliegendeWurst: The nodes are cleared here: https://github.com/KeYProject/key/blob/0ac6d7f2f89621db5bcd223c76470fdcc857ac4d/key.ui/src/main/java/de/uka/ilkd/key/gui/prooftree/GUIProofTreeModel.java#L377-L380

FliegendeWurst: I think your approach would be more efficient. I'll use it instead if it works as expected.

PR #3222, opened 2023-07-25
Print real condition in if labels

FliegendeWurst:
This PR modifies the labels in the proof tree to show the actual condition of ifs instead of the boolean variable introduced by the unfold taclet. With this change, the branch labels in the proof tree are much more useful.
|previous|now|
Sum_pre | Sum_post
Contains_pre | Contains_post


key.core/src/main/java/de/uka/ilkd/key/logic/op/LocationVariable.java
unp1, line 1: As at the moment the use is limited to the GUI. A suggestion would be to extend the proof (or services)  oand the TacletApp object with a mapping from ProgramVariable to TacletApp. Invoking TacletApp#execute would then add transfer its local map to the services/proof map.

key.core/src/main/java/de/uka/ilkd/key/logic/op/LocationVariable.java
```diff
@@ -30,6 +30,10 @@ public LocationVariable(ProgramElementName name, KeYJavaType t) {
         super(name, t.getSort(), t, null, false, false, false);
     }
 
+    public LocationVariable(ProgramElementName name, KeYJavaType t, Object origin) {
```
wadoon, line 33: I am not a friend of using an type `Object` here. Is there any reason why this is needed?

key.core/src/main/java/de/uka/ilkd/key/logic/op/ProgramVariable.java
```diff
@@ -262,4 +274,8 @@ public MatchConditions match(SourceData source, MatchConditions matchCond) {
      */
     abstract public Operator rename(Name name);
 
+    public Object getOrigin() {
```
wadoon, line 277: I plea for a different name, as this collides `HasOrigin#getOrigin`, which uses the parser to give a (debug) information where a symbol was defined.

key.core/src/main/java/de/uka/ilkd/key/logic/op/LocationVariable.java
FliegendeWurst: I used `Object` to keep the interface flexible. I can change it to TacletApp.

key.core/src/main/java/de/uka/ilkd/key/logic/op/ProgramVariable.java
FliegendeWurst: I have changed it to `createdByRuleApp`.

key.core/src/main/java/de/uka/ilkd/key/logic/op/LocationVariable.java
FliegendeWurst: Hmm, I don't really see the reason to not store this information on the LocationVariable itself. The RuleApp stored in the variable is not related to the UI.

FliegendeWurst: New Plan: WeakHashMap in Proof of LocationVariable -> RuleApp

unp1: Additional remark: Use Proof#userData framework

FliegendeWurst: This is now implemented

key.core/src/main/java/org/key_project/proof/LocationVariableTracker.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.WeakHashMap;
+
+import de.uka.ilkd.key.logic.OpCollectorJavaBlock;
+import de.uka.ilkd.key.logic.op.LocationVariable;
+import de.uka.ilkd.key.proof.Proof;
+import de.uka.ilkd.key.proof.ProofEvent;
+import de.uka.ilkd.key.proof.RuleAppListener;
+import de.uka.ilkd.key.rule.RuleApp;
+
+public class LocationVariableTracker implements RuleAppListener {
```
WolframPfeifer, line 16: Can you please create a bit of JavaDoc here? It should be clear also without the context of this PR what the class is/can be used for.

key.ui/src/main/java/de/uka/ilkd/key/core/KeYMediator.java
```diff
@@ -103,6 +105,10 @@ public class KeYMediator {
      * Currently opened proofs.
      */
     private final DefaultListModel<Proof> currentlyOpenedProofs = new DefaultListModel<>();
+    /**
+     * Currently loaded proofs.
+     */
+    private final WeakHashMap<Proof, Object> loadedProofs = new WeakHashMap<>();
```
WolframPfeifer, line 111: As discussed in our meeting: Since this is just used as a set (the value is always set to null), it may be possible to just use the list model instead. If not, please state in the comment that it is used as a set here.

WolframPfeifer (approved):
Thanks, nice feature! I have two minor complaints (mainly about the documentation), apart from that, the PR is fine.

For documentation: Since branch names are stored in proofs, unfortunately this PR does not take effect when loading proofs.

In the future, we could think of overwriting saved branch labels with the better ones from this PR. However, this would also overwrite manually named branches (I am not sure if anyone uses the feature of manual branch names ...).

key.ui/src/main/java/de/uka/ilkd/key/core/KeYMediator.java
FliegendeWurst: I had another look. It seems the `currentlyOpenedProofs` aren't actually managed by the mediator, they are just stored there. It is difficult to figure out exactly when proofs are added to this list. For now I think it is easier to handle duplicated proof loaded events (by removing this map and associated checks).

Issue #3158, opened 2023-06-05
Loop scope rule uses variables outside the name space


mattulbrich:
## Description

The loop scope invariant rule introduces variables (usually called `h` and `a` that are not within the namespace of the current goal).

In some cases this may even lead to two variables of the same name, but with different types on the sequent.

## Reproducible

always  

### Steps to reproduce 

1. Load the sum and max example from the example browser..
2. In the proof search strategy set "Loop treatment" to "none".
3. Set "Arithmetic treatment" to "basic".
4. Push the green arrow button
5. On the one open goal apply the rule "loopScopeInvDia".
6. Navigate to the 2nd open goal

Two things are bad now:
* There is an update `h := heap`, but you cannot perform a cut `h = h` because `h` is an unknown name.
* There is an update `a := _a.length - k` assigning an int value to `a`. However, `a` is the parameter of the method and an `int[]` variable.

The rule fails to register the assigned variables in the goal context (in other rules that usually happens by declaring the new names within modalities).

### Additional information

Discovered, since this is a deal breaker for research on a scripting engine.

---

* Commit: 470b0d67e179aefc5995d4bb1c0bcd0da7654cb8

mattulbrich:
Another related problem with the loop scope rule is that it introduces conflicting location variables with the same name.

1. Load Binary Search and press autoplay.
2. Have a look at node 125 (invariant preserved and used)

It contains two variables called `a`. One of type any (introduced to store the variant) and one of type int[] which is the array to be searched.

Here a screenshot indicating the problem:
![image](https://github.com/KeYProject/key/assets/17046629/741a785d-be1a-44ef-ae49-1d5ff82e4e96)

on 0ac6d7f2f89621db5bcd223c76470fdcc857ac4d

PR #3224, opened 2023-07-25
Fix for issue #3158

unp1:
Fix for issue #3158
- The program variable replacer also needs to consider lefthandside's of elementary updates
- Avoid collisions with names occurring in the namespace (and not only inside programs)
- Introduce new program variables properly in loop invariant rules
- Minor clean ups


unp1:
This PR will break saved proofs as the loop scope taclets needed to be modified

key.core/src/main/resources/de/uka/ilkd/key/proof/rules/loopScopeRules.key
```diff
@@ -62,6 +62,13 @@
 
         "Invariant Preserved and Used":
             \replacewith(
+              \modality{#dia}{
+                    #typeof(#heapBefore_LOOP) #heapBefore_LOOP;
+                    #typeof(#savedHeapBefore_LOOP) #savedHeapBefore_LOOP;
+                    #typeof(#permissionsBefore_LOOP) #permissionsBefore_LOOP;
+                    #typeof(#permissionsBefore_LOOP) #permissionsBefore_LOOP;
```
FliegendeWurst, line 69: Why is this heap duplicated?

key.core/src/main/resources/de/uka/ilkd/key/proof/rules/loopScopeRules.key
```diff
@@ -125,6 +133,13 @@
 
         "Invariant Preserved and Used":
             \replacewith(
+              \modality{#box}{
+                                #typeof(#heapBefore_LOOP) #heapBefore_LOOP;
+                                #typeof(#savedHeapBefore_LOOP) #savedHeapBefore_LOOP;
+                                #typeof(#permissionsBefore_LOOP) #permissionsBefore_LOOP;
+                                #typeof(#permissionsBefore_LOOP) #permissionsBefore_LOOP;
```
FliegendeWurst, line 140: Same here

key.core/src/main/resources/de/uka/ilkd/key/proof/rules/loopScopeRules.key
unp1: Thanks! Is resolved in new commit.

unp1:
The reduced test coverage is an unrelated file occurs because the exception is no longer thrown and hence the corresponding catch block in DiskFileRepo is no longer reached. That is expected behavior and should be ignored.

unp1:
Will resume checks after next push

mattulbrich:
The comment I made in #3158 seems to be fixed. "a" is now "a_1":
image

mattulbrich (approved):
This fixes the problem reported in the issue.

We haven't had these modalities introducing new program variables in a while, ... but if it works ... Would a direct addition to the namespace also be possible?

It's a bit strange that h can be referred to from the poststate of <>, but not toplevel. It is also actually not allowed in KeY. I tried a cut with
```
\<{ int x = 42; }\>x=42
```
which is a syntax error.

unp1:
> This fixes the problem reported in the issue.
> 
> We haven't had these modalities introducing new program variables in a while, ... but if it works ... Would a direct addition to the namespace also be possible?

Sadly that is not as easy and it would require non-trivial changes to the taclet application process. The time when the variables are created in the taclet app is not the same time when the taclet app is applied. Hence, we need to book keep which variables might need to be replaced when applying such a taclet etc. 

But I agree it would be nice not to have to split the introduction of new variables in two steps.

> 
> It's a bit strange that h can be referred to from the poststate of <>, but not toplevel. It is also actually not allowed in KeY. I tried a cut with
> 
> ```
> \<{ int x = 42; }\>x=42
> ```
> 
> which is a syntax error.

I remember some discussions (years ago) that we wanted to allow that again (it was allowed pre KeY 2.0), as a simple way to introduce new program variables.

WolframPfeifer:
I would favor this PR for the 2.12.0 release over #3243, since I think it is a more complete solution to the naming problems. Are there other disadvantages besides the broken reloading of older proofs? Because we broke that anyway due to the changed integer defaults and the changed OSS ruleset ...

What do you think?

@unp1
@mattulbrich

mattulbrich:
The feature of introducing new variables like `x` in `<int x = 42>x >0` is not supported outside this use case. Is it clear that we want to have that feature at all? I would argue that the scope of x is local to the modality. A new variable should be added otherwise. ...

I seem to remember that that was a reason to split this PR

unp1:
> The feature of introducing new variables like `x` in `<int x = 42>x >0` is not supported outside this use case. Is it clear that we want to have that feature at all? I would argue that the scope of x is local to the modality. A new variable should be added otherwise. ...
> 
> I seem to remember that that was a reason to split this PR

Hi,
 it might make sense to use this fix for the moment as w/o this fix two things are problematic:

* parsing: as different variables of different types cannot be disambiguated
* sequent comprehension: different variables with same names used

with this patch there is still the parsing problem due to the scoping, but someone just trying to understand the sequent will not be confused.

Best regards,
Richard

WolframPfeifer:
> > The feature of introducing new variables like `x` in `<int x = 42>x >0` is not supported outside this use case. Is it clear that we want to have that feature at all? I would argue that the scope of x is local to the modality. A new variable should be added otherwise. ...
> > I seem to remember that that was a reason to split this PR
> 
> Hi, it might make sense to use this fix for the moment as w/o this fix two things are problematic:
> 
>     * parsing: as different variables of different types cannot be disambiguated
> 
>     * sequent comprehension: different variables with same names used
> 
> 
> with this patch there is still the parsing problem due to the scoping, but someone just trying to understand the sequent will not be confused.
> 
> Best regards, Richard

I am a bit confused now: By "this fix", you mean this PR (#3224) or the cherry-picked one (#3243)?

Since that seems to lead to a longer discussion, we could also consider keeping that status quo. I do not think it is that critical, since we have had that bug for years now ...

unp1:
Hi, 

> I am a bit confused now: By "this fix", you mean this PR (#3224) or the cherry-picked one (#3243)?

By this fix I meant this PR.

I am fine with either decision.

Best regards, 
Richard

PR #3226, opened 2023-07-26
Fix default look-and-feel in settings dialog

flo2702:
When first started, KeY uses the Java cross-platform look-and-feel called Metal. But opening the settings dialog automatically selects the System look-and-feel as default.

Thus, when installing KeY on a platform that provides a system look-and-feel, you will see the Metal LnF at first, but as soon as you generate a settings file by opening the settings dialog, KeY will switch to the System LnF even if you don't select anything.

This PR fixes this by always applying Metal as the default.


FliegendeWurst:
> Thus, when installing KeY on a platform that provides a system look-and-feel, you will see the Metal LnF at first, but as soon as you generate a settings file by opening the settings dialog, KeY will switch to the System LnF even if you don't select anything.

Interesting, on my system it was the other way around (first System, then Metal). My fix 614852f78a514dac4ccc34151a0293442271564e in #3216 ensures that Metal is explicitly set when starting KeY.

PR #3231, opened 2023-07-26
method names should not be created via VariableRenamer

unp1:


unp1:
Converted to draft. As I am not sure it won't have side-effects when parsing interactively. I have an idea for a safer change.

wadoon (approved):

PR #3236, opened 2023-08-02
Fix most remaining font scaling issues

FliegendeWurst:
https://github.com/KeYProject/key/issues/3214#issuecomment-1660607454

This PR fixes several issues:
- decreasing or increasing the font size is effective immediately, in all places (*except the settings tree view)
- the special setting of the proof tree / sequent font size is also scaled with the global factor

cc @WolframPfeifer


WolframPfeifer (approved):
Works nicely! This really a usability improvement, thanks!

PR #3237, opened 2023-08-03
Linearized proof tree

FliegendeWurst:
This PR changes the proof tree to be slightly more linear. In effect, the "Normal Execution" branch will (visually) continue on the parent branch.  See the screenshots below for what the new view filter would look like. If we agree to add this option, the implementation still needs to be changed before merging.

linearTree3
|previous|now|
linearTree1 | linearTree2

### DIFF
--- a/key.ui/src/main/java/de/uka/ilkd/key/gui/prooftree/GUIAbstractTreeNode.java
+++ b/key.ui/src/main/java/de/uka/ilkd/key/gui/prooftree/GUIAbstractTreeNode.java
@@ -2,8 +2,10 @@ package de.uka.ilkd.key.gui.prooftree;
 
 
 import java.lang.ref.WeakReference;
+import java.util.ArrayList;
 import java.util.Enumeration;
 import java.util.LinkedList;
+import java.util.List;
 import javax.annotation.Nonnull;
 import javax.swing.tree.TreeNode;
 
@@ -18,6 +20,8 @@ public abstract class GUIAbstractTreeNode implements TreeNode {
     // and ProofTreeView.delegateView.lastPathComponent
     private final WeakReference<Node> noderef;
 
+    private TreeNode parent;
+
     protected GUIProofTreeModel getProofTreeModel() {
         return tree;
     }
@@ -31,7 +35,9 @@ public abstract class GUIAbstractTreeNode implements TreeNode {
 
     public abstract int getChildCount();
 
-    public abstract TreeNode getParent();
+    public TreeNode getParent() {
+        return parent;
+    }
 
     public abstract boolean isLeaf();
 
@@ -67,10 +73,10 @@ public abstract class GUIAbstractTreeNode implements TreeNode {
         return path.toArray(new TreeNode[0]);
     }
 
-    protected TreeNode findBranch(Node p_node) {
+    protected GUIAbstractTreeNode findBranch(Node p_node) {
         TreeNode res = getProofTreeModel().findBranch(p_node);
         if (res != null) {
-            return res;
+            return (GUIAbstractTreeNode) res;
         }
 
         String label = ensureBranchLabelIsSet(p_node);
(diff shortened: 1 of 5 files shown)


mattulbrich:
How is determined which is the "main" branch and which are the "side" branches. For a cut, e.g., both should be side branches. I like this option (we have done it like this in ivil and a bit in dive), and would favour to see it on the main branch.

FliegendeWurst:
> How is determined which is the "main" branch and which are the "side" branches.

Currently only the "Normal Execution [...]" branch is considered the main branch.

FliegendeWurst:
The implementation has been updated to also consider e.g. the TRUE branch of `cut_direct` as the main branch:
image

This is achieved using a new goal template tagging mechanism.

key.core/src/main/resources/de/uka/ilkd/key/proof/rules/javaRules.key
```diff
@@ -4,6 +4,9 @@
 
 // This file contains rules for handling java programs
 
+// The "Normal Execution" prefix of branch names is treated
+// specially by the "Linearize Proof Tree" functionality
+// (see GUIBranchNode)
```
Drodt, line 9: If you have now added the ability to tag goal templates, why still treat the branches with "Normal Execution" as the name as special? I would prefer to just have the tag option and add it to all relevant rules.

key.core/src/main/antlr4/KeYParser.g4
```diff
@@ -752,7 +752,7 @@ option_expr
 
 goalspec
 :
-  (name=string_value COLON)?
+  (name=string_value (LBRACKET tag=simple_ident RBRACKET)? COLON)?
```
Drodt, line 755: This is fine for this PR but we should think about/discuss if there is any general value in goal template tags. What other use cases can we find?

In a future PR, I would like this not to be any identifier. We should declare these tags similar to rule sets.

Drodt (changes requested):

key.core/src/main/resources/de/uka/ilkd/key/proof/rules/javaRules.key
FliegendeWurst: Done (tagging). The reason I treated "Normal Execution" as special is that it will always have the main tag, so it felt a bit redundant.

Drodt (dismissed):

Drodt:
It seems broken now.

image

FliegendeWurst:
Works for me

image

Drodt:
Can you try the Binary Search example?

FliegendeWurst:
Seems to work as well.

image

Drodt:
Sorry, may mistake. The setting was switched off again :facepalm:

FliegendeWurst:
Yeah, all of those settings aren't saved I think.

FliegendeWurst:
Seems odd to have a `class TacletGoalTemplate extends org.key_project.prover.rules.tacletbuilder.TacletGoalTemplate`. I'll fix the conflict later

FliegendeWurst:
Resolved conflicts, still works after adding the tag to the new goal template in key.ncore.

FliegendeWurst:
CI failure unrelated

Drodt (commented):
My only concern is this: If the linear tree and "Hide Intermediate Proof Steps" are both enabled, the goals tagged as "main" disappear:

image

Either we don't allow this combination or we deal with it differently.

FliegendeWurst:
> There is no "Closed Goal" node for the main branch next to the "Null Reference" branch

Yes, that is expected. That goal has been split further below, right? Each split eventually leads to a closed goal node.

Drodt (approved):

wadoon:
Updated branch. 

@Drodt Added for discussion in the KaKeY today.

Drodt (approved):

PR #3249, opened 2023-08-20
Stop automode or exit on SIGINT

FliegendeWurst:
When KeY is interrupted using a SIGINT, the current action will be finished: either automode is stopped or KeY is terminated.
This is very useful if the GUI is not responding (e.g. in out of memory situations).


wadoon:
I do not think it is wise to access to `sun.*` classes. Also, if the KeY is programmed in a sound manner, it should not be needed, as the UI keeps responsible. 

Please provide more descriptive information.

FliegendeWurst:
> I do not think it is wise to access to `sun.*` classes. Also, if the KeY is programmed in a sound manner, it should not be needed, as the UI keeps responsible.
> 
> Please provide more descriptive information.

If KeY runs out of memory, the GUI is no longer responsive. Stopping automode may free up some memory. Without this kind of functionality it is impossible to recover from an OOM situation.

mattulbrich:
Well, KeY has some nasty concurrency bugs, and sometimes it would be nice to trigger it to leave automode from the outside. OOM is only one possibility. It is rather OS-specific and the API may disappear w/o notice. However, I would say this is a helpful technical feature we need not advertise too much. Perhaps add a note that this is helpful but can be removed without breaking anything.

FliegendeWurst:
I have amended the code to handle possible errors. Testing on Windows reveals that the functionality also works there (Ctrl+C on the command line). I think Mac will be similar to Linux. Both use almost the same signal infrastructure.

WolframPfeifer (approved):
Under Ubuntu it works fine for me. Under Windows, it seems that due to the behavior of the terminal (both cmd and powershell), it is only possible to send a single SIGINT via Ctrl+C (afterwards you have a prompt again, despite KeY still running). However, I don't think it is necessary to do something here, as this is mainly a debug feature ...

PR #3241, opened 2023-08-04
Post-Release 2.12.0: Java and Version

wadoon:
* Switch the Java version to 17 (in preparation of Java 21)
* Set KeY's version to 2.14.0

To be merged after #3242

PR #3252, opened 2023-08-20
Applying Java 17 refactorings

wadoon:
Applies refactorings after the switch to Java 17 for after #3241.

Including: 

* Transform immutable data classes to Java records
* Rewrite `instanceof` and casts to Pattern expressions
* Use `.toList()` instead of `.collect(Collectors.toList)`
* Use Text Block Literals (`""" ... """`) on large strings instead of concatenation cascades.


WolframPfeifer:
The missing space when using instanceof (e.g. `if (t.op()instanceof ProgramVariable var) {`) with method calls/arrays was indeed a bug in the Eclipse formatter: https://bugs.eclipse.org/bugs/show_bug.cgi?id=574437
I fixed it by upgrading the spotless version and thus also the formatter.

However, I had to fix the formatter version to 4.22, since more recent versions have another bug, such that it crashes while trying to format our `SymbolicExecutionTreeBuilder` class.

WolframPfeifer (approved):

PR #3263, opened 2023-08-26
Handle bound variables in SMT unknown translation

### LINKED ISSUE #3262: SMT translation of unknown values incorrect
## Description

It is is possible to prove the following using Z3:
```
 \forall Seq s; s = seqDef{int u;}(0, s.length, s[u]),
 seqSub(s1, 0, maxx - 1).length = maxx - 1
==>
 s1.length = maxx - 1
```
Equivalent `.key`:
```
\functions {
  Seq s1;
  int maxx;
}

\problem {
  \forall Seq s; (s = (seqDef{int u;}(0, s.length, any::seqGet(s, u))))
  & seqLen(seqSub(s1, 0, maxx - 1)) = maxx - 1
 ->
  s1.length = maxx - 1
}
```

### Additional information

This is the relevant part of the SMT translation:
```
; --- Sequent
(assert
  (!
    (forall ((var_s U)) (=> (instanceof var_s sort_Seq) (= var_s unknown_0)))
    :named
    L_1))
(assert
  (!
    (=
      (k_seqLen (k_seqSub u_s1 (i2u 0) (i2u (- (u2i u_maxx) 1))))
      (i2u (- (u2i u_maxx) 1)))
    :named
    L_2))
(assert
  (! (not (= (k_seqLen u_s1) (i2u (- (u2i u_maxx) 1)))) :named L_3))
(check-sat)
; --- Translation of unknown values
; unknown_0 :  seqDef{u:int}(Z(0(#)),seqLen(s),any::seqGet(s,u))
```
In particular I think this allows the SMT solver to prove `seqSub(s1, 0, maxx - 1) = unknown_0 = s1`, which is obviously wrong.

cc @WolframPfeifer @mattulbrich

FliegendeWurst:
Previously, terms that cannot be translated to SMT would always be modeled as a constant `unknown_X`.
This is not correct when the term depends on a quantified variable.
The fix: model unknown terms as a function `unknown_X var_1 var_2` where `var_*` are the free variables in the unknown term.

Fixes #3262.


mattulbrich (changes requested):
As discussed: The bound variables can be retrieved locally and need not be carried around so much.

FliegendeWurst:
Ready for review.

mattulbrich (approved):

WolframPfeifer:
For the future: Maybe we should update the comments generated for unknown symbols:
```
; --- Translation of unknown values
; unknown_0 :  seqDef{u:int}(Z(0(#)),seqLen(s),any::seqGet(s,u))
; unknown_1 :  seqDef{u:int}(Z(0(#)),seqLen(s1),any::seqGet(s1,u))
```
Since unknown_0 is a function, I would propose something like this:
```
; --- Translation of unknown values
; unknown_0(s:Seq) :  seqDef{u:int}(Z(0(#)),seqLen(s),any::seqGet(s,u))
; unknown_1 :  seqDef{u:int}(Z(0(#)),seqLen(s1),any::seqGet(s1,u))
```

PR #3264, opened 2023-08-27
Further Refactorings for Java 17+

wadoon:
Refactoring and Code clean up.

* switch statements into switch expression where it was useful. 
  (avoids fall throughs and often more compact code)

* some smaller (non-structured) clean ups
  * ensure camel cases variable names
  * fix Javadoc
  * and other code smells recognize by IntelliJ


key.core.rifl/src/main/java/de/uka/ilkd/key/util/rifl/RIFLHandler.java
Drodt, line 1: Taking this as the example for all switch expressions: Why is the content of the switch not indented? I.e.:
```java
switch (localName) {
    case "sourcedompair", "source" -> startSource();
    ...
}
```

Drodt (changes requested):

key.core.rifl/src/main/java/de/uka/ilkd/key/util/rifl/RIFLHandler.java
wadoon: I just follow the formatting rules given by spotless. This might be question for @WolframPfeifer.

unp1: Checked the changes and they look fine from my side.

Drodt (approved):

PR #3251, opened 2023-08-20
Fixing small issues to fix release: KeY 2.12.1

mattulbrich:
Some glitches have slipped through for the release 2.12.0.

(branched from KeY-2.12.0)

### .key files cannot be edited if they have an error

The "edit file" button does not work if there is a syntax error in a key file to be loaded. Instead the following stacktrace occurred:
```
Exception in thread "AWT-EventQueue-0" java.lang.NullPointerException: Cannot invoke "String.equals(Object)" because the return value of "java.net.URI.getScheme()" is null
	at de.uka.ilkd.key.gui.actions.EditSourceFileAction.tryGetFile(EditSourceFileAction.java:209)
	at de.uka.ilkd.key.gui.actions.EditSourceFileAction.actionPerformed(EditSourceFileAction.java:283)
	at java.desktop/javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:1972)
	at java.desktop/javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2313)
	at java.desktop/javax.swing.DefaultButtonModel.fireActionPerformed(DefaultButtonModel.java:405)
	at java.desktop/javax.swing.DefaultButtonModel.setPressed(DefaultButtonModel.java:262)
	at java.desktop/javax.swing.plaf.basic.BasicButtonListener.mouseReleased(BasicButtonListener.java:279)
	at java.desktop/java.awt.Component.processMouseEvent(Component.java:6626)
```
Fixed with b21475eac5bdd584536c738b911787a3d91506e7

PR #3268, opened 2023-09-01
Automatically abbreviate anon heap names

FliegendeWurst:
By automatically abbreviating heap names, it is much easier to navigate and inspect the sequent view. The screenshots below show the same proof node. **Discussion points**: Should this option be automatically enabled? Is the naming scheme (`heap_loop_X`) fine?

| Previously (300k characters) | Now (35k characters, 8.5x shorter!) |
Screenshot_20230901_163156 | Screenshot_20230901_163057

Heap names after method calls don't need to be abbreviated, they are already just referred to by "heapAfter_commonEntry" or similar.

This PR builds on top of #3251 and #3258. Only the last five commits in this branch are relevant.

### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/rule/AbstractLoopInvariantRule.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/rule/AbstractLoopInvariantRule.java
@@ -38,6 +38,10 @@ import org.key_project.util.collection.ImmutableSet;
  * @author Dominic Scheurer
  */
 public abstract class AbstractLoopInvariantRule implements BuiltInRule {
+    /**
+     * The suffix given to anon heap names created because of loops.
+     */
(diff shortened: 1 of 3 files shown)


mattulbrich:
A few remarks from our discussion:
* Insider knowledge: Arrays are best created in a separate method; this clutters the heap considerably less.
* Probably there could and should be a "contract-like rule" for array creation as well, which does not go through all stages of object creation.
* For method invocations, there is a new heap symbol, same could be incorporated into the scoped loop rule, the result would be something like `loopHeap_17 = anon(.....) ==> {heap := loopHeap || ...} [...]...`
* `create(heap,o)` could already set all fields to their default value, so no need for this stage of the initialisation (/cc @weigl for the new infrastructure?) We could get rid of `memset`, too.

WolframPfeifer:
We decided that the solution with abbreviations is not what we want to have, thus I close this PR now.

PR #3282, opened 2023-09-22
Update to Java 21 Runtime for Testing

wadoon:
This PR enables the newest LTS version of Java in Githubs CI. 

:warning: This does not increase the `--release` or language level for KeY. New features are not enabled. 

:exclamation: Java 21 still not available for Github Actions.

### DIFF
--- a/.github/workflows/code_quality.yml
+++ b/.github/workflows/code_quality.yml
@@ -29,7 +29,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-java@v3
         with:
-          distribution: 'temurin'
+          distribution: 'corretto'
           java-version: '21'
           cache: 'gradle'
       - name: Build with Gradle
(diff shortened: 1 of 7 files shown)


WolframPfeifer (approved):
Is there any impact from switching from Adoptium Temurin JDK builds to Amazon Corretto? If not, maybe we should stick with the more open variant ...

Apart from that, I am OK with the PR.

wadoon:
> Is there any impact from switching from Adoptium Temurin JDK builds to Amazon Corretto? If not, maybe we should stick with the more open variant ...
> 
> Apart from that, I am OK with the PR.

Temurin was not available at this point in time. The JDKs are normally all the same. There are no behavioral differences, all of them are OpenJDK rip-offs tested against the JCK. The differences manly are lying in the EoL or some additional libraries (Corretto has an additional crypto provider, Azul has integration OpenJFX). Useful resources seems to be https://whichjdk.com/#amazon-corretto and https://stackoverflow.com/questions/53305934/differences-amazon-corretto-and-openjdk.

Maybe temurin is available on Github Actions now.

PR #3287, opened 2023-09-27
SMT: add Z3_QF variant

FliegendeWurst:
This PR adds a variant of the Z3 SMT configuration which excludes quantifiers, making it much faster for certain problems. I will update this PR with my usecase if I can find it again.

Additionally the PR enhances the gradle task definitions by specifying inputs and outputs, which allows gradle to automatically respond to new .props files.

### DIFF
--- a/key.core/src/main/resources/de/uka/ilkd/key/smt/solvertypes/Z3_QF.props
+++ b/key.core/src/main/resources/de/uka/ilkd/key/smt/solvertypes/Z3_QF.props
@@ -1,7 +1,7 @@
 params=-in -smt2
 timeout=-1
 name=Z3_QF
-info=Z3 solver by Microsoft. Translation excludes quantifiers, making this variant much faster for certain problems.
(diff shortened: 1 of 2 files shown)


key.core/src/main/resources/de/uka/ilkd/key/smt/solvertypes/Z3_QF.props
```diff
@@ -0,0 +1,22 @@
+params=-in -smt2
+timeout=-1
+name=Z3_QF
+info=Z3 solver by Microsoft. Translation excludes quantifiers, making this variant much faster for certain problems.
```
WolframPfeifer, line 4: Since axioms from the preamble can still contain quantifiers (only sequent formulas with quantifiers do not get translated), we agreed that the description should reflect this.

WolframPfeifer (changes requested):
Thanks for adding this solver variant! Let's see if it proves useful in practice.

Please see change the description as we discussed, otherwise I am OK with this.

WolframPfeifer (approved):

PR #3289, opened 2023-09-28
UI: reduce number of redundant sequent draws

FliegendeWurst:
The purpose of this PR is to improve the performance of the KeY UI. Large sequents may take a few seconds to render, so it is vital to ensure they are only rendered when needed.

Test progress:

- [x] loading an example
- [x] manual rule app
- [x] automode to close proof
- [x] View → Use Pretty Syntax, Use Unicode Symbols, Use Syntax Highlighting, Term Labels, Hide Package Prefix toggling
- [x] prune
- [x] font size increase/decrease

Broken:

- [ ] macro from context menu (node view doesn't update)
- [ ] switching proofs (see #3347)

Unknown:
- [ ] creating/disabling an abbreviation
- [ ] one step back action
- [ ] sequent update after macro (with no steps applied)
- [ ] Select goal above / below

The PR also adds some debug logging to identify any issues caused by the changes in other scenarios.


FliegendeWurst (commented):
#3335 also modifies event handling and generation (may impact this PR)

FliegendeWurst:
> At the moment it just removed an unnecessary (as duplicate) selection vent of a changed proof

Yes, and that is also part of this PR. But I'm not 100% certain you did the same fix.

WolframPfeifer:
We are waiting for #3335 with this PR.

FliegendeWurst:
The changes in this PR are no longer necessary.

PR #3294, opened 2023-10-04
Update beanshell to 2.0b6

unp1:


Drodt (approved):

Drodt:
I just got an error just now while running some more tests on this branch:

The tests `testModelFieldTest_doubleX()` and `testModelFieldTest_f()` in `de.uka.ilkd.key.proof_references.testcase.analyst > TestClassAxiomAndInvariantProofReferencesAnalyst` fail. Both, essentially, due to the same error:

`AssertionFailedError: expected: <equals(test.ModelFieldTest::$f(heap,self),javaMulInt(Z(2(#)),int::select(heap,self,test.ModelFieldTest::$x)))> but was: <equals(test.ModelFieldTest::$f(heap,self),mul(Z(2(#)),int::select(heap,self,test.ModelFieldTest::$x)))>`

The other result looks similar.

I suspect that this is not actually due to this PR. Thoughts, @unp1 ?

wadoon (approved):

PR #3296, opened 2023-10-13
Backport of KEY-2.12.1 into main

wadoon:
Backporting the bug fixes in KEY-2.12.1 to main. This PR supersedes #3251 to avoid merge conflicts due to file header changes. Necessary changes were cherry-picked.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -50,7 +50,7 @@ static def getDate() {
 def build = System.env.BUILD_NUMBER == null ? "" : "-${System.env.BUILD_NUMBER}"
 
 group = "org.key-project"
-version = "2.13.0$build"
+version = "2.12.0$build"
 
 subprojects {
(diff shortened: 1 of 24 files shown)


build.gradle
```diff
@@ -50,7 +40,7 @@ static def getDate() {
 def build = System.env.BUILD_NUMBER == null ? "" : "-${System.env.BUILD_NUMBER}"
 
 group = "org.key-project"
-version = "2.13.0$build"
+version = "2.12.0$build"
```
WolframPfeifer, line 43:
```suggestion
version = "2.13.0$build"
```
We need to revert this, otherwise we would have the wrong version number on main, right?

WolframPfeifer (approved):
Maybe we should revert the version number change (see my suggestion), or would it be better to have a separate commit/PR on main?

PR #3299, opened 2023-10-16
Update nightlydeploy.yml

wadoon:
Setting an explicit Java version for the deployment

### DIFF
--- a/.github/workflows/nightlydeploy.yml
+++ b/.github/workflows/nightlydeploy.yml
@@ -24,6 +24,11 @@ jobs:
       #     fetchReviewers: true
 
       - uses: actions/checkout@v3
+      - name: Set up JDK 17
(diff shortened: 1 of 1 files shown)


.github/workflows/nightlydeploy.yml
```diff
@@ -24,6 +24,11 @@ jobs:
       #     fetchReviewers: true
 
       - uses: actions/checkout@v3
+      - name: Set up JDK 17
+        uses: actions/setup-java@v3
+        with:
+          java-version: ${{ matrix.java }}
```
WolframPfeifer, line 30: Copy-and-paste error when copying from tests.yml?
```suggestion
          java-version: 17
```

WolframPfeifer (approved):

PR #3302, opened 2023-10-17
Fix passing of warnings

wadoon:
No warning was shown if a name collision happens in the signature of a problem. Program variables take preceding on the evaluation. The check was present in the code, but the generated warning did not bubble up to the `ProblemInitializer`.

For example, consider following input file: 

```key
\functions        { int f; }
\programVariables { int f; }
\problem          { f = f  }
```

that loading ends up into a warning message now:

image


FliegendeWurst:
Why does the example
```
\functions        { int f; }
\programVariables { int f; }
\problem          { f = f  }
```
generate >9 warnings? I would expect at most 4.

wadoon:
> Why does the example
> 
> ```
> \functions        { int f; }
> \programVariables { int f; }
> \problem          { f = f  }
> ```
> 
> generate >9 warnings? I would expect at most 4.

The line is produced here: 

https://github.com/KeYProject/key/blob/c9ba634a4e03e44abc8b5703570f17bf03fa9978/key.core/src/main/java/de/uka/ilkd/key/nparser/builder/DeclarationBuilder.java#L160C1-L160C1


I guess we declare the sort multiple times.

wadoon:
@unp1 Maybe we should solve this at a larger scale. We had sometimes the discussion for theory support in KeY. [I copied my old notes but I need to sort them]( https://github.com/KeYProject/key/wiki/Better-Theories-in-KeY), then we would have local namespace by default.

wadoon:
@unp1 @mattulbrich 
Should we just disable the warning for sorts for now (It had no impact earlier.), so this PR can fly into `main`?

mattulbrich:
After discussion in KeY meeting: Program variable and function of same name should not only be a warning but an actual error.

wadoon:
@unp1 @mattulbrich The discussion of the KaKeY meeting is implemented. Error on shadowing of functions, and no warning on sort duplication. Both places are commented in the source code.

wadoon:
> Thanks. The code changes look good. I assume the failing tests is because of the stricter check for duplicate function names and we need to fix our test suite?

Surprisingly yes, 

```
Function 'head' is already defined! at <unknown>:3:3 (<unknown>:3:3)
```

in `TestTermParser`.

image

mattulbrich:
> @wadoon @mattulbrich: Should we also consider duplicate predicates as errors? At the moment that is not considered to be an error, e.g., geq in the example below:
> 
> // Input file for KeY standalone prover version 0.497 \sorts { s; }
> 
> \predicates { p; q; geq(int,int); }
> 
> \problem { ((p -> q) -> !q -> !p) & geq(0,1)
> 
> } [contraposition.key.zip](https://github.com/KeYProject/key/files/13713400/contraposition.key.zip)

I too would say that function and predicate symbols are to be treated equally here.

PR #3303, opened 2023-10-17
Create a JSON-RPC for KeY

wadoon:
# A remote KeY api as promised in KeY++

## Design Descisions

* We use handles to refer to large entities like, `KeYEnvironment`, `Proof`, or `Node`. 
  These handles are called `*Id` and are aligned hierarchical:
      
      EnvironemntId < ProofId < NodeId 

  If you have a `NodeId`, you can get a `ProofId` by calling `nodeId.proofId()`.

* We do not use expose any `key.core` classes. For example, Key's `Example`s are converted into `ExampleDesc` for the serialization. Every information holding class ends with `Desc`. 

* Given complex arguments (especially optional/nullable parameters) are packed into a class which ends with `Params`.

## TODO 

* [x] Implement a client in non-java, non-jvm (Python)


api.meta.md
```diff
@@ -0,0 +1,480 @@
...
+
+
+### client/showDocument (`server -> client`) 
+
+```
+Client.client/showDocument( arg0 : ShowDocumentParams ) -> ShowDocumentResult
+```
+
+
+### client/sm (`server ~~> client`)
```
Drodt, line 306: I would prefer verbose names. Avoid abbreviations like "sm".

api.meta.md
```diff
@@ -0,0 +1,480 @@
...
+
+```
+Client.client/showDocument( arg0 : ShowDocumentParams ) -> ShowDocumentResult
+```
+
+
+### client/sm (`server ~~> client`) 
+
+```
+Client.client/sm( arg0 : ShowMessageParams ) **async**
```
Drodt, line 309: Similarly, avoid generic argument names like `arg0`.

key.core.symbolic_execution/src/test/java/de/uka/ilkd/key/symbolic_execution/testcase/AbstractSymbolicExecutionTestCase.java
```diff
@@ -906,15 +904,15 @@ protected static void assertConstraint(IExecutionConstraint expected,
      * Makes sure that the given nodes contains the same {@link IExecutionVariable}s of the call
      * state.
      *
-     * @param expected The expected node.
-     * @param current The current node.
-     * @param compareVariables Compare variables?
+     * @param expected           The expected node.
```
Drodt, line 907: Is this a new spotless setting?

Drodt (commented):
I like the approach. It seems like a good setup.

For what languages do you want to write a client? What's on your TODO list?

api.meta.md
wadoon: Sorry, this is due to the lack of Java reflection. I think I need to compile with `-g` for debugging information. 

These documentation are generated from the Java information.

wadoon: Since Java 8, the `-parameters` switch ensures the parameter names in class files.

wadoon:
> For what languages do you want to write a client? What's on your TODO list?

I make a simple one in Python. The Python stubs are mostly generated by using Java Reflection. I am creating a JSON with meta information, markdown and Python classes. 

I don't know whether you want a rust client. In general, you need a background thread for reading messages, and a shared map from request ids to wait conditions/locks for block synchronous requests until return value is received. 

The Python API is more for testing, I am not planning a production-worthy state. 

Do we have a use case? For JVM programs I would still recommend to use KeY directly.

api.meta.md
wadoon: Yes, something from the beginning. 

I am still very unhappy with the client API. It is mainly the progress listener, which is not very good, especially with parallel processes.

wadoon:
*KaKeY:* Make this state fit for merging.

unp1:
I pushed some minot changes to fix some compilation bugs and enhanced the server side to allow for reconnects. The Java API work now (besides the rusty and solidity KeY versions) with KeY-tau (https://github.com/KeYProject/key-tau)

wadoon:
@unp1 Please, no merge of main into PRs. This makes things really complicated.

wadoon:
KaKeY: 2026-06-19 -- Externalize

wadoon:
Externalized into https://github.com/KeYProject/key-rpc
Remaining parts in #3849 

/close

PR #3305, opened 2023-10-19
Proof caching: use dependency graph to increase hit rate

FliegendeWurst:
Previously when searching for a reference, the entire sequent of the old proof had to be present in the new proof.
Now, only the useful subset of the old sequent has to be present.
Tested by a unit test.

Additional functionality:
- new "shortened chains" dependency graph rendering mode.
- button in toolbar (and checkbox in options menu) to disable proof caching
- new context menu action on nodes closed by cache to re-open them
- when a referenced proof branch is pruned (= open again), any references to that branch are automatically removed (reopening previously "closed by cache" proofs)


key.core/src/main/java/de/uka/ilkd/key/logic/Semisequent.java
```diff
@@ -29,15 +30,31 @@ private Semisequent() {
...
+     * sequent by subsequently inserting all formulas.
+     *
+     * @param seqList list of sequent formulas
+     */
+    public static Semisequent create(Collection<SequentFormula> seqList) {
+        if (seqList.isEmpty()) {
+            return EMPTY_SEMISEQUENT;
+        }
+        return new Semisequent(seqList);
     }
```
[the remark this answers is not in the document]
FliegendeWurst: Hi, just pushed another commit that restores the previous constructor (since `ImmutableList` doesn't implement `Collection`). That way it actually compiles successfully. And there is no performance hit.

keyext.slicing/src/main/java/org/key_project/slicing/analysis/AnalysisResults.java
```diff
@@ -139,6 +150,32 @@ public boolean indicateSlicingPotential() {
...
+                if (seq.numberInAntec(i)) {
+                    ante.add(seq.getFormulabyNr(i));
+                } else {
+                    succ.add(seq.getFormulabyNr(i));
+                }
+            }
+        }
+        return Sequent.createSequent(Semisequent.create(ante), Semisequent.create(succ));
+    }
+
```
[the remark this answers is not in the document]
FliegendeWurst: > How often is that method called?

It depends on the proof. Not too often, most of the time. Thank you for your suggestion, I will incorporate it into the code.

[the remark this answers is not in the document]
FliegendeWurst: By the way, the markdown syntax for code blocks is:

\```
public static void doIt() {
}
\```

keyext.slicing/src/main/java/org/key_project/slicing/ui/SlicingLeftPanel.java
```diff
@@ -287,6 +291,7 @@ private JPanel getDependencyGraphPanel() {
         panel1.setBorder(new TitledBorder("Dependency graph"));
 
         abbreviateFormulas = new JCheckBox("Abbreviate formulas");
+        abbreviateChains = new JCheckBox("Shorten long chains");
```
WolframPfeifer, line 294: I think this option is not self-explanatory. In particular, I wonder what the label of the new edge is if multiple edges are contracted to a single one. Could you provide here (and maybe also for "abbreviate formulas")?

keyext.slicing/src/main/java/org/key_project/slicing/graph/EquivalenceDirectedGraph.java
```diff
@@ -67,4 +67,15 @@ public Collection<GraphNode> getVerticesModProofIrrelevancy(GraphNode v) {
             return List.of(v);
         }
     }
+
+    public EquivalenceDirectedGraph copy() {
```
WolframPfeifer, line 71: I think it would be a good idea to override the clone() method here, since it is allowed to refine the return type covariantly.

WolframPfeifer (changes requested):
Thanks for this improvement! As already discussed, for easier generation of a meaningful changelog, the PR title could be improved. I have only two small remarks/requests (see comments in code).

keyext.slicing/src/main/java/org/key_project/slicing/graph/EquivalenceDirectedGraph.java
FliegendeWurst: clone() has `protected` access, it would not work here.

keyext.slicing/src/main/java/org/key_project/slicing/ui/SlicingLeftPanel.java
FliegendeWurst: I'll add some tooltips. Maybe a picture in key-docs too. For now, the edge label is simple the label of the last link in the chain.

For example `A  --1-->  B  --2-->  C  --3-->  D` would shorten to `A  --1-->  B  --3--> D`

FliegendeWurst: We discussed that `A  --1-->  B  --2-->  C  --3-->  D` should shorten to `A  --1-->  B  --2, ..., 3-> D`. I'll implement this change now.

FliegendeWurst: This is now implemented :)

FliegendeWurst:
> The only small suggestion from my side would be to remove the white box from the icon, such that it fits our icon design better.

This is now done.

WolframPfeifer (approved):

PR #3310, opened 2023-10-20
Provide additional String formating utility method to reduce code duplication

unp1:
Some minor cleanup with the side effect of providing a test suite for class Strings in key.util

### DIFF
--- a/key.util/src/main/java/org/key_project/util/Strings.java
+++ b/key.util/src/main/java/org/key_project/util/Strings.java
@@ -4,10 +4,11 @@
 package org.key_project.util;
 
 
-import java.util.Iterator;
 import java.util.function.Function;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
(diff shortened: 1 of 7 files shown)


key.util/src/main/java/org/key_project/util/Strings.java
```diff
@@ -53,4 +56,83 @@ public static boolean isJMLComment(String comment) {
...
+     *
+     * @param it the Iterator over the collection to be printed
+     * @param open the String used to open the list
+     * @param sep the String separating the different elements
+     * @param close the String used to close the list
+     * @param mapper a Function that maps the elements of type S to their String representation
+     * @return the CharSequence in the described format
+     * @param <S> the type of the elements of the iterated collection
+     */
+    public static <S, T> String formatAsList(Iterator<S> it,
```
wadoon, line 72: Why not using the simpler streaming API?

```java
iter.stream().map(a->mapper.apply(a))
.collecting(Collectors.joining(delim,pre,post);
```

wadoon (approved):

PR #3316, opened 2023-10-21
Allow origin labels to be switched off completely

### LINKED ISSUE #3315: Disabling OriginLabels causes proof to crash
## Description

> Please describe your concern in detail!

If one switches Term Origin tracking off during the proof (after some steps) and then restarts the strategies, the proof crashes.

The reason is that the existing sequents in the goals are changed (removal of the term labels). This does not work as many internal data structures keeping track of the proof state contain the original sequent formulas and terms, which afterwards can not be found anymore.

Suggestion:
The user should decide before loading the proof, if origin tracking should be enabled. If the user decided to disable it, then they should not be created (or introduced) at all.

One easy option to achieve this could look roughly as follows:

- Enforce all creations of OriginLabels to use a corresponding method of the OriginTermLabelFactory
- Move all uses of the OriginTermLabelFactory to TermBuilder methods that get the necessary details (not the complete labels) and which return the labelled term 
- The TermBuilder gets the factory (or null depending on the settings). If none is available it just does not create the label.
- This is still not completely nice, as the OriginTermLabels are still hard coded in, but afterwards it should be easier to find a general solution.

## Reproducible
  
> Is the issue reproducible?
> Select one of: always, sometimes, random, have not tried, n/a 

always

### Steps to reproduce 

1. Load the BinarySearch example
2. Let the prover run for a short while and press stop
3. Toggle the origin labels off
4. Continue wit the strategies
5. Exception

No exception. Toggling on or off should be a setting before loading the PO.

### Additional information

The error message on the console is:
at de.uka.ilkd.key.util.Debug.fail(Debug.java:105)
	at de.uka.ilkd.key.strategy.FindTacletAppContainer.<init>(FindTacletAppContainer.java:52)
	at de.uka.ilkd.key.strategy.TacletAppContainer.createContainer(TacletAppContainer.java:69)
	at de.uka.ilkd.key.strategy.TacletAppContainer.createInitialAppContainers(TacletAppContainer.java:197)
	at de.uka.ilkd.key.strategy.RuleAppContainer.createAppContainers(RuleAppContainer.java:116)

---

* Commit: ca0c52224ad45e8d7e98f8674c4f088caa7ff483

unp1:
This PR allows to fully disable origin label creation if tracking is disabled. Switching tracking of an reloading, does no longer introduce origin labels in the first place.

- OriginLabels are only created via OriginTermLabelFactory
- Services have an OriginTermLabelFactory
- TermBuilder adds origin labels only when services has a factory set
- Changing origin tracking requires reloading
- Proof object no longer depends on origin labels 
- Proof object no longer responsible for source view related initialization of relevant files (moved to source view)

The following should be considered for the midterm, but should not be a blocker for merging this PR:
- remove hard coded OriginLabel knowledge from services, termbuilder etc. and provide a more general framework (this item can also be postponed)

@Reviewers:
- any idea how to pass knowledge about origin tracking to the PreParser in a nicer manner?

Fixes #3315

### DIFF
--- a/key.ui/src/main/java/de/uka/ilkd/key/gui/originlabels/ToggleTermOriginTrackingAction.java
+++ b/key.ui/src/main/java/de/uka/ilkd/key/gui/originlabels/ToggleTermOriginTrackingAction.java
@@ -4,8 +4,7 @@
 package de.uka.ilkd.key.gui.originlabels;
 
 import java.awt.event.ActionEvent;
-import javax.swing.Action;
-import javax.swing.JOptionPane;
+import javax.swing.*;
 
 import de.uka.ilkd.key.gui.MainWindow;
 import de.uka.ilkd.key.gui.actions.KeyAction;
@@ -13,7 +12,6 @@ import de.uka.ilkd.key.gui.actions.MainWindowAction;
 import de.uka.ilkd.key.gui.actions.QuickLoadAction;
 import de.uka.ilkd.key.gui.actions.QuickSaveAction;
 import de.uka.ilkd.key.gui.fonticons.IconFactory;
-import de.uka.ilkd.key.proof.Proof;
 import de.uka.ilkd.key.settings.ProofIndependentSettings;
 import de.uka.ilkd.key.settings.TermLabelSettings;
 
@@ -23,8 +21,6 @@ import de.uka.ilkd.key.settings.TermLabelSettings;
  * @author lanzinger
  */
 public class ToggleTermOriginTrackingAction extends MainWindowAction {
-    private static final long serialVersionUID = -2092724865788720558L;
-
     /**
      * Create a new action.
      *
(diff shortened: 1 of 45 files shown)


wadoon (commented):
I am not able to see how to turn it on and off. 

Could you provide some information to it.

key.ui/src/main/java/de/uka/ilkd/key/gui/originlabels/ToggleTermOriginTrackingAction.java
```diff
@@ -55,61 +52,47 @@ private void handleAction() {
         TermLabelSettings settings =
             ProofIndependentSettings.DEFAULT_INSTANCE.getTermLabelSettings();
 
-        if (proof != null) {
-            Services services = proof.getServices();
+        /*
```
wadoon: I would say that `handleAction()` is now useless and can be removed

wadoon (approved):

wadoon (approved):

PR #3318, opened 2023-10-22
Fix for enabling runAllProofs restricted to specific groups

unp1:
- Add missing setRunOnlyOn(...) method to enable running only defined groups of test proofs


FliegendeWurst (approved):
This is really useful when running runAllProofs manually! :+1:

PR #3321, opened 2023-10-24
Optimize term label updates

unp1:
- TermLabel refactorings container can use mutable linked sets
- Optimized TermLabel processing using a linked hash set for common operations instead of a list
- Avoid unnecessary creation of equal terms leading to additional goal updates


wadoon (approved):

wadoon:
@unp1 Somebody has to resolve the merge conflicts :(

Issue #3297, opened 2023-10-13
Node selection after loading a proof


unp1:
## Description

Loading an open proof selects the root node instead of an open goal.
I am not sure if that may be intended behavior, because one can be of course the 
opinion that this is the better way.

In any case, for teaching I sometimes load proofs that have just one branch open and before that branch would be
directly selected, which makes the workflow a bit easier.

In any case, if the change was intended, please close the bug with won't fix.

## Reproducible

always

### Steps to reproduce 

> Describe the steps needed to reproduce the issue.

1. Save an open proof (with on goal open)
2. Load the proof
3. Root is selected after loading 

My expectation would be that the open goal is selected.

### Additional information

> Add more details here. In particular: if you have a stacktrace, put it here.

---

* Commit: 05635afb51f8125c62fa9338fdc8f3978ce672a5

FliegendeWurst:
I agree that selecting an open goal is more useful than selecting the root node. Don't know why it was changed.
I think the fix would be a call to `mediator.getSelectionModel().defaultSelection()` after loading the proof.

PR #3324, opened 2023-10-27
Feature: If possible save the last selected node in the proof, to reselect it upon loading.

unp1:
This commit addresses issue #3297.


wadoon:
> Sadly not as we do no save unique node identifiers in our proofs and the node numbers upon reload may differ (e.g. when the proof was pruned at some point).

Sarah and I have used a slightly different system to identify nodes uniquely. We used the path from root to the node in the same manner as PiO and PiS. [See here](https://github.com/wadoon/key-interactionlog/blob/e459d70d327bc44ecae27ee2a875d0d9074d5c59/src/main/kotlin/org/key_project/ui/interactionlog/model/interactions.kt#L198C15-L198C15). Of course, if you apply run-length encoding you can save a lot of entries in the list (mostly zero, and some ones).

PR #3325, opened 2023-10-28
A heap indicator in KeY's status bar

wadoon:
This PR brings two things for a clear indication of the current heap size settings. 

1. The logging level for heap information is increased from debug to info in `Main#main()`.
   Now the line 

   ```
    [14:33:40.367] INFO  Main - Memory: total 250.0 MB, max 3948.0 MB, free 236.08212280273438 MB
   ```
   should always appear. 

2. A GUI extension for the status bar: 

   image

   which allows to trigger `System.gc()` and to start the `jconsole` monitoring tool.

### DIFF
--- a/key.ui/src/main/java/de/uka/ilkd/key/gui/utilities/HeapStatusExt.java
+++ b/key.ui/src/main/java/de/uka/ilkd/key/gui/utilities/HeapStatusExt.java
@@ -1,12 +1,8 @@
+/* This file is part of KeY - https://key-project.org
+ * KeY is licensed under the GNU General Public License Version 2
+ * SPDX-License-Identifier: GPL-2.0-only */
 package de.uka.ilkd.key.gui.utilities;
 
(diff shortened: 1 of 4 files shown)


key.ui/src/main/java/de/uka/ilkd/key/gui/utilities/HeapStatusExt.java
```diff
@@ -0,0 +1,158 @@
...
+
+            int max = (int) (rt.maxMemory() / mb);
+            int free = (int) (rt.freeMemory() / mb);
+            int total = (int) (rt.totalMemory() / mb);
+
+            progressBar.setMaximumSize(new Dimension(100, 25));
+
+            progressBar.setMaximum(total);
+            final var used = total - free;
+            progressBar.setValue(used);
```
FliegendeWurst, line 85: Doesn't really make sense to me. I think this should be `used / total`.
image

PR #3345, opened 2023-11-10
Minor refactoring of Proof to reduce dependencies (for instance to MergeRule)

unp1:
This commit factors out pruning, saving and reference resolving functionality to reduce the dependencies of Proof and to make it later easier to realise correct synchronization as well as saving proofs.

Not for this PR, but we should have a discussion about what our proof object should have as basic data structure. My idea:
- merge references and linked goals to once concept independent of merge rule or proof caching
- allow to register reference resolvers that can deal with references
- ask before saving (in the GUI) if references should be resolved
  - if yes, resolve and then pass the proof to the `OutputStreamProofSaver`
  - if not, the proof is passed to `OutputStreamProofSaver` which just saves the references (and does not resolve)
  - this allows to remove the line 
        `CopyReferenceResolver.copyCachedGoals(proof, null, null, null);`
   from `OutputStreamProofSaver`. As the saver should not have specific knowledge and it should not change the proof object while saving

### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/proof/reference/CopyReferenceResolver.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/proof/reference/CopyReferenceResolver.java
@@ -1,0 +1,63 @@
+/* This file is part of KeY - https://key-project.org
+ * KeY is licensed under the GNU General Public License Version 2
+ * SPDX-License-Identifier: GPL-2.0-only */
+package de.uka.ilkd.key.proof.reference;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Consumer;
+
+import de.uka.ilkd.key.proof.Goal;
+import de.uka.ilkd.key.proof.Node;
+import de.uka.ilkd.key.proof.Proof;
+import de.uka.ilkd.key.proof.io.IntermediateProofReplayer;
+import de.uka.ilkd.key.proof.replay.CopyingProofReplayer;
+
+public class CopyReferenceResolver {
+    /**
+     * For each branch closed by reference to another proof,
+     * copy the relevant proof steps into this proof.
+     *
+     * @param toComplete the proof whose references to closed branched shall be resolved by
+     *        copying the branches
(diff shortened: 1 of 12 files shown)


key.core/src/main/java/de/uka/ilkd/key/proof/reference/CopyReferenceResolver.java
```diff
@@ -0,0 +1,63 @@
...
+        // first, ensure that all cached goals are copied over
+        List<Goal> goals = toComplete.closedGoals().toList();
+        List<Goal> todo = new ArrayList<>();
+        for (Goal g : goals) {
+            Node node = g.node();
+            ClosedBy c = node.lookup(ClosedBy.class);
+            if (c == null) {
+                continue;
+            }
+            if (referencedFrom != null && referencedFrom != c.proof()) {
```
Drodt, line 41: Does `referencedFrom == null` ever happen? If so, should we return before the loop when it's null?

FliegendeWurst: It is documented above: `@param referencedFrom filter, if not null copy only from that proof`
(why do I even bother to write javadocs if nobody reads them anyhow...)

`referencedFrom == null` means to copy from all referenced proofs.

FliegendeWurst:
> Not for this PR, but we should have a discussion about what our proof object should have as basic data structure. My idea:
> 
>    * merge references and linked goals to once concept independent of merge rule or proof caching
> 
>    * allow to register reference resolvers that can deal with references
> 
>    * ask before saving (in the GUI) if references should be resolved
>       
>      * if yes, resolve and then pass the proof to the `OutputStreamProofSaver`
>      * if not, the proof is passed to `OutputStreamProofSaver` which just saves the references (and does not resolve)

I can see how to resolve a cached proof reference (just copy the steps). But I don't really understand the difference between *resolving* and *saving* a linked goal yet. A linked goal is always "resolved" by the branch it is linked to, right?

wadoon (approved):

PR #3346, opened 2023-11-12
CI: make codecov checks non-failing

FliegendeWurst:
Insufficient test coverage shouldn't lead to a red X in the PR list. The :x: should be reserved for more urgent problems (formatting / compilation / failing tests) that actually block merging.

Docs for codecov.yml: https://docs.codecov.com/docs/commit-status


Drodt:
> Is it possible to just raise the percentage bar from which on a red cross is used. Setting it to 0.5 % or 1% would get rid of all the usual ones, but still highlight more severe losses of coverage?

There is a `threshold` option, which should work for our case: https://docs.codecov.com/docs/github-3-customizing-codecov

FliegendeWurst:
> There is a `threshold` option, which should work for our case: [docs.codecov.com/docs/github-3-customizing-codecov](https://docs.codecov.com/docs/github-3-customizing-codecov)

Good idea. I think target: 38% and threshold: 1% would work well.

wadoon (approved):

FliegendeWurst:
Hold up, I didn't change it to 38% +- 1 yet. But that can be done in a follow-up PR if it's desired ..

Issue #3347, opened 2023-11-12
Switching proofs resets selection


FliegendeWurst:
## Description

When switching proofs, the selected node is not restored correctly.

## Reproducible
  
always

### Steps to reproduce 

1. Load two proofs
2. Select a node X in proof A
3. Switch to proof B
4. Switch to proof A

Expectation: I see node X
Actual result: I see some other node

### Additional information

Probably a regression from #3341

---

* Commit: ae6eface4142f11e8690b0c89b2cdf8ab78d7f96

PR #3349, opened 2023-11-13
Fix (for issue #3347) that node selection gets forgotten when switching between proofs

unp1:


FliegendeWurst (changes requested):
Still broken, but now it only updates the sequent view to show a closed goal, and not the proof tree.

FWIW I fully expected these regressions when I originally made changes to the selection event handling in #3289. This area of KeY is extremely fragile.

key.ui/src/main/java/de/uka/ilkd/key/core/KeYSelectionModel.java
```diff
@@ -17,12 +17,8 @@
 public class KeYSelectionModel {
     private static final Logger LOGGER = LoggerFactory.getLogger(KeYSelectionModel.class);
 
-    // the listener to be informed before all
-    // this is the KeYMediator listener and should be removed later
-    // at the moment it is used to ensure a sane state of the mediator
-    // before other selection listeners receive change information
-    // TODO: remove usage of primary
```
FliegendeWurst, line 24: Does this PR resolve the TODO?

FliegendeWurst (approved):
Issue is fixed. Code changes look good. But it is hard to check that all the removed setSelectedProof calls etc. were really not needed.

Drodt (approved):

PR #3027, opened 2023-02-06
Overflow checking

jwiesler:
This is a follow up to #3014.

This MR implements overflow checking of all arithmetic operations in code by opening a new goal where the absence of overflows has to be proven.

Furthermore, it deduplicates and rearranges a lot of rules.

There are some todos where I was unsure, if you know an answer to any of them please add a comment.

PR #3353, opened 2023-11-15
Fix overflow checking

### LINKED ISSUE #3352: Taclet option for checking overflows (in Java code) does not work
## Description

In the current version of KeY, the integer semantics with explicit overflow checks as introduced by #3027 are broken on at least 3 levels:
1) The annotation `/*@ code_safe_math */ is ignored and does not change the taclet option (as I expected). However, it is parsed without an error and seems to be simply ignored.
2) After manually setting the taclet option via the menu (and starting the proof again as always), the option for checking overflows is completely ignored, no additional branches for overflow checks are created. Proofs created with `intRules:arithmeticSemanticsCheckingOF` are exactly identical to those with `intRules:arithmeticSemanticsIgnoringOF`. In the rule files of key, introduction of the additional branches is guarded by `intRules:checkedOverflow`, which of course does not exist. Using an undeclared taclet option in any rule file should really throw an error in the GUI!
3) After repairing point 2, the additional branches for overflow checks are generated, but instantly closed, since `inRange(...)` is always rewritten to true!

Furthermore, there is no documentation of the final solution (description of what has been implemented in #3014 and #3027). The implemented solution deviates from the description in the wiki (https://github.com/KeYProject/key/wiki/Spec-math-modes). We really should have a user-level documentation in key-docs.

I am a bit shocked that such a major point of KeY was apparently not even tested manually and survived for half a year. We really should have unit tests here, e.g. a small proof obligation (in a key file) for each of the options that is only provable if the correct semantics are used.

## Reproducible

always

### Steps to reproduce 

1. Load the following class:
  ```java
/*@ spec_bigint_math */
/*@ code_safe_math */   // this seems to be ignored by KeY
class Overflow {
    /*@ normal_behavior
      @  ensures \result == x + y;
      @*/
    int add(int x, int y) {
        return x + y;
    }
}
  ```
2. Start the proof of the contract.
3. Run the automatic proof search. 

The proof is closed instantly, while it should have an additional (unprovable) branch containing the overflow check for `+`. This is not the case due to point 1 in the above list.

4. Manually set the taclet option for overflow checking via the GUI (workaround for point 1).
5. Reload the file and start the proof.
6. Run the automatic proof search.
 
Again, the proof is closed. Now it contains the additional branch for overflow checking. However, this branch is closed, which should not be the case, since the specification uses (unbounded) bigints, while the code potentially has an overflow.

WolframPfeifer:
The purpose of this PR is to repair the overflow checking of KeY and thus fix #3352.

TODO:
* [x] fix creation of branches for overflow checking
* [x] fix the actual proof obligation of the overflow check (should use `inInt(...)`, `inRangeInt(...)`, taclets in `inRules(Un)CheckedSemantics.key`)
* [x] throw an error if a choice used in a taclet does not exist


WolframPfeifer:
While working on this PR, I stumbled upon some unclear points:
* Does the `code_bigint/safe/java_math` annotation belong to the contract or to the method?
* Is it allowed to have multiple contracts with different code math modes?

In addition, I am not sure what syntax is currently supported by KeY's parser ...

Does anyone know the status quo? How should it be implemented?

WolframPfeifer:
I am currently struggling with the assignmentToUpdate rules.

The original rule (i.e., status quo in KeY since #3027, with fix for the wrongly named taclet option) for e.g. subtraction looks like this:
```
    assignmentSubtractionInt {
        \find(\modality{#normalassign}{..
                    #loc = #seCharByteShortInt0 - #seCharByteShortInt1;
                ...}\endmodality (post))
        (intRules:arithmeticSemanticsCheckingOF) {
            "Overflow check":
                \replacewith(inInt(sub(#seCharByteShortInt0, #seCharByteShortInt1)))
        };
        \replacewith({#loc := javaSubInt(#seCharByteShortInt0, #seCharByteShortInt1)}
            \modality{#normalassign}{.. ...}\endmodality (post))
        \heuristics(executeIntegerAssignment)
        \displayname "subtraction"
    };
```
 As @unp1 pointed out, if the modality is in the antecedent, you probably do not want to open a branch with the overflow "check" in the antecedent, as this leads to an unclosable branch usually.

If you would like to prevent that, you would have to duplicate all the rules (about 30 at the moment): 
```
    assignmentSubtractionInt {
        \find(\modality{#normalassign}{..
                    #loc = #seCharByteShortInt0 - #seCharByteShortInt1;
                ...}\endmodality (post))
        \succedentPolarity
        (intRules:arithmeticSemanticsCheckingOF) {
            "Overflow check":
                \replacewith(inInt(sub(#seCharByteShortInt0, #seCharByteShortInt1)))
        };
        \replacewith({#loc := javaSubInt(#seCharByteShortInt0, #seCharByteShortInt1)}
            \modality{#normalassign}{.. ...}\endmodality (post))
        \heuristics(executeIntegerAssignment)
        \displayname "subtraction"
    };

    assignmentSubtractionInt2 {
        \find(\modality{#normalassign}{..
                    #loc = #seCharByteShortInt0 - #seCharByteShortInt1;
                ...}\endmodality (post))
        \antecedentPolarity
        \replacewith({#loc := javaSubInt(#seCharByteShortInt0, #seCharByteShortInt1)}
            \modality{#normalassign}{.. ...}\endmodality (post))
        \heuristics(executeIntegerAssignment)
        \displayname "subtraction"
    };
```
The open question is: Do we have to add another branch in the antecedent case to be able to assume that `inInt(a - b)`? Or do we have to check it? I really do not understand what would be correct here ...

Another option without the rule duplication would be to just limit application to the succedent. This has been done for example for the rule `assignmentDivisionInt` with the `runtimeExceptions:ban` semantics. However, this is also strange: There is just no rule to further symbolically execute a division (with this semantics selected) in a modality on the left-hand side.

key.core/src/main/java/de/uka/ilkd/key/nparser/builder/TacletPBuilder.java
```diff
@@ -681,6 +681,17 @@ public ChoiceExpr visitOption_expr_paren(KeYParser.Option_expr_parenContext ctx)
 
     @Override
     public ChoiceExpr visitOption_expr_prop(KeYParser.Option_expr_propContext ctx) {
+        String category = ctx.option().cat.getText();
+        String value = ctx.option().value.getText();
+        String choiceStr = category + ":" + value;
+        /*
```
wadoon, line 687: This comment is not helpful because the required action is unknown.

key.core/src/main/resources/de/uka/ilkd/key/proof/rules/intRulesUncheckedSemantics.key
```diff
@@ -50,6 +50,9 @@
         \replacewith(true)
         \heuristics(concrete)
     };
+}
+
+\rules(programRules:Java & (intRules:arithmeticSemanticsIgnoringOF | intRules:arithmeticSemanticsCheckingOF)) {
```
wadoon, line 55: This leads to the question of whether we might want to have rules predicate so that we can easily model "X" or "not X" in the Taclet base.

wadoon (approved):

key.core/src/main/java/de/uka/ilkd/key/nparser/builder/TacletPBuilder.java
```diff
@@ -681,6 +681,17 @@ public ChoiceExpr visitOption_expr_paren(KeYParser.Option_expr_parenContext ctx)
...
+         * Make sure that the choice (category and value!) is known to KeY. This prevents from
+         * accidentally deactivating (parts of) taclets due to non-existing choices (see
+         * https://github.com/KeYProject/key/issues/3352).
```
WolframPfeifer, line 690:
```suggestion
         * Make sure that the choice (category and value!) is known to KeY, i.e. that it is declared
         * in the file `optionsDeclarations.key`. This prevents from accidentally deactivating
         * (parts of) taclets due to non-existing choices (see
         * https://github.com/KeYProject/key/issues/3352).
```

key.core/src/main/resources/de/uka/ilkd/key/proof/rules/intRulesUncheckedSemantics.key
WolframPfeifer: I do not understand your comment. Is it a concrete suggestion for this PR? If not, I would like to postpone it, since we should not wait any longer with the overflow fix.

PR #3355, opened 2023-11-18
Avoid name clashes in rule applyUpdateOnRigid

### LINKED ISSUE #3254: KeY does not handle name clashes for logical variables correctly
## Description

It is possible to have a term with a logical variable whose name clashes with another bound variable in scope

The term `\forall int a; {x:=a}\forall java.lang.Object a; a = x` can be simplified to `\forall int a; \forall java.lang.Object a; a = a` using update simplification (without one step simplifier). In this term the two `a` refer to different variables and `a = a` is actually equivalent to false (due to the disjointness of their types).

## Reproducible
  
always

### Steps to reproduce 

Load the attached file [problem.proof.txt](https://github.com/KeYProject/key/files/12393642/problem.proof.txt) (remove .txt)

There should not be two different bound variables named identically within one scope; but there are.

I assume something similar could be arranged with a constant `a` instead of a bound variable.

### Additional information
image

becomes

image

---

* Commit: 2.12.0

tobias-rnh:
Fix #3254 
To prevent name clashes with bound variables, the free variables in the update and the bound variables of a formula are compared.
In case of a clash, the bound variable is renamed and a substitution is performed on the subterms of the formula automatically before applying the update.

grafik
now becomes
grafik

The new clash free name is created by appending a counter to the original name of the bound variable.
However: as only the variables contained in the update and the formula the update is applied on are considered, the following situation is still possible:

grafik
still becomes
grafik
 
Currently, there is no way to get names on higher levels than the update. So if the different meanings of ```a1``` in this situation are still of concern, it might be needed to find a solution on a higher level than ApplyUpdateOnRigidCondition, e.g. when parsing.

Tests for ApplyUpdateOnRigidCondition were also added.


Drodt (approved):
The code looks good but I will hold off on merging until @mattulbrich gives the go-ahead

mattulbrich (approved):
The step explained above certainly increases usability and syntactic clarity of KeY.
Thanks!

PR #3356, opened 2023-11-19
remove spin lock

wadoon:


wadoon:
I removed the spin lock also from `DefaultProofControl`. It is rather a simple change, I did not try to avoid code duplication. 

We have plenty of `Thread.sleep`s in the test cases. And there is a strange class `Watchdog`, which says it tries to detect deadlocks, but this makes no sense because Java has a built-in mechanism for deadlock detection.

FliegendeWurst:
> And there is a strange class `Watchdog`, which says it tries to detect deadlocks, but this makes no sense because Java has a built-in mechanism for deadlock detection.

The class is supposed to detect when the GUI thread is blocked. That's not automatically handled by Java.

Issue #3362, opened 2023-11-28
Assignable prevents proving the algorithm


fab918:
## Description

When I use "assignable" in the Key Editor, I have my proof which fails, yet the method does not modify anything. If assignable is removed, it works. This doesn't seem to limit to "\nothing", if I put array or array[*] it still doesn't work.

(I take advantage of this, is it possible to have more information on what is not working? I did not find any information on this subject, it is not clear where the problem lies)

## Reproducible
  
yes

### Steps to reproduce 

```java
public class Test {
    /*@ requires array != null && array.length > 0;
      @ ensures \result == (\sum int i; 0 <= i && i < array.length; array[i]);
      @ assignable \nothing;
      @*/
    public static int sum(int[] array) {
        int total = 0;
        
        //@ maintaining 0 <= i && i <= array.length;
        //@ maintaining total == (\sum int j; 0 <= j && j < i; array[j]);
        //@ decreases array.length - i;
        for (int i = 0; i < array.length; i++) {
            total += array[i];
        }
        return total;
    }
}
```

> What is your expected behavior and what was the actual behavior?

pass the test

mattulbrich:
Hi fab918, thanks for the report.

Try adding `//@ assignable \nothing;` to the loop specification (like above `decrease`).
That should help.

mi-ki:
I had implemented such an alias some years ago, but apparently this did not survive the release cycles. Shouldn't be a problem to reintroduce it though, in my opinion.

fab918:
@mi-ki Is there any trace of your contribution? I did a search on PRs but I don't see it.

mi-ki:
@fab918 I just took a look, but found out I did not introduce an alias for `loop_modifies` or `loop_writes`, but for `loop_variant` and `assigns` (yet another standard from somewhere else, apparently). This was commit af97f72d1994cb50c74eb04c77649f198236f8ec. However, I am not sure whether their translation still works as it used to do, since many hard-to-follow things happened to the parser in-between.

fab918:
@mi-ki Thanks, unfortunately I don't think it always works.

 If anyone has any advice on how to get started, I'd love to hear it.

mi-ki:
@fab918 I created a new branch which should hopefully fix the previously lost changes and also add the two keywords from OpenJML. Maybe, you can test whether it works for you?

mi-ki:
It's been a while and after various internal discussions, the PR #3365 made it onto the main branch.
[...]

PR #3365, opened 2023-11-28
Introduce JML aliases of frame conditions for better tool compatibility

mi-ki:
This pull request allows handling various aliases for frame conditions within method/block/loop contracts and loop specifications. These aliases cover the ones which are used in various other verification tools with contracts for Java-like programs, e.g., Krakatoa, OpenJML, Dafny, ACSL++, or CorC.

In particular, the following aliases are henceforth supported for frame conditions in method, block, and loop contracts as well as loop specifications:

`assignable`, *`assigns`*, *`assigning`*, `modifiable`, *`modifies`*, *`modifying`*, `writable`, *`writes`*, *`writing`*.

Upon usage of the italic versions, we throw a warning in order to distinguish expected semantics, e.g., from runtime verification, as KeY does static analysis, but (other than potentially seeing a warning) these versions are supported in the same way as the other aliases.

For loop specifications, we also support the respective aliases with the prefix `loop_`.

During discussions on this pull request, it was noted that the semantics implemented within KeY might be better captured by `modifiable`, but for legacy reasons, we do not move away from (also) using `assignable`.

## Related Issue

This pull request addresses #3362.

## Intended Change

This pull request introduces the aliases for frame conditions which are shown above. Thereby, it increases compatibility with further tools using JML, JML variants, or JML-alikes, such as Krakatoa, OpenJML, Dafny, or ACSL++.

## Type of pull request

- [ ] Refactoring (behavior should not change or only minimally change)
- [ ] Other: 

## Ensuring quality
    
- [ ] I made sure that introduced/changed code is well documented (Javadoc and inline comments).
- [x] I have tested the feature as follows: Checked it for the example from the referenced issue.

Some of the changes are results from work at the HacKeYthon together with @unp1.


fab918:
I tried the following example:

```
    /*@ requires array != null && array.length > 0;
      @ ensures \result == (\sum int i; 0 <= i && i < array.length; array[i]);
      @ assignable \nothing;
      @*/
    public static int sum(int[] array) {
        int total = 0;
        
        //@ maintaining 0 <= i && i <= array.length;
        //@ maintaining total == (\sum int j; 0 <= j && j < i; array[j]);
        //@ decreases array.length - i;
        //@ loop_writes \nothing;
        for (int i = 0; i < array.length; i++) {
            total += array[i];
        }
        return total;
    }
```

**In the KeY-2.12.2 (2023-11-10) release:** this file can't be loaded:

```
de.uka.ilkd.key.speclang.translation.SLTranslationException: Could not translate JML specification. You have either tried to use an unsupported keyword (loop_writes) or a JML field declaration without a ghost or model modifier.  Caused by: java.lang.RuntimeException: de.uka.ilkd.key.speclang.translation.SLTranslationException: Could not translate JML specification. You have either tried to use an unsupported keyword (loop_writes) or a JML field declaration without a ghost or model modifier.
```

**In the PR**: the file can be loaded and you are correctly returned to the main interface.

**However, the proof doesn't work,** it looks like the annotation is just skipped. Besides, it's not in bold as you can see in the screenshot, so I guess you need to add something else for the alias to work.

Screenshot 2023-11-28 at 20 52 37

WolframPfeifer (commented):
First of all, I think it is a good idea to be able to parse these keywords. However, I am not sure if we want to allow these as aliases (at least not without a warning). Intuitively, `modified` (**is** changed by the loop/method) and `modifiable` (**may be** changed) have slightly different semantics. Since KeY uses the latter, I am not sure if it is a good idea to allow these aliases.

In this PR, `writes` and `loop_assigns`, 

The original JML reference manual is old and outdated; as far as I can see, there is no framing clause for loops at all. The new one is still unfinished (and will take a while), I could not find a conclusion on the allowed keywords for framing clauses (neither [in the PDF](https://www.openjml.org/documentation/JML_Reference_Manual.pdf) nor in the [discussion on GitHub](https://github.com/JavaModelingLanguage/RefMan/issues)). I think that we should come up with our own consistent concept here. My suggestion would be:

1. framing clauses for methods:
    * `assignable`, `modifiable`, `writable`
    * `assigns`, `modifies`, `writes` (*)
    * I would **not** support `assigning` etc.
2. framing clauses for loops:
    * all of the above
    * all of the above with additional `loop_` prefixes (although personally I prefer the variants without prefix)

For (*) and the variants with prefix, there should be a warning, since KeY uses modifiable/assignable semantics. Actually, one could argue that there is even a subtle difference between those two (as far as I remember also mentioned in the KeY book in a gray box). However, that would probably go too far ...

Btw.: The clauses still need to be translated to logic, I think this has to be done somewhere in JMLSpecFactory/TextualJMLSpecCase.

WolframPfeifer:
Btw.: The support of syntax highlighting in the SourceView panel (screenshot above) is completely independent, as it is done by a handwritten highlighter and has its own list of supported keywords. Not ideal, but very difficult to do better ...

WolframPfeifer:
Since the reason for the failed proof attempt in #3362 was the missing framing clause, it would probably make sense to show a warning if no framing clause is present (default is `assignable \everything`, which is rarely what you want).

mattulbrich:
Or take the method's assignable clause as a default like a number of
other tools do. It is a good default.

key.core/src/main/java/de/uka/ilkd/key/speclang/njml/JmlTermFactory.java
```diff
@@ -920,14 +920,14 @@ public Term createLocSet(ImmutableList<SLExpression> exprList) {
                         singletons = singletons.append(t);
                     } else if (t.op() instanceof ProgramVariable) {
                         // this case may happen with local variables
-                        exc.addIgnoreWarning("local variable in assignable clause");
+                        exc.addIgnoreWarning("local variable in modifiable clause");
                         LOGGER.debug("Can't create a locset from local variable " + t + ".\n"
```
WolframPfeifer, line 924:
```suggestion
                        LOGGER.debug("Cannot create a locset from local variable " + t + ".\n"
```
Just to be consistent with the error messages ;-)

key.core/src/main/java/de/uka/ilkd/key/speclang/njml/Translator.java
```diff
@@ -2531,6 +2552,22 @@ protected void addWarning(ParserRuleContext node, String description) {
         exc.addWarning(description, node.start);
     }
 
+    private void warnPotentiallyUnintendedFramingSemantics(
+            ParserRuleContext ctx, String clauseName) {
+        clauseName =
+            clauseName.startsWith("loop_") ? clauseName.replaceFirst("loop_", "") : clauseName;
+        for (final String s : DISCOURAGED_CLAUSE_NAMES) {
+            if (clauseName != null && clauseName.startsWith(s)) {
+                addWarning(ctx, clauseName + " does not conform to KeY's supported JML language, "
```
WolframPfeifer, line 2561: This message imo reads a bit as if we would support the "one and only" JML (in particular, the "other JML-like languages" part). I think the message would sound better if it talked about "JML dialects" or similar.

key.core/src/main/java/de/uka/ilkd/key/speclang/njml/Translator.java
```diff
@@ -65,6 +65,9 @@
  */
 class Translator extends JmlParserBaseVisitor<Object> {
 
+    private static String[] DISCOURAGED_CLAUSE_NAMES =
```
WolframPfeifer, line 68: Should this not be final?

key.core/src/main/java/de/uka/ilkd/key/speclang/jml/translation/JMLSpecFactory.java
```diff
@@ -628,8 +628,8 @@ private Term translateUnionClauses(Context context, ImmutableList<ProgramVariabl
             if (translated.equalsModIrrelevantTermLabels(tb.strictlyNothing())) {
                 if (originalClauses.size() > 1) {
                     throw new SLTranslationException(
-                        "\"assignable \\less_than_nothing\" does not go with other "
-                            + "assignable clauses (even if they declare the same).",
+                        "\"modifiable \\less_than_nothing\" cannot be joined with other "
```
WolframPfeifer, line 631: I think I have never seen this message so far. However, I think it would be a good idea to explicitly say `\\strictly_nothing` here (or are there multiple different `\\less_than_nothing` terms?).

WolframPfeifer (commented):
Thanks, apart from some minor comments, I think that the changes are good. One thing I noticed though: Could it be the case that the performance of the parser dropped? It seems that loading the examples takes longer now (I did not measure, just some impression).

In addition, I put the PR onto the discussion list for our developer meetings (KaKeY), since I am not sure if moving towards "modifiable" is a good idea after having "assignable" basically as default for such a long time (even if it is only in code for now). Maybe we should discuss if everybody is ok with that.

key.core/src/main/java/de/uka/ilkd/key/speclang/jml/translation/JMLSpecFactory.java
mattulbrich: `\less_than_othing` is the old name of `\strictly_nothing`

key.core/src/main/java/de/uka/ilkd/key/speclang/njml/JmlTermFactory.java
mattulbrich: Would we really make modifyable the new default keyword?

mattulbrich:
While the title says this PR introduces new alias names, it seems that it changes the default keyword.
I am not sure we all agree on doing that ...

mi-ki:
> While the title says this PR introduces new alias names, it seems that it changes the default keyword. I am not sure we all agree on doing that ...

Thanks for your feedback! Can you please say what you mean by _default_ keyword? I did not implement any automatic specification generation/inference or similar, and this PR is really only about aliases.

What you may be referring to, is how some variables in KeY's Java code are named. Here, I tried to make the naming more consistent and honest (and hence: less misleading) about what KeY actually implements (see, e.g., also [https://dafny.org/latest/DafnyRef/DafnyRef#sec-modifies-clause](https://dafny.org/latest/DafnyRef/DafnyRef#sec-modifies-clause)) but this does not affect the user, I believe.

key.core/src/main/java/de/uka/ilkd/key/speclang/jml/translation/JMLSpecFactory.java
mi-ki: Thanks for the feedback, I fixed it (and also renamed one obsolete mention of the old version).

key.core/src/main/antlr4/JmlParser.g4
```diff
@@ -175,9 +175,11 @@ loop_specification
     | loop_separates_clause
     | loop_determines_clause
     | assignable_clause
+    | loop_assignable_clause
```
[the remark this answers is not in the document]
mi-ki: Thanks, fixed it.

mattulbrich:
It seemed to me that you renamed a few things from "assignable" to "modifiable", also in error messages. That led me to the conclusion that "modifiable" is the new default keyword.
Btw. in the old days in KeY (and currently in Danfy) the keyword was `modifies`.

mi-ki:
Thanks for spotting that, I reverted my changes in the two respective error messages now and added explanatory comment. Yep, in the link that I had provided, Dafny also compares to KeY's semantics as well as ACSL(++) and Ada/SPARK. From their description, it seems that Ada/SPARK’s dataflow contracts *actually* have a semantics of *assignable*.

key.core/src/main/java/de/uka/ilkd/key/speclang/njml/Translator.java
mi-ki: I rewrote the error message accordingly.

key.core/src/main/java/de/uka/ilkd/key/speclang/njml/JmlTermFactory.java
mi-ki: I changed the two error messages accordingly. As mentioned, this pull request does not infer or generate any specification. Hence, other than in the error messages, the user is not confronted with any new keywords, if they did not specify them themselves in the (user-given) program specification.

key.core.symbolic_execution/src/test/java/de/uka/ilkd/key/symbolic_execution/testcase/TestTruthValueEvaluationUtil.java
```diff
@@ -89,10 +89,10 @@ public void testTruthValueLabelBelowUpdatesDifferentToApplicationTerm() throws E
     }
 
     /**
-     * Tests example: /set/truthValueExceptinalAssignableNothingTest
+     * Tests example: /set/truthValueExceptionalModifiableNothingTest
```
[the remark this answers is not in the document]
mi-ki: Renamed it now, thanks!

WolframPfeifer (approved):
Thanks, I think having support for more framing keywords (also with the warning about possibly unintended semantics) is beneficial.

Also, thanks for removing ambiguities from the variable names, where `mod` stood for either `modifier` or `modality` or `modifiable` ;-)

PR #3368, opened 2023-12-01
Fix update of proof tree in case of filter changes (fixes #3367)

### LINKED ISSUE #3367: Proof tree filters not applied correctly
## Description
The filters in the proof tree ("Hide Intermediate Proofsteps" etc.) are not applied correctly, at least in the case of a freshly started proof.

## Reproducible
always

### Steps to reproduce 

1. Start any proof and apply some rules or macros to it.
2. Select "Hide Intermediate Proofsteps" in the proof tree settings (gear wheel).

I would expect that the intermediate proof steps are filtered out. However, the proof tree stays as it is (no filters applied). 
Furthermore, if the menu is opened again, the checkbox is not ticked.

### Additional information

* By bisecting, I found the first bad commit: 3ab8750d0e35f3a5d694d4c0e8bf196a7f1b2607.
* The problem seems to have something to do with the focus: At least in my testing, it only happened in freshly started proofs. If multiple proofs are present, after switching between them, the filters work fine.

unp1:
## Related Issue

This pull request addresses #3367

## Intended Change

Activating a filter works immediately. 

The underlying problem is that the GUIProofTreeModel stores the selection and does not update with the selection model in the ProofTreeView.

This PR does not maintain the consistency. The redundant information was seemingly only thought to be used to store the selected node when switching proofs and restoring it when switching back. The method setFilter should for that reason not query the model for the selection but the view to get an up-to-date path.

Previously it got null (despite a node being selected in the view) and  did not activate the filter.

## Type of pull request

- [ ] Other: 

## Ensuring quality
    
- [x] I have tested the feature as follows: The steps for reproduction of #3367 do no longer reproduce the error. Playing around with multiple proofs and selecting filters seems to work.

### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/settings/ViewSettings.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/settings/ViewSettings.java
@@ -78,22 +78,22 @@ public class ViewSettings extends AbstractPropertiesSettings {
     /**
      * Pretty Syntax is true by default, use Unicode symbols not
      */
-    private static final String PRETTY_SYNTAX = "PrettySyntax";
+    public static final String PRETTY_SYNTAX = "PrettySyntax";
 
     /**
      *
      */
-    private static final String USE_UNICODE = "UseUnicodeSymbols";
+    public static final String USE_UNICODE = "UseUnicodeSymbols";
 
     /**
      *
      */
-    private static final String SYNTAX_HIGHLIGHTING = "SyntaxHighlighting";
+    public static final String SYNTAX_HIGHLIGHTING = "SyntaxHighlighting";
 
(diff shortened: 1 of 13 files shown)


WolframPfeifer (commented):
I tested the GUI for the use cases that I could think of. Everything worked except the following minor bug: The selection highlights of expanded OSS nodes are not painted anymore (this worked in 2.12.0). The nodes are selected and the sequent is updated as expected (a feature implemented by @FliegendeWurst), just the selection highlight is not (it stays at the parent node).

WolframPfeifer (approved):
Thanks! I tried to come up with the common use cases and tested them. Everything seems to work now.

PR #3369, opened 2023-12-01
Save and restore proof tree view filters upon proof switching

### LINKED ISSUE #3367: Proof tree filters not applied correctly

unp1:
- Fix update of proof tree in case of filter changes (fixes #3367)
- Store ProofTreeViewSettings per proof outside model and restore all settings upon switching

## Related Issue

This pull request addresses #3367 

The PR is an addition to the immediate bug fix PR #3368. It adds a new feature.

## Intended Change

It removes some information from the GUIProofTreeModel that was kept redundant and only used for restoring the view state after proof switching.

The proof tree view states are now exported and saved explicitly thus allowing to also restore the filter settings on a per proof basis.

## Type of pull request

- [X] New feature (non-breaking change which adds functionality)
- [ ] Other: 

## Ensuring quality
    
- [X] I made sure that introduced/changed code is well documented (javadoc and inline comments).
- [X] I have tested the feature as follows: Opened different proofs with different filter settings and switching between them

### DIFF
--- a/key.ui/src/main/java/de/uka/ilkd/key/gui/actions/HidePackagePrefixToggleAction.java
+++ b/key.ui/src/main/java/de/uka/ilkd/key/gui/actions/HidePackagePrefixToggleAction.java
@@ -59,10 +59,8 @@ public final class HidePackagePrefixToggleAction extends MainWindowAction {
     }
 
     protected void handleViewSettingsChanged(PropertyChangeEvent e) {
-        if (ViewSettings.HIDE_PACKAGE_PREFIX.equals(e.getPropertyName())) {
-            updateSelectedState();
-            mainWindow.makePrettyView();
-        }
+        updateSelectedState();
+        mainWindow.makePrettyView();
     }
 
 }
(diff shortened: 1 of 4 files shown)


WolframPfeifer:
In our last KaKeY meeting, we decided to keep the current (non-local) filters, since (a) making them proof dependent would create a new category of settings (per-proof view settings) and (b) I prefer the filter settings to be global and basically no one else had a clear preference here.

@unp1 Can this PR be closed now?

PR #3370, opened 2023-12-04
Get rid of experimental flag using more fine-grained flags

wadoon:
## Motivation 

The `experimental` flag is a bad idea. You cannot control which feature you want to activate.
Instead, you are getting a full load of functionality one activated. This PR allows you to have more fine-grained feature flags, with a well-defined backward behavior for `--experimental`. 

This PR was created for #3022 to activate the labels w/o activated all of the other untested functions. 

## Intended Change

* A new settings `FeatureSettings` which is basically a set string (features to be activated)
* A set of feature flags in places where `Main.isExperimentalMode()` were requested.
* A settings panel 
  image
* A behavior for `--experimental`, activating all features.
* Feature flags are stored and read in the proof-independent settings
* Changes in feature flags can have an effect depending on the feature 

## Type of pull request

- [x] Other: The effect of CLI parameter `--experimental`. 

## Ensuring quality
    
### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/settings/AbstractPropertiesSettings.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/settings/AbstractPropertiesSettings.java
@@ -270,7 +270,7 @@ public abstract class AbstractPropertiesSettings extends AbstractSettings {
 
         @Override
         public T fromObject(@Nullable Object o) {
-            return null;
+            return fromObject.apply(o);
         }
     }
 }
--- a/key.core/src/main/java/de/uka/ilkd/key/settings/AbstractPropertiesSettings.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/settings/AbstractPropertiesSettings.java
@@ -111,7 +111,7 @@ public abstract class AbstractPropertiesSettings extends AbstractSettings {
         if (cat == null)
             return;
         propertyEntries.forEach(it -> {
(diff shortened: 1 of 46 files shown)


wadoon:
> * Activating the flags work (i.e. the checkbox is marked and the mark seems to be saved), but the extension seem not to be activated (before and after restart), e.g. activating the Test UI Bundle does not show the Test menu

Not all features are "hot" reloadable. I made the "Run All Proofs" menu item hot reload now. 

I also added a flag which says if a feature requires a restart. The settings dialogue gives an information message if the state of a non-hot-loadable feature has taken place. 
(Note this does not scale: If there is a different settings panel which also wants to give a message on settings application we would have two message dialogs. But for now, this solution is valid.)

wadoon:
> Exception in thread "AWT-EventQueue-0" java.lang.ClassCastException: class javax.swing.JLabel cannot be cast to class javax.swing.JPanel (javax.swing.JLabel and

This is an old error, I have fixed multiple times. 

The API says that a settings provider provides a JComponent, someone added a cast to JPanel, which fails for plugins enabled under `--experimental`. 

It is a good example why experimental flag or feature flags in general may not be good design choice.

wadoon:
> [13:25:38.759] ERROR SimpleSettingsPanel - Cannot invoke "javax.swing.JSpinner.setForeground(java.awt.Color)" because "this.maxField" is null

A new one! I did not understood why the validator was called for a JSpinner before the class was completely created. I added a guard.

PR #3374, opened 2023-12-16
JML enabled keys indicator for the Status Line 

wadoon:
Addition to #3373 which brings the following UI element: 

image

A label which represents the current value of the enabled keys in `GeneralSettings`.

### DIFF
--- a/key.ui/src/main/java/de/uka/ilkd/key/gui/JmlEnabledKeysIndicator.java
+++ b/key.ui/src/main/java/de/uka/ilkd/key/gui/JmlEnabledKeysIndicator.java
@@ -25,14 +25,21 @@ public class JmlEnabledKeysIndicator implements KeYGuiExtension, KeYGuiExtension
 
     public JmlEnabledKeysIndicator() {
         lblIndicator.setBorder(BorderFactory.createEmptyBorder(5, 5, 5, 5));
-        lblIndicator.setToolTipText("Currently enabled keys for the JML condition evaluation");
+        lblIndicator.setToolTipText("Currently enabled keys for tool-specific JML annotations");
         settings.addPropertyChangeListener(GeneralSettings.KEY_JML_ENABLED_KEYS, it -> update());
         update();
     }
 
     private void update() {
-        var lbl = String.join(", ", settings.getJmlEnabledKeys());
-        lblIndicator.setText(lbl);
+        final var current = settings.getJmlEnabledKeys();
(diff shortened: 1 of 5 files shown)


mattulbrich:
Nice idea, but it looks a bit ugly as it is in the screenshot, doesn't it? Should read "Currently" in the tooltip.

wadoon:
Added more margin: 

image

key.ui/src/main/java/de/uka/ilkd/key/gui/JmlEnabledKeysIndicator.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+@KeYGuiExtension.Info(experimental = false, name = "JML Enabled Keys Indicator for the Status Line")
+public class JmlEnabledKeysIndicator implements KeYGuiExtension, KeYGuiExtension.StatusLine {
+    private final GeneralSettings settings =
+        ProofIndependentSettings.DEFAULT_INSTANCE.getGeneralSettings();
+    private final JLabel lblIndicator = new JLabel();
+
+    public JmlEnabledKeysIndicator() {
+        lblIndicator.setBorder(BorderFactory.createEmptyBorder(5, 5, 5, 5));
+        lblIndicator.setToolTipText("Currently enabled keys for the JML condition evaluation");
```
mattulbrich, line 28: How about: "Currently enabled keys for tool-specific JML annotations"

or is "JML condition evaluation" a technical term?

key.ui/src/main/java/de/uka/ilkd/key/gui/JmlEnabledKeysIndicator.java
```diff
@@ -0,0 +1,42 @@
...
+        update();
+    }
+
+    private void update() {
+        var lbl = String.join(", ", settings.getJmlEnabledKeys());
+        lblIndicator.setText(lbl);
+    }
+
+    @Override
+    public List<JComponent> getStatusLineComponents() {
```
mattulbrich, line 39: I'd have another small request: Following the design principle of the "dark cockpit philosophy", one should only indicate those things which are out of the ordinary. In 99.9% of the cases, we will use the default setting here. So it should be displayed, only if the deviate from the standard.

That would probably be just one if statement here.

mattulbrich (changes requested):

mattulbrich (approved):

PR #3377, opened 2023-12-20
Fix for the LoopInvariantConfigurationDialog

unp1:
## Related Issue

This PR fixes a bug that prevented the interactive provision of a loop invariant.

## Intended Change

It should be possible to enter loop invariants manually.

## Type of pull request

- [ ] Other: 

## Ensuring quality
    
- [x] I have tested the feature as follows: Applied a manually entered loop invariant

The problem was if no "free modifies"/"free assignable" was available then creating the whole modifies set  "union(freeMod, mod)" led to a NullPointerException. The fix uses "strictly_nothing" as default for free assignables of loops, if non given.
     

### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/speclang/LoopSpecImpl.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/speclang/LoopSpecImpl.java
@@ -136,7 +136,7 @@ public final class LoopSpecImpl implements LoopSpecification {
     // -------------------------------------------------------------------------
 
     private Map /* Operator, Operator, Term -> Term */<Term, Term> getReplaceMap(Term selfTerm,
-            Map<LocationVariable, Term> atPres, Services services) {
+            Map<LocationVariable, Term> atPres) {
         final Map<Term, Term> result = new LinkedHashMap<>();
 
         // self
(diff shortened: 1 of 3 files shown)


WolframPfeifer:
I do not really understand the problem from the description. By "free modifier", do you mean "assignable_free"? How can this be used in interactively specified loops at all?

WolframPfeifer:
> Yes and yes :-) The problem is that the assignable_free cannot be specified and therefore "null" was returned as instantiation which then lead to an NPE.
> 
> Looking at the code, a specified (via JML) free assignable would cause the actual modifies set to be "union(mod, freeMod)", hence I assume, strictly_nothing should be the intended default (?), which is now used if that no free_assignable can be found.
> 
> A simple example is the attached file. The invariant specification to enter in the dialog is
> 
> invariant: i>=0 & i<=n modifies: {} variant: n - i
> 
> In the current version that leads to an NPE [simple.txt](https://github.com/KeYProject/key/files/13727256/simple.txt)
> 
> Changed additional information section in the PR description for more clarity.

Thanks for the explanation. I have to admit that I find the ..._free clauses quite confusing. I talked to @flo2702 today, who implemented the ..._free clauses, and he convinced me that indeed \strictly_nothing is the correct default here.

With the changes in this branch, something is very strange now: I can not even close an easy proof with a loop that writes only a single field. The problem seems to be that there are two different `self` variables (different hashes) ...

WolframPfeifer:
> > With the changes in this branch, something is very strange now: I can not even close an easy proof with a loop that writes only a single field. The problem seems to be that there are two different `self` variables (different hashes) ...
> 
> Just to be sure, you mean when using the dialog for the loop invariant (not as JML annotation)? If yes, I assume the bug is not introduced by this patch as it does not touch anything remotely related to "self", but I can have a look and try to fix it.
> 
> If no, that is really strange (and I will have a look in any case ;-)
Yes, when using the dialog. I think this error has been on the main branch for a while, but since the dialog did not work anyway, probably nobody noticed it.

WolframPfeifer (approved):
The invariant configurator in the GUI works again. Thanks!

PR #3379, opened 2023-12-28
Renovation of PO loading

wadoon:
## Intended Change

This is an overhaul of the `\proofObligation` inside of KeY files and some internal api changes. 

1. `\proofObligation` is migrated from a property string to the new configuration grammar. An entry looks as follows:
```
\proofObligation { 
    "class" : "de.uka.ilkd.key.proof.init.FunctionalOperationContractPO",
    "contract" : "SITA3[SITA3::rearrange()].JML normal_behavior operation contract.0",
    "name" : "SITA3[SITA3::rearrange()].JML normal_behavior operation contract.0"
 }
```
Earlier it was a string holding the information as a `Properties`. This allows in feature to have type-safe PO parameters. This affects the grammar, parser and also the store functionality of each PO class. 

2. Renovation of `#loadFrom`. Old KeY uses the `class` information to load this class and invokes the static method `#loadFrom(...)`. This PR replaces this behaviour using the `ServiceLoader`. This makes adding a new PO (loader) more convenient and type-safe. 

**A fallback for old `\proofObligation` is implemented.** Reading an old `\proofObligation "...";` entry should be translated into the new architecture. 

## Type of pull request

- **X** Breaking change (fix or feature that would cause existing functionality to change)
- **X** Other: KeY-Problem-Files, Grammar

## Ensuring quality
    
- **X** I made sure that introduced/changed code is well documented (javadoc and inline comments).
- **X** I made sure that new/changed end-user features are documented as well as before. 
- **X** I have tested the feature as follows: store a proof file


wadoon:
@unp1 Should be ready if green. To implement a fallback was easier than to add a useful error message. So it should read old files, too.

wadoon:
This idles here in a ready state and looking for a review for at least a month now. @mattulbrich @unp1 @WolframPfeifer

mattulbrich:
There are failing test cases. Is this ready for review?

wadoon:
> There are failing test cases. Is this ready for review?
It was green back then. But after auto-merging it became failing. 

But it is reviewable. Especially, the discussion on whether we want to let the old proof files be with the string-based properties format for proof obligations. In the history of the PR,  a compatibility layer was added. 

If this is decided and reviewed, I start fixing the test cases, but running behind failing unit-tests is easier than PR reviewers.

key.core/src/main/java/de/uka/ilkd/key/proof/init/WellDefinednessPOLoader.java
```diff
@@ -0,0 +1,46 @@
...
+import de.uka.ilkd.key.speclang.Contract;
+
+import org.jspecify.annotations.NullMarked;
+
+/**
+ * @author Alexander Weigl
+ * @version 1 (29.12.23)
+ */
+@NullMarked
+public class WellDefinednessPOLoader implements ProofObligationLoader {
```
mattulbrich, line 17: Javadoc.

key.core/src/main/java/de/uka/ilkd/key/proof/init/loader/DependencyContractPOLoader.java
```diff
@@ -0,0 +1,57 @@
...
+package de.uka.ilkd.key.proof.init.loader;
+
+import de.uka.ilkd.key.proof.init.*;
+import de.uka.ilkd.key.settings.Configuration;
+import de.uka.ilkd.key.speclang.Contract;
+
+import org.jspecify.annotations.NullMarked;
+
+@NullMarked
+public class DependencyContractPOLoader implements ProofObligationLoader {
```
mattulbrich, line 13: Javadoc.

key.core/src/main/java/de/uka/ilkd/key/proof/init/loader/FunctionOperationContractPOLoader.java
```diff
@@ -0,0 +1,80 @@
...
+import de.uka.ilkd.key.speclang.FunctionalOperationContract;
+
+import org.jspecify.annotations.NullMarked;
+
+/**
+ * @author Alexander Weigl
+ * @version 1 (28.12.23)
+ */
+@NullMarked
+public class FunctionOperationContractPOLoader implements ProofObligationLoader {
```
mattulbrich, line 20: Javadoc.

mattulbrich (commented):
I do like the idea. Astonishingly large change apparently.

mattulbrich:
> Especially, the discussion on whether we want to let the old proof files be with the string-based properties format for proof obligations. In the history of the PR, a compatibility layer was added.
> If this is decided and reviewed, I start fixing the test cases, but running behind failing unit-tests is easier than PR reviewers.

I'd would suggest to migrate the old proof file in the repo to the new format (it looks like this has been done ...)

Having the ability to parse Strings does not hurt either for a while. Going from a properties object to a key-value object should also be relatively local.

PR #3380, opened 2023-12-28
Alternative for: "If possible save the last selected node in the proof, to reselect it upon loading."

wadoon:
Alternative version for #3324 using the settings instead of storing it in the proof. 

Using a path encoding of the current node (list of child indexes) and a run-length encoding to keep it small. 

**Missing:** Finding the right spot where the `KeYUserProblemFile` is available and the proof is loaded. The current place is unsuitable.


wadoon:
@unp1 I now call `getMediator().nonGoalChosen(node)` with a valid node, but it is not selected. Where is the other place with the logic that the last node should be taken?

wadoon:
@unp1 @mattulbrich I updated this to the latest `main`. The current obstacle is the lack of communication bandwidth, which prevents the provision of additional information to the UI world. 

I can vaguely remember that we had a conclusion in Feb'2024 that the additional (KeY-/tool-maintained) information should be separated from `\settings {}`.  Correct?

PR #3385, opened 2024-01-13
Introducing a new function symbol seqUpd

mattulbrich:
## Intended Change

There was no explicit update function for sequences in KeY (any more). I added `seqUpd` for that purpose.
The axiomatic rules is in `seqRules.key` and essentially reads:

     \find(alpha::seqGet(seqUpd(seq, idx, value), jdx))
     \replacewith(\if(0<=jdx & jdx < seqLen(seq) & idx=jdx) \then((alpha)value) \else(alpha::seqGet(seq, jdx)))

* The JML translation was adapted
* The JML in Java translation was extended and adapted

## Type of pull request

- [x] There are changes to the taclet rule base
- [ ] Other: 

## Ensuring quality
    
- [ ] I made sure that introduced/changed code is well documented (javadoc and inline comments). *The added code essentially copies similar code around it and has the same level of documentation.*
- [ ] I made sure that new/changed end-user features are well documented (https://github.com/KeYProject/key-docs). *There is no documentation for \seq to add to.*
- [ ] I added new test case(s) for new functionality. *There are no test cases for the other similar seqXXX functions to which one would add new test cases.*
- [x] I have tested the feature as follows: *by extending an existing case study in KeY*


key.core/src/main/resources/de/uka/ilkd/key/proof/rules/seqRules.key
```diff
@@ -241,6 +253,17 @@
...
     };
 
+    \lemma
+    getOfSeqUpd {
+        \schemaVar \term Seq seq;
+        \schemaVar \term int idx, jdx;
+        \schemaVar \term any value;
+
+        \find(alpha::seqGet(seqUpd(seq, idx, value), jdx))
+        \replacewith(\if(0<=jdx & jdx < seqLen(seq) & idx=jdx) \then((alpha)value) \else(alpha::seqGet(seq, jdx)))
```
WolframPfeifer, line 263:
```suggestion
        \replacewith(\if(jdx>=0 & jdx <= -1 + seqLen(seq) & jdx=idx) \then((alpha)value) \else(alpha::seqGet(seq, jdx)))
```
I think we could save some rule applications for normalization (~10 per application of the taclet) here if we write down the (in-)equations already in normalized form. On the contrary, the taclet gets less readable. Not sure what we want to have here ...

WolframPfeifer (changes requested):
Nice addition!

I see one minor problem for the code quality: The relevant classes/tokens/methods are sometimes called `seqPut`, sometimes `seqReplace`, sometimes `seqUpd`. I think that this is quite confusing, in particular for new developers. Can we just name everything consistently `seqUpd`?

key.core/src/main/resources/de/uka/ilkd/key/proof/rules/seqRules.key
mattulbrich: I would favour to keep it in the readable form. KeY is an interactive TP. Normalising as early as possible does not make things more comprehensible at all. We can afford these ten rule apps.

PR #3386, opened 2024-01-14
Combine equalsMod... methods

tobias-rnh:
## Intended Change

There were four different functions concerned with equality of terms other than the usual `equals`, namely `equalsModRenaming`, `equalsModTermLabels`, `equalsModIrrelevantTermLabels` and `equalsModProofIrrelevancy`. Now, there is a new Interface `EqualsModProperty`, which is extended by `Term`, with a single method `equalsModProperty` that takes two parameters: the object to be compared and one of the new type `TermProperty`, that implements the desired behavior of what to be ignored during the equality check.
The changes can be found in a new package called **equality** in the logic package of key.core. The `TermProperty`s are designed as a singleton and can be accessed through their respective static constants, e.g. `IRRELEVANT_TERM_LABELS_PROPERTY` in `IrrelevantTermLabelsProperty`.
`Term` now no longer implements the Interface `EqualsModProofIrrelevancy`. Therefore it might be a problem if you wanted to use terms in an `EquivalenceDirectedGraph`. 

## Type of pull request

- [x] Refactoring (behaviour should not change or only minimally change) _Behavior of equalsMod... methods combined into a singular method called equalsModProperty_
- [x] Breaking change (fix or feature that would cause existing functionality to change) _Term no longer implements EqualsModProofIrrelevancy. One can therefore not, for instance, use Terms in the EquivalenceDirectedGraph_
- [ ] Other: 

## Ensuring quality
    
- [x] I made sure that introduced/changed code is well documented (javadoc and inline comments). _Old methods concerning equalsModRenaming logic have been moved to RenamingProperty.java. Some of them were not documented before and still are not. Other new code has been documented._
- [ ] I made sure that new/changed end-user features are well documented (https://github.com/KeYProject/key-docs). _This should not affect the end-user_
- [x] I added new test case(s) for new functionality. _As some of the equalsMod... methods were not tested before, I added new tests in the new test class TestEqualsModProperty_
- [x] I have tested the feature as follows: _Above mentioned test cases_

     


tobias-rnh:
While testing the behaviour of the old equalsModProofIrrelevancy for terms to move it to a TermProperty, I discovered the following behaviour: 
If you take an unlabeled term and compare it to the same term with a proof-irrelevant label attached, equalsModProofIrrelevancy returns true, no matter on what term it is called. 
If however you take two labeled terms, that only have proof-irrelevant labels attached, but differ in the amount of those, equalsModProofIrrelevancy returns false, no matter on what term it is called. 

I could imagine that this behaviour is not intended. But just to make sure that I do not break the intended behaviour, I wanted to ask @FliegendeWurst and @WolframPfeifer, how equalsModProofIrrelevancy should act in these situations.

If this behaviour is indeed not intended, I would simply change it in my implementation. Otherwise, I will of course keep it this way.
Thank you for your time!

FliegendeWurst:
> If however you take two labeled terms, that only have proof-irrelevant labels attached, but differ in the amount of those, equalsModProofIrrelevancy returns false, no matter on what term it is called.

That is indeed a bug. It should return true.

tobias-rnh:
I have again checked that all calls of equalsModProperty use the right property to match the calls to prior equalsMod... methods.
Only in TestApplyUpdateOnRigidCondition, equalsModProofIrrelevancy was replaced by equalsModProperty with RENAMING_PROPERTY.
This is okay and wanted as it actually should have been equalsModRenaming in the first place. 

Apart from that, everything is as it should be and tests were still passing locally.

key.core/src/main/java/de/uka/ilkd/key/logic/Term.java
```diff
@@ -43,8 +43,11 @@
  * supported: {@link Term#execPostOrder(Visitor)} and {@link Term#execPreOrder(Visitor)}.
  */
 public interface Term
-        extends SVSubstitute, Sorted, EqualsModProofIrrelevancy, org.key_project.logic.Term {
-    @Override
+        extends SVSubstitute, Sorted, TermEqualsModProperty, org.key_project.logic.Term {
+
+    /**
+     * The top operator (e.g., in "A and B" this is "and", in f(x,y) it is "f").
+     */
```
Drodt, line 50: Why is this no longer an override?

tobias-rnh: That is indeed strange. I took another look at the first merge I did in IntelliJ and it added the correct import and overrides at the other positions but not there. I will add the override and remove the comment.

Drodt: Somehow that broke the Unit tests. Does this also happen on your machine?

tobias-rnh: Tests are still running on my machine, I will reply again when they are finished.

tobias-rnh: I ran testFast and the only tests failing as per usual were the ones concerned with symbolic execution.
grafik

tobias-rnh: Running test also only fails for symbolic execution.
grafik

tobias-rnh: I was just checking this this morning, but it strangely did not fail locally.
grafik

(And just to be absolutely sure I synced my fork and tried it again... Still passes.)

Drodt: That makes some sense. Then there must be a bug outside of the KeY code. Either CI or even a Java bug. Especially since the Windows tests pass.

tobias-rnh: The unit tests are now passing but testRunAllFunProofs fails. This also happens locally and even on the main branch.

tobias-rnh: Running the test locally results in the following problem:
grafik
Followed by:
grafik

DerDrodt: The tests work on my main branch

Drodt: The tests work on my local main branch

tobias-rnh: After syncing the fork, we are now back at only the unit tests failing. Might this be solved clearing the cache again?

Drodt: Probably, yes. I just cleared some caches and I am gonna restart it in a bit. Thanks for checking!

Drodt (approved):

Issue #3378, opened 2023-12-20
Drag and drop does no longer indicate target formula


mattulbrich:
## Description

It used to be that the formula which becomes the target of a drag-and-drop activity was highlighted in green.

This is no longer the case since commit 4759370f719ed506746e7a2a13832eac48cfc6c1 back in June.

## Reproducible
  
always

### Steps to reproduce 

Drag and drop a value onto a universal quantifier in the antecedant (e.g.) There no highlight on the target formula

### Additional information

This is broken since the mentioned commit. Can we just revert that very commit?

---

* Commit: 4911758ba07017f4af60dc0314ac665c14a542d0

mattulbrich:
The original problem has been fixed on #3391. On #3506 there is a followup correction of the heatmap highlighting.

PR #3391, opened 2024-01-30
Revert "Fix CurrentGoalView highlights not being removed"

mattulbrich:
## Related Issue

This pull request addresses #3378.

## Intended Change

This reverts commit 4759370f719ed506746e7a2a13832eac48cfc6c1.

As discussed during a developers meeting, the highlighting is quite important during drag'n'drop.

## Type of pull request

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] There are changes to the (Java) code
- [ ] Other: 

## Ensuring quality
    
- [X] I have tested the feature as follows: *Tried it in the gui*

### DIFF
--- a/key.ui/src/main/java/de/uka/ilkd/key/gui/nodeviews/CurrentGoalView.java
+++ b/key.ui/src/main/java/de/uka/ilkd/key/gui/nodeviews/CurrentGoalView.java
@@ -203,7 +203,6 @@ public final class CurrentGoalView extends SequentView implements Autoscroll {
             return;
         }
         var time = System.nanoTime();
-        getHighlighter().removeAllHighlights();
 
         removeMouseListener(listener);
 


WolframPfeifer (changes requested):
The drag'n'drop highlights work now, but unfortunately it breaks other things: Toggling the Heat-Maps does not work anymore, it is necessary to apply/remove a rule to trigger the repaint.

WolframPfeifer:
@unp1 Might be connected to your PR.

mattulbrich:
> The drag'n'drop highlights work now, but unfortunately it breaks other things: Toggling the heat maps does not work anymore, it is necessary to apply/remove a rule to trigger the repaint.

Toggling heat maps does not work on main either. That is a different problem. This is not the problem of this change ... But I will look into this anyway.

mattulbrich:
Moreover, on main, if the formula is a single line, then mouse highlighting does not work, too. Again, nothing to do with this change.

wadoon (approved):

WolframPfeifer (approved):
Thank you so much for finally making this work again! Also, I think that it was a good choice to disable the caret completely in the SequentView, since the weird selection behavior led to some confusion in the past (in particular with new users of KeY).

I only found a single minor issue (probably, it was already there on main): On closed nodes of the proof, toggling the heatmap does not take immediate effect. You have to switch nodes to see it.

However, I think this should not block us from merging, so I approve this and leave it up to you whether you want to fix that. I looked into it a bit, but at the moment it seems that in MainWindow there is a confusion between SequentView and the subclasses CurrentGoalView/EmptySequent/InnerNodeView: Some actions (such as updating heatmaps) have only an effect CurrentGoalView, not the others. However, I was not able to fix it quickly, I think it needs quite some work to clean up, so I postponed it for now.

WolframPfeifer:
@mattulbrich Do you have a good idea how to fix the problem with the non-goal views? If so, could you update it, please? If not, feel free to merge. The remaining UI bug is minor.

mattulbrich:
I have worked on this. Is this fixed now, @WolframPfeifer ?

@wadoon How do I remerge this PR now?

PR #3399, opened 2024-02-11
Removal of Triple, and Quadruple

wadoon:
## Refactoring 

This PR starts the removal of generic data classes in favor of Java records (aka. named tuples). In this PR, only `Triple` and `Quadruple` are removed. 

In general, generic parametric data classes (`Tuple`, `Union`, ...) lead to unreadable, incomprehensible code, e.g., what does the entity `Triple<StatementBlock, URI, Integer>` express? Answer: It is the key in a map to find block contracts. Using the record `BlockContractKey` expresses this better than triple of something. 

Therefore, records give you the chance of defining useful variable names, and documentation,  and also avoid auto-boxing (hence NPE). 

## Intended Change

No use of `Triple<…>` and `Quadruple<…>`.

## Type of pull request

## Ensuring quality
    
- **X** by TEST CASES!!! and compiler


jml_classlevel.txt
FliegendeWurst, line 1: I don't think this file is supposed to be here.

key.core.symbolic_execution/src/main/java/de/uka/ilkd/key/symbolic_execution/SymbolicExecutionTreeBuilder.java
```diff
@@ -1500,10 +1499,10 @@ protected AbstractExecutionMethodReturn<?> createMehtodReturn(AbstractExecutionN
                             // Find the call Node representation in SED, if not available ignore it.
                             IExecutionNode<?> callSEDNode = keyNodeMapping.get(callNode);
                             if (callSEDNode instanceof ExecutionMethodCall) { // Could be the start
-                                                                              // node if the initial
-                                                                              // sequent already
-                                                                              // contains some
-                                                                              // method frames.
+                                // node if the initial
```
FliegendeWurst, line 1502: Why are there so many superfluous whitespace changes in this PR?

wadoon:
Moved to #3529 (from external branch to internal branch), due to write permission issues on sonarqube workflow.

PR #3400, opened 2024-02-11
Immutable and valid JML set statement

wadoon:
Version of #3195 but with an immutable Java AST. 

This PR implements JML set-statement in a non-Java-like fashion (no reduction to copy assignment) preserving the semantics of Java. 

Translation of JML goes directly from JML to KeY terms lazily inside the built-in rule `SetStatementRule`. 

## TODO 

* [x] Cleanup
* [x] Program variable replacement has still an issue. There are duplicated program variable objects that represent the same variable in the program and have the same name.

## Type of pull request

- **X**  Bug fix (non-breaking change which fixes an issue)
- **X** New feature (non-breaking change which adds functionality)
- **X** Breaking change (fix or feature that would cause existing functionality to change)
- **X** There are changes to the **built-in** rule base

## Ensuring quality
    
- I made sure that introduced/changed code is well documented (javadoc and inline comments)


mattulbrich:
This version verifies the following program:
```
class A {

    //@ ensures true;
    void m() {
        
        //@ ghost \bigint x = 0;

        int i = 0;
        while(i < 2) {
            //@ ghost \bigint y = 1;
            //@ set x = x + y;
            i++;
        }
        //@ assert x == 0;
        // but should be x == 2;
    }
}
```
which is not correct at all. The problem is that in every loop iteration y becomes a different symbol. However, in `x+y` the y remains always the original variable and is not updated to the respective instances. That is the reason why in assumptions and assertions the copy is kept and actually modified (breaking immutability).

mattulbrich:
Even simpler: The following can be proved:
```
    void m() {
        //@ ghost \bigint x = 0;
        //@ set x = x + 1;
        //@ assert x == 0;
    }
```
The x in the set statement does not refer to the correct variable (neither when reading, nor when writing)

wadoon:
Then I would say, the simple does not work and we need to go back to the drawing board. 

@unp1 Do you have insides into the mechanism of variable renaming? Is there a chance, to get around the `ProgVarReplaceVisitor` completely? Why do we not book this information along a goal and/or a modality?

wadoon:
Everything is terrible. Current test program: 

```java 
class Test {
    //@ ghost int x;

    int a;

    //@ requires true; ensures x==a;
    void foo() {
        //@set x = 0;
        a = 0;
        //@set x = x + 1;
        a = a + 1;
        //@set x = x + 1;
        a = a + 1;
        //@set x = x + 1;
        a = a + 1;
        //@set x = x + 1;
        a = a + 1;
        //@set x = x + 1;
        a = a + 1;
    }


    //@ ghost int rec;
    int cer;

    //@ requires a >= 0; ensures rec == cer; measured_by a;
    int voo(int a) {
        if (a == 0) {
            //@ set rec = 0;
            cer = 0;
            return 0;
        } else {
            int r = voo(a - 1) + 1;
            //@ set rec = r;
            cer = r;
            //@ assert r >=0;
            return r;
        }
    }

}
```

There are only two possibilities: 

- Using the spec repository
- Implement the JML classes in the KeY-Java-Ast for renaming.

wadoon:
@mattulbrich The mutable information of `SetStatements` moved to `SpecificationRepository`. 
Hence, the AST is immutable again. If the change suits you, we need to adopt the `JmlAssert` also. 

The example above is proofable except for the faulty assert statement. 


There are currently two construction sites:

1. `PrettyPrinter` can not show the current expression, as it has no access to services or the spec repo.
2. It is unclear whether a copy of the statement is necessary in `ProgVarReplacer` or `IntroAtPreDefOps`.

key.core/src/main/java/de/uka/ilkd/key/java/statement/SetStatement.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    /**
+     * Removes the attached parser context from this set statement
+     *
+     * @deprecated weigl: The use of {@link org.antlr.v4.runtime.ParserRuleContext} directly is
+     *             discouraged. It adds
+     *             an (often) unnecessary dependency to the ANTLR across the project.
+     * @return the parser context that was attached
+     */
```
mattulbrich, line 54: What should one use instead? Perhaps mention that it is to be removed with new java+jml framework.

key.core/src/main/java/de/uka/ilkd/key/ldt/HeapLDT.java
```diff
@@ -150,6 +152,51 @@ private String getFieldSymbolName(LocationVariable fieldPV) {
     // public interface
     // -------------------------------------------------------------------------
 
+    /**
+     * Wrapper class
+     */
+    public static final class SplitFieldName {
```
mattulbrich, line 158: below a record was used. here too?

key.core/src/main/java/de/uka/ilkd/key/proof/mgt/SpecificationRepository.java
```diff
@@ -1847,4 +1850,87 @@ public ImmutableSet<WellDefinednessCheck> getAllWdChecks() {
...
+        return statementMap.put(statement, spec);
+    }
+
+    /**
+     * This record represents mutable information for JML statements.
+     * JML statements need to maintain the current variable set as well as the updated information for the KeY terms
+     * they describe.
+     * <p>
+     * For {@link de.uka.ilkd.key.java.statement.JmlAssert} this is the formula behind the assert.
+     * For {@link de.uka.ilkd.key.java.statement.SetStatement} this is the target and the value terms.
```
mattulbrich, line 1873: that's a bit unclear. why is it mutable? in a record?
What is "This is"? Would that not always be one formula/term at most? Why collections?
That would make nasty bugs otherwise. There was something with loops specs once ...

key.core/src/main/java/de/uka/ilkd/key/rule/SetStatementRule.java
```diff
@@ -0,0 +1,135 @@
...
+        assert formula.op() instanceof UpdateApplication
+                : "Currently, this can only be applied if there is an update application in front of the modality";
+
+        Term update = UpdateApplication.getUpdate(formula);
+        Term target = UpdateApplication.getTarget(formula);
+
+        SetStatement setStatement =
+            Optional.ofNullable(JavaTools.getActiveStatement(target.javaBlock()))
+                    .filter(SetStatement.class::isInstance).map(SetStatement.class::cast)
+                    .orElseThrow(() -> new RuleAbortException("not a JML set statement."));
```
mattulbrich, line 93: just a comment: I don't know, but this functional style does not look so convining in Java ...

key.core/src/main/java/de/uka/ilkd/key/proof/mgt/SpecificationRepository.java
```diff
@@ -1847,4 +1850,87 @@ public ImmutableSet<WellDefinednessCheck> getAllWdChecks() {
...
+    private final Map<Statement, JmlStatementSpec> statementMap = new IdentityHashMap<>();
+
+    @Nullable
+    public JmlStatementSpec getStatementSpec(Statement statement) {
+        return statementMap.get(statement);
+    }
+
+    public JmlStatementSpec addStatementSpec(Statement statement, JmlStatementSpec spec) {
+        return statementMap.put(statement, spec);
+    }
```
mattulbrich, line 1865: Is it always ensured that there are different assertions/set statements for different branches? When unrolling, inling etc.?

mattulbrich (approved):
As discussed: It is a pity we have to do it like this, but better than mutable statements I assume. Thanks.

key.core/src/main/java/de/uka/ilkd/key/rule/SetStatementRule.java
wadoon: I know. A reason I use Kotlin. This code is from Julian. 

I did not add any functionality, I just moved it around.

key.core/src/main/java/de/uka/ilkd/key/proof/mgt/SpecificationRepository.java
wadoon: You/we need to ensure, that a new `SetStatement` is created. 

Something that needs to be assured by `ProgVarReplacer`, and was not done before.

wadoon:
I would also now transform JMLAssert to `SpecificationRepository`.

key.core/src/main/java/de/uka/ilkd/key/java/statement/SetStatement.java
wadoon: I encapsulate the ParserRuleContexts normally in the [`KeyAst`](https://github.com/KeYProject/key/blob/e0995588635b2e5e38c1791e254de76ab4391d36/key.core/src/main/java/de/uka/ilkd/key/nparser/KeyAst.java) classes, which holds the PRC as an attribute but do not expose this. Instead, it offers some basic functionality. 

Of course, you can retrieve a PRC in these classes, but you have to use `ParserFacade#getParseRuleContext` which is the only leakage. 

(Sometimes, in rare situations you need the PRCs, but mostly only in the core framework.)

key.core/src/main/java/de/uka/ilkd/key/proof/mgt/SpecificationRepository.java
wadoon: I wrote a little bit more.

wadoon:
Added a new highlighting for JML statement:

image

The old highlighting was green, and was inherited from comments.

wadoon:
State not so bad:

Unit-Tests:
 
image

RAP:

image


image

mattulbrich:
> Added a new highlighting for JML statement:
> 

Would it perhaps make sense to deliberately deviate from JML like syntax here to avoid confusion on the user's side?

like `//@ set a := 42` for instance or `//@ set heap := heap[o.ghst := 42]` which is more faithful to the JavaDL side.

PR #3406, opened 2024-02-13
Try to achieve a better parser error message

wadoon:
This PR removes the LL(*) fallback from the ParsingFacade to achieve better error message.

### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/nparser/ParsingFacade.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/nparser/ParsingFacade.java
@@ -21,7 +21,6 @@ import de.uka.ilkd.key.util.parsing.BuildingException;
 
 import org.antlr.v4.runtime.*;
 import org.antlr.v4.runtime.atn.PredictionMode;
-import org.antlr.v4.runtime.misc.ParseCancellationException;
 import org.antlr.v4.runtime.tree.TerminalNode;
 import org.jspecify.annotations.NonNull;
 import org.jspecify.annotations.Nullable;
@@ -133,15 +132,7 @@ public final class ParsingFacade {
         // we don't want error messages or recovery during first try
         p.removeErrorListeners();
         p.setErrorHandler(new BailErrorStrategy());
-        KeYParser.FileContext ctx;
-        try {
-            ctx = p.file();
-        } catch (ParseCancellationException ex) {
-            LOGGER.warn("SLL was not enough");
-            p = createParser(stream);
-            ctx = p.file();
-        }
-
+        KeYParser.FileContext ctx = p.file();
         p.getErrorReporter().throwException();
         return new KeyAst.File(ctx);
     }


key.core/src/main/java/de/uka/ilkd/key/nparser/ParsingFacade.java
```diff
@@ -128,21 +129,22 @@ public static KeyAst.File parseFile(File file) throws IOException {
 
     public static KeyAst.File parseFile(CharStream stream) {
         KeYParser p = createParser(stream);
-
+        var ep = p.getErrorReporter();
         p.getInterpreter().setPredictionMode(PredictionMode.SLL);
         // we don't want error messages or recovery during first try
-        p.removeErrorListeners();
-        p.setErrorHandler(new BailErrorStrategy());
```
mattulbrich, line 135: Usually this BailOutStrategy is used to avoid that ANTLR tries to repair. Why is it better to remove it?

key.core/src/main/java/de/uka/ilkd/key/nparser/ParsingFacade.java
```diff
@@ -128,21 +129,22 @@ public static KeyAst.File parseFile(File file) throws IOException {
...
-        KeYParser.FileContext ctx;
+        KeYParser.FileContext ctx = null;
         try {
             ctx = p.file();
-        } catch (ParseCancellationException ex) {
+        } catch (ParseCancellationException ex1) {
```
mattulbrich, line 138: Is it guaranteed to always be a ParseCancellationException?

mattulbrich (changes requested):
In your example the totally unclear SLL error message is still shown. Why? If the other is the "real" error message, can we restrict ourselves to it?

key.core/src/main/java/de/uka/ilkd/key/nparser/ParsingFacade.java
wadoon: I have integrated the bailout strategy into the `SyntaxErrorReporter`. Now every keyparser does not try to recover. This should solve skipped definitions.

key.core/src/main/java/de/uka/ilkd/key/nparser/ParsingFacade.java
wadoon: Yes. The `ParseCancellationException` is thrown from the `SyntaxErrorReporter` on recovering from parse errors. `ParseException`s are caught during the Parser, and given to the error listener and error strategy. 

It is a little bit complicated.

wadoon:
> In your example the totally unclear SLL error message is still shown. Why? If the other is the "real" error message, can we restrict ourselves to it?

It is not final as I did not understand much of ANTLR internals. 

Error messages are now caught up in the `ErrorStrategy`, so e.g., I lose have the ANTLR error messages not available anymore.

mattulbrich:
image
[Rosenkrantz,Stearns: Properties of deterministic top-down grammars](https://www.sciencedirect.com/science/article/pii/S0019995870904468?via%3Dihub)

wadoon:
SLL is sufficient for the KeY taclet base. 

Should we just drop LL mode?

wadoon:
SLL is sufficient for the KeY taclet base. 

Should we just drop LL mode?

wadoon:
**KaKeY decision:** Fallback to LL(*) will be removed.

wadoon:
```

    @Override
    public Token recoverInline(Parser recognizer) throws RecognitionException {
        InputMismatchException e = new InputMismatchException(recognizer);
        for (ParserRuleContext context = recognizer.getContext(); context != null; context = context.getParent()) {
            context.exception = e;
        }
        var tok = e.getOffendingToken();

        var message = "I got offended by the token '%s'. Expected tokens are: %s".formatted(tok,
                e.getExpectedTokens().toString(recognizer.getVocabulary()));

        syntaxError(recognizer, e.getOffendingToken(),
                tok.getLine(), tok.getCharPositionInLine(), message, e);
        throw new ParseCancellationException(e);
    }

    @Override
    public void recover(Parser recognizer, RecognitionException e) throws RecognitionException {
        for (ParserRuleContext context = recognizer.getContext(); context != null; context = context.getParent()) {
            context.exception = e;
        }
        var tok = e.getOffendingToken();
        final var parseCancellationException = new ParseCancellationException(e);

        syntaxError(recognizer, e.getOffendingToken(),
                tok.getLine(), tok.getCharPositionInLine(), parseCancellationException.getMessage(), e);

        throw parseCancellationException;
    }
```

wadoon:
closed in favor of hekeython2 approaches.

PR #3408, opened 2024-02-14
Cleanup of declaration of taclet options

WolframPfeifer:
This PR does some cleanup in the .key LDT files:
* The declaration of taclet options is moved from `ruleSetsDeclarations.key` into a separate file `optionsDeclarations.key`.
* sort and function declarations are moved from `ruleSetsDeclarations.key` and `javaRules.key` into the new file `javaHeader.key`.
* The loading order is changed slightly: First load the taclet options, then the rule sets, and then all the theories.
* Provide a better error message when a sort declaration is ignored due to an earlier declaration.
* A old dangling comment of a legacy taclet option which does not exist anymore (`optimisedSelectRules`) was removed.

## Type of pull request

- [x] There are changes to the taclet rule base (functionality should not change though)
- [ ] Other: 

## Ensuring quality
    

key.core/src/main/resources/de/uka/ilkd/key/proof/rules/ruleSetsDeclarations.key
```diff
@@ -288,7 +13,6 @@ The optimized rules pull out select terms before they evaluate them. Because oth
 // 1) NodeInfo#symbolicExecNames
 // 2)
```
FliegendeWurst, line 14:
```suggestion
// 2) OneStepSimplifier#ruleSets
```

WolframPfeifer: You are right, that is another place where hard-coded rule sets are used. I also added macro declarations to the list.

WolframPfeifer:
@mattulbrich This is the PR I mentioned yesterday. It just refactors the declarations in some .key files a bit and should not introduce any changes (apart from a more helpful message when a symbol is declared twice). It would be nice if you could take a quick look ... Thanks in advance!

PR #3424, opened 2024-02-21
added checkbox to disable example loader directly in dialog

JonasKlamroth:
## Intended Change
Added a checkbox to the example loader dialog that allows to hide it on subsequent startups. I consider this a reasonable change as this functionality is already supported by KeY but hidden in the preferences which most user might never open. Additionally i think this is standard behavior for similar IDEs or programs. No default behavior is changed.

## Type of pull request

- [ ] Other: 

## Ensuring quality
    
- [x] I made sure that new/changed end-user features are well documented (https://github.com/KeYProject/key-docs).
- [x] I have tested the feature as follows: Multiple startups with all possible configurations. 
- [x] I have checked that runtime performance has not deteriorated.

### DIFF
--- a/key.ui/src/main/java/de/uka/ilkd/key/gui/ExampleChooser.java
+++ b/key.ui/src/main/java/de/uka/ilkd/key/gui/ExampleChooser.java
@@ -21,6 +21,8 @@ import javax.swing.tree.TreePath;
 import javax.swing.tree.TreeSelectionModel;
 
 import de.uka.ilkd.key.gui.utilities.GuiUtilities;
+import de.uka.ilkd.key.settings.ProofIndependentSettings;
+import de.uka.ilkd.key.settings.ViewSettings;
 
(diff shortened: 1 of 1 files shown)


key.ui/src/main/java/de/uka/ilkd/key/gui/ExampleChooser.java
```diff
@@ -312,6 +322,7 @@ public void mouseClicked(MouseEvent e) {
         buttonPanel.add(cancelButton);
         GuiUtilities.attachClickOnEscListener(cancelButton);
 
+
```
FliegendeWurst, line 325:
```suggestion
```

FliegendeWurst (approved):
GUI change looks good. Code is reasonable.

wadoon:
force pushed for reformat.

WolframPfeifer (approved):
Thanks! That is a nice usability feature!

PR #3427, opened 2024-02-22
Hackeython: ADTs in JML

Drodt:
## Intended Change

Based on #3420. Changes the JML pareser to accept definitions of ADTs. Datatypes have a list of _constructors_ separated by `|`, and a body of functions.

```java
class Test {
	/*@ datatype List {
	  @ 	Nil()
	  @	| Cons(\any head, List tail);
	  @
	  @ 	int size() {
	  @		return 0;
	  @ 	}
	  @ }
	  @*/
}
```

Currently, ADTs can only be defined on class-level, not inside methods or outside of classes.

There also is no semantics for these ADTs. KeY simply ignores them. Ideally, we want to use the defined types and functions in specification and translate them to KeY ADTs.

We also want to use the new Java switch expressions to allow pattern matching on ADTs, i.e,:
```java
switch (list) {
  case Nil -> 0;
  case Cons(_, xs) -> 1 + xs.size();
}
```

Open questions are:
- How are JML ADTs translated to KeY?
- How are switch expressions with pattern matching translated?
- Can we define a shorthand infix notation for ADT function, e.g., `l1 + l2` for `l1.append(l2)`?
- How do we resolve ADT function calls in specification?
- What happens if we already have a `List` class/interface?

OpenJML has a similar concept of ADTs and pattern matching. For demonstration, [there is a list example](https://github.com/OpenJML/OpenJML/blob/master-21/OpenJMLTest/test/datatype/Test.java).

## Type of pull request

- [ ] There are changes to the (Java) code
- [x] Other: Changes to ANLTR grammars

## Ensuring quality
    
### DIFF
--- a/key.core/src/main/antlr4/JmlLexer.g4
+++ b/key.core/src/main/antlr4/JmlLexer.g4
@@ -96,6 +96,7 @@ CAPTURES: 'captures' Pred -> pushMode(expr);
 CODE: 'code'; //?
 CONSTRAINT: 'constraint' Pred -> pushMode(expr);
 CONTINUES: 'continues' -> pushMode(expr);
+DATATYPE: 'datatype';
 DEBUG: 'debug'; //?
 DECREASING: ('decreasing' | 'decreases' | 'loop_variant') Pred -> pushMode(expr);
 DETERMINES: 'determines' -> pushMode(expr);
@@ -155,6 +156,7 @@ C_IDENT: '\\'? LETTER (LETTERORDIGIT)* -> type(IDENT);
 C_COLON: ':' -> type(COLON);
 C_DOT: '.' -> type(DOT);
 C_COMMA: ',' -> type(COMMA);
+C_INCLUSIVEOR: '|' -> type(INCLUSIVEOR);
 
 SL_COMMENT: {jmlMarkerDecision.isComment("//")}? ('//' ('\n'|'\r'|EOF) | '//' ~'@' ~('\n'|'\r')*) -> channel(HIDDEN);
 ML_COMMENT: {jmlMarkerDecision.isComment("/*")}? '/*' -> more, pushMode(mlComment);
--- a/key.core/src/main/antlr4/JmlParser.g4
+++ b/key.core/src/main/antlr4/JmlParser.g4
@@ -22,6 +22,7 @@ classlevel_element
   | monitors_for_clause | readable_if_clause | writable_if_clause
   | datagroup_clause    | set_statement      | nowarn_pragma
   | accessible_clause   | assert_statement   | assume_statement
+  | datatype_declaration
   ;
 
 methodlevel_comment: (modifiers? methodlevel_element modifiers?)* EOF;
@@ -41,7 +42,28 @@ modifier
   | CODE_JAVA_MATH | CODE_SAFE_MATH | CODE_BIGINT_MATH)
  ;
 
+datatype_declaration:
+    DATATYPE
+    dt_name=IDENT
+    LBRACE
+    datatype_constructor (INCLUSIVEOR datatype_constructor)*
+    SEMI_TOPLEVEL
+    (datatype_function)*
+    RBRACE
+    ;
 
+datatype_constructor:
+  con_name=IDENT
+  (
+    LPAREN
+    (argType+=type argName+=IDENT
+     (COMMA argType+=type argName+=IDENT)*
+    )?
+    RPAREN
+  )?
+;
+
+datatype_function: type IDENT param_list method_body;
 
 class_axiom: AXIOM expression SEMI_TOPLEVEL;
 initially_clause: INITIALLY expression SEMI_TOPLEVEL;
--- a/key.core/src/main/java/de/uka/ilkd/key/logic/Namespace.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/logic/Namespace.java
@@ -184,7 +184,7 @@ public class Namespace<E extends Named> implements java.io.Serializable {
     }
 
     /** Convenience method to look up. */
-    public E lookup(String name) {
+    public @Nullable E lookup(String name) {
         return lookup(new Name(name));
     }
 
--- a/key.core/src/main/java/de/uka/ilkd/key/nparser/AdtHelper.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/nparser/AdtHelper.java
@@ -1,0 +1,225 @@
+package de.uka.ilkd.key.nparser;
+
+import de.uka.ilkd.key.java.Services;
+import de.uka.ilkd.key.logic.*;
+import de.uka.ilkd.key.logic.op.*;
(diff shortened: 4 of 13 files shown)


wadoon:
One of the primary todos might be to prepare a proposal for [JMLref](https://github.com/JavaModelingLanguage/RefMan).

mattulbrich:
OpenJML used "," where Daniel @Drodt proposed "|". Small detail. I think in the face of enums, "," would fit better into the Java biotope.

mattulbrich:
I just had a look at the axioms created for ADTs. There seems to be an unexpected "not free in" condition.
Perhaps we can review this, too, while we are working on it during the hackeython.

```
Color_Axiom {
  \schemaVar \formula phi;
  \schemaVar \variables Color x;
  
  \find(\forall x; phi)
  \varcond(\notFreeIn(x, phi))
  \add(
         \forall x; phi
     <->   {\subst x;Red}phi
         & {\subst x;Green}phi
         & {\subst x;Blue}phi
    ==>
  )
  \displayname "Axiom_for_Color"
}
```

mattulbrich:
Discussion at the HackKeYthon:

* free generated: important, but what about not-free variants (like finite sets, maps)
* `switch` with pattern matching as generalisation of `if` in KeY JavaDL.
* Termination witness proofs not needed for subterms of termination witness inputs (at least when using pattern matching)
* Syntax and Java-similarity: Merge the intuition of enum and record.
* Selectors are important
* Parametric polymorphism is important

@TudorBalan

mattulbrich:
> > Discussion at the HackKeYthon:
> > free generated: important, but what about not-free variants (like finite sets, maps)
> Can you elaborate what you mean by free here?

Free here means that different ground terms denote different semantic values. Like a language with concatenation the free monoid on the alphabet (words are equal only if the same sequence of letters. Lists in this sense are free (wrt. to their constrcutors) while finite sets are not ({1,2} being the same as {2,1}) (wrt to insert and empty).

> > Termination witness proofs not needed for subterms of termination witness inputs (at least when using pattern matching)
> > ```
> What are the issues when not pattern matching?

In the example of a list: Calling `a.tail()` is not necessarily decreasing because a can be nil. But if a variable is bound within a pattern matching, then it is syntactically clear that this variable is smaller in the sub-adt sense. Otherwise it comes with some semantic conditions (like `a != nil` for `a.tail()`).

> > Syntax and Java-similarity: Merge the intuition of enum and record.
Enum: List the different possibilities of values as elements in the type body
Records: List the named values with types as arguments to the constructors.
In both cases: Allow further operations after ";".
> What intuition? How do you plan to merge?

The similarity with enums would obviously lead to a comma as separator between constructors. :stuck_out_tongue: 
But pipe is fine, too, I guess.
Actually, Daniel @Drodt 's proposal has this already implemented. Is rather obvious. So does David Cok's in OpenJML.

> > Selectors are important
Selectors = Destructors

> > * Parametric polymorphism is important
> Any progress on that front? Did you consider the more type class oriented approach or is that too dissimilar to Java?

See elsewhere ...

mattulbrich:
Another important observation is the following:

In order to avoid name clashes, constructors need to be prefixed with the type. I.e. for

    package pack.path;
    datatype List { Nil | Cons(int value, List list) }

one needs to write `pack.pah.List.Nil` to refer to nil to avoid name clashes.
As usual with Java, one can use import to be allowed to shorten this like:

    import static List.*;

Now `Nil` suffices ... In KeY, probably the full name should be used ...

mattulbrich:
> After implementing switch expressions into the parser for key files, we discussed how it should be represented as a term in key and the semantics of a switch expression

Thanks, @philipppk! A few more explaining words:

When translating such switch-case JML expressions into JavaDL, we can use existing infrastructure or use new infrastructure. Using the Hilbert-Epsilon operator is nice because it reuses existing structures. However, during normalisation the structure of the argument of the operator may be modified such that specialised taclet or pretty printing may no longer apply. And the epsilon-presentation is not exactly user-friendly.

When using a specialised JavaDL syntax with new constructors (cases, nodefault, default, switch), normalisation cannot destroy the structure of expression, yet we need many new rules to deal with the situation.

At any rate:

"if" should/could be retrofitted as a special case of switch case.
"ifEx" has a certain similarity with the case function above, also regarding polymorphic properties....

mattulbrich:
I would like to to tell  David Cok from openjml about these advances and would like to mention that syntax-wise the languages are compatible. This would mean that we could perhaps support both `|` and `,` as separators between constructors for compatibility reasons.
OK?

mattulbrich:
*Followup in next KeY meeting:* We could introduce a switch operator per term with the appropriate signature thus avoiding a vari-arity symbol.

mattulbrich:
> _Followup in next KeY meeting:_ We could introduce a switch operator per term with the appropriate signature thus avoiding a vari-arity symbol.

Actually, it turns out that this can be generally implemented into KeY generally w/o touch a single line of Java code.

The example [switch.key](https://github.com/user-attachments/files/19070345/switch.key.txt) (pls remove the file extension .txt) shows how a new function symbol `switchList` can be introduced realising switch expressions for the List adt. It takes three arguments: `switchList(term, nilCase, consCase)` with the nilCase a constant value and the consCase a "double abstraction", i.e. a term of the form "λhead. λtail. term_using(head, tail)".

Those are not part of KeY at the moment, so I added a new type binding (whose intuition is any→any) with a constructor `bind` (whose intuition is λ abstraction) and an observer `G::apply` (whose intuition is function application of λ-calculus).
There is the obvious ß-reduction axiom that defines the setup.

Since KeY does not support parameterised types, it is all a bit ugly regarding the typing, but still sound.

wadoon:
I started with the JML implementation part. 
There are some questions about ADT in the real-world/key-world.

- How to deal with mutual definition of ADT in JML? Note JML comments are merged and a block is processed at the same time. 
- Are ADTs living in fully-qualified namespaces? 
- Does ADTs need to be imported? (e.g., `//@ import model my.adt.List;`) 
- Inner ADTs are allowed? 
- ...

wadoon:
Short floor discussion:

* ADTs should be fully-qualified 
* ADTs should also be possible on Type-declaration-level (directly in compilation units)
* model-import should be possible (but hard to implement in recoder)

wadoon:
@mattulbrich @WolframPfeifer 

Just a quick thought: Why do we invent our own ADT stuff when there are established tools for functional verification? Would it not be funny to have something like "functional Java" (which is obviously F*/Haskell with Java-style Syntax)? (Or do we have reached this point with allowing more stuff in model methods?)

mattulbrich:
> Just a quick thought: Why do we invent our own ADT stuff when there are established tools for functional verification? Would it not be funny to have something like "functional Java" (which is obviously F*/Haskell with Java-style Syntax)? (Or do we have reached this point with allowing more stuff in model methods?)

That is indeed the idea of switch expressions, model methods and ADT declarations. ...

PR #3430, opened 2024-02-22
Generate easier POs for non-trivial diverges clauses

### LINKED ISSUE #29: Non-trivial diverges clauses make you prove things twice.
This issue was created at [git.key-project.org](https://git.key-project.org/key/key/-/issues/29) where the discussions are preserved.
    
----
    
* Mantis: [MT-1499](http://i12www.ira.uka.de/~klebanov/mantis/view.php?id=1499)
* Submitted on: 2014-10-28 by (at)mulbrich
* Updated: 2014-10-29

### Description

> When KeY encounters a non-trivial (i.e. neither true nor false) diverges clause,
> it creates two contracts: one with <>- and one with []-modality.
> 
> The negation of the diverges condition is added as an assumption to the 
> <> contract, but no assumption is made for the [] case.
> 
> This implies that the cases proved in <> need to repeated in [].
> Adding another assumption would reduce verification effort in such cases.

### Steps to reproduce

```
Load the following class into KeY
class A {
    boolean div, post;

    /*@ diverges div; 
      @ ensures post;*/
    void m() {}
}

you will get 2 proof obligations which read (essentially)

   !div -> <m()> post
and 
    [m()] post

The latter should be 
   div -> [m()] post
```

## Notes

_(at)grahl at 2014-10-29_
> An alternative PO would be [m()]post & (!div -> <m()>true). This is equivalent to the two above and easier to prove than the current version (but maybe not easier than Mattias' suggestion). 
> 
> This formulation reflects more the intuition behind diverges. But Mattias' suggestion is better for proving, since it provides usable DL contracts (with nontrivial postconditions).
> 
> 

## History

* (at)mulbrich -- (`NEW_BUG`) _2014-10-28_

* (at)grahl -- (`BUGNOTE_ADDED`) _2014-10-29_

* (at)grahl -- (`BUGNOTE_UPDATED`) _2014-10-29_

## Attributes

* Category: JML (semantics)
* Status: NEW
* Severity: MINOR
* OS: 
* Target Version: 
* Resolution: OPEN
* Priority: LOW
* Reproducibility: HAVENOTTRIED
* Platform: 
* Commit: 56061ba8f523ccb3a8d58d92447c4bad60b7c372
* Build: 
* Tags []
* Labels:  ~JML (Semantics)  ~Bug  ~LOW  

---

[View in Mantis](http://i12www.ira.uka.de/~klebanov/mantis/view.php?id=1499)
    
----
    
Information: 
* created_at: 2017-05-29T02:26:22.421Z
* updated_at: 2021-11-07T13:41:41.579Z
* milestone: 
* user_notes_count: 0

Drodt:
## Related Issue

This pull request fixes #29.

## Intended Change

A non-trivial (i.e, neither `true` nor `false`) diverges clause results in two contracts to be verified. Previously, there was one where the value of the clause (here `div`) is not true and the method therefore terminates, and one where the method does not terminate.

E.g., the two POs would look like `pre & !div -> <m()>post` and `pre -> [m()] post`. This resulted in more complex proofs, because the case where `div` is false had to be considered as well for the second PO.

This PR changes the POs to
1. `pre & !div -> <m()>post` and
2. `pre & div -> [m()] post`

Hence, we have a proper cut.

## Type of pull request

- [ ] Other: 

## Ensuring quality
    

key.core/src/main/java/de/uka/ilkd/key/speclang/jml/translation/JMLSpecFactory.java
```diff
@@ -926,10 +926,13 @@ public ImmutableSet<Contract> createFunctionalOperationContracts(String name, IP
             result = result.add(contract);
         } else {
             // create two contracts for each diamond and box modality
+            Map<LocationVariable, Term> boxPres = new LinkedHashMap<>(pres);
```
FliegendeWurst, line 929: Shouldn't this be a copy of `clauses.requires`? Previously that map was used when constructing the FunctionalOperationContract.

WolframPfeifer:
@Drodt Just looking through old PRs ... What is the state of this one? Is it finished and reviewable?

key.core/src/main/java/de/uka/ilkd/key/speclang/jml/translation/JMLSpecFactory.java
Drodt: That seems like an oversight in the old version. I see no reason why this should be `clauses.requires`. The pre-conditions should be identical for diamond and box modalities.

mattulbrich (approved):

Issue #1663, opened 2022-12-23
Parse Java files with unknown library methods


wadoon:
This issue was created at [git.key-project.org](https://git.key-project.org/key/key/-/issues/1663) where the discussions are preserved.
    
----
    
## Description

The story is well-known (see #491) and a solution exists but I think there could be a better solution. If this turns out to be completely unfeasible, we can of course close this issue.

As usual, you want to prove something but you can't parse the Java sources, since there is a method call to some (external/core) library. And the method signature is not present. However, you don't really need a contract or the implementation as the present proof obligation does not depend on this method call.

## Example
[SimpleMin.java](/uploads/fd8c24c629fa01a6bb4801e244cb541f/SimpleMin.java)
```java
public class SimpleMin {

    /*@ public normal_behavior
          requires true;
          ensures a >= b ==> \result == b;
          ensures a <  b ==> \result == a;
      (at)*/
    public static int min(int a, int b) {
        int m = a;
        if (b < m) m = b;
        a = b = 0;
        return m;
    }

    public static void main (String args[]) {
        int a = 3;
        int b = 7;
        System.out.printf("%d = min(%d, %d)\n", min(a,b), a, b);
    }
}
```

## Status Quo

You could either comment out every such method call (in the example to `System.out.printf`) or provide add a method stub in a `\classpath`.

## Suggestion

Automatically infer a method signature from the method call (in the example `Top printf(String, int, int, int)`). Then there is still no contract or implementation but we would be able to parse the sources without issues. Only later, if the symbolic execution reaches such a method, we would get stuck in the proof.
    
----
    
Information: 
* created_at: 2022-01-06T18:22:25.637Z
* updated_at: 2022-01-11T17:28:43.177Z
* closed_at: None  (closed_by: )
* milestone: 
* user_notes_count: 5

PR #3431, opened 2024-02-22
Hackeython unknown methods

flo2702:
## Related Issue

This pull request addresses #1663

## Intended Change

When using a library method without an explicit contract, instead of getting stuck in the proof, KeY now assumes the following default contract

```
public behavior
    ensures true;
    diverges true;
    signals_only Throwable;
    assignable \everything;
```

In addition, there is a proof setting `soundDefaultContracts` which can be turned off. In this case, the following default contract is used instead, which may be useful for exploratory proofs

```
public normal_behavior
    ensures true;
    assignable \strictly_nothing;
```

## Type of pull request

- [ ] Other: 

## Ensuring quality
    
- [x] I made sure that new/changed end-user features are well documented (https://github.com/KeYProject/key-docs).

This was done during HacKeYThon 2 by @JonasKlamroth and @flo2702 


mattulbrich (approved):

WolframPfeifer:
I just stumbled across this PR recently and noticed that my original use case is not yet solved:
```
import java.util.Arrays;

class C {
    public static void main(String[] args) {
        int[] a = {5,7,3,9,3,5,9,0,2};
        System.out.println(Arrays.toString(a));
    }
}
```

Trying to load this file results in an exception of the following form:
```
Could not resolve MethodReference "Arrays.toString(a)" @6/28 in ...
...

Caused by: recoder.service.UnresolvedReferenceException: Could not resolve MethodReference "Arrays.toString(a)" @6/28 in ...
```

I noticed that there is a taclet option `soundDefaultContracts` now, but it is undocumented. Also, the exception above occurs independent of the value of the taclet option. I wonder:
* What is the effect of this PR exactly?
* How does it work with imports?
* How does it work for classes that are present in the library (aka JavaRedux)?

@flo2702 @JonasKlamroth @mattulbrich (addressing all authors or reviewers):
* Could one of you add a documentation for the taclet option in optionsDecl.key?
* Would it make sense to have a short section about the feature in the KeY developer manual (ideally with a link to it in the error message)?
* Feature Request: List the auto-generated contracts somewhere, make them viewable (e.g. in "Proof"->"Show Used Contracts", marked as "inferred" or similar). Or, are they already listed?

PR #3440, opened 2024-03-08
Allow realizing multiple cached branches at once

FliegendeWurst:
## Intended Change

This PR changes the "Copy referenced proof steps here" action to allow applying it on any node that has leaf children with caching info.
It will then copy the referenced proof steps for all cached branches.

## Type of pull request

- [ ] Other: 

## Ensuring quality
    
- [x] I have tested the feature as follows: manual test (UI code)


WolframPfeifer (changes requested):
This feature worked for me during review. Thanks! However, the user feedback should be improved a bit in my opinion: I nearly always get a dialog listing (only) the nodes where the caching failed (in the screenshot, actually two of the three goals are closed via cache):
image

A bit unrelated, I have another small feature request: "Prune Proof" should work with goals that are closed by cache. At the moment, I have to click on "Proof caching" -> "Re-open cached goal", which is a bit cumbersome. From a user point of view, both are the same operation (cutting away proof nodes).

FliegendeWurst:
> I nearly always get a dialog listing (only) the nodes where the caching failed (in the screenshot, actually two of the three goals are closed via cache):

I changed this dialog to list both successful matches and mismatches.

> A bit unrelated, I have another small feature request: "Prune Proof" should work with goals that are closed by cache.

Done

WolframPfeifer:
I just updated this to the new main branch. The changes in ProofPruner don't work (they seem to add an entry `null` to the list of open goals, which then later results in a NPE). This is also the cause for the test failure. Since the interaction between proof pruning and caching is a separate case (and seems to need a bigger fix), we should try to fix that in a separate PR.

I tested the rest of the functionality and it seems to work, with one small exception: The button in the status bar, which shows the number of cached goals, displays the wrong number after closing multiple cached goals via `Context menu --> Proof Caching --> Copy referenced proof steps here` (looks like the last number before closing - 1).

WolframPfeifer:
I removed the ProofPruner changes as they were broken, and made sure the UI (taskbar button) is updated.

WolframPfeifer (approved):

PR #3445, opened 2024-03-21
fix smt solver downloader script

wadoon:
Fix the downloading of SMT solver in the CI pipeline. 

Version of z3 and CVC5 are now pinned.

### DIFF
--- a/.github/dlsmt.sh
+++ b/.github/dlsmt.sh
@@ -56,7 +56,7 @@ echo "::endgroup::"
 #################################################
 echo "::group::{install cvc5}"
 if -f cvc5-Linux-static/bin/cvc5; then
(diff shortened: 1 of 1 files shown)


.github/dlsmt.sh
```diff
@@ -51,14 +55,19 @@ echo "::endgroup::"
 
 #################################################
 echo "::group::{install cvc5}"
-if -f cvc5-Linux; then 
+if -f cvc5-Linux-static/bin/cvc5; then
   echo "::notice::{Z3 found. Caching works! Skip installation}"
```
mattulbrich, line 59: I assume that should read "CVC5 found". Right?

mattulbrich (approved):
Just a typo in a message.

PR #3446, opened 2024-03-22
Update pull_request_template.md

mattulbrich:
## Intended Change

Updated and improved the pull request template.

Removing the checkboxes which always (if unticked) indicated that the PR is only partially finished in the overview. Replaced by item list from which items can be deleted.

Refining the structure. Adding a new subsection "Plan".

## Type of pull request

## Ensuring quality
    
n/a

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -2,8 +2,8 @@
 Thanks for submitting this pull request for KeY.
 Since the project has a strict review policy, please make the
 reviewer's job easier by providing the necessary information
-in the text below. The comments may remain since they will be
-invisible when showing the PR.
+in the text below. The comments can be deleted, but may also 
+remain since they will be invisible when showing the PR.
 -->
 
 ## Related Issue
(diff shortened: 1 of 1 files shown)


wadoon:
Can we get rid of the not semantically-correctly used todo items?

* [ ] Yes
* [ ] No
* [ ] Maybe?

I find it very annoying if you look on a PR and Github says "5 out of 10 tasks done".

WolframPfeifer (approved):
Thanks, I also find this very annoying. Unfortunately, checkboxes that are not task lists are not really supported by GitHub at the moment, even though this feature that was requested multiple times. One of these feature requests also mentioned a workaround: You can wrap checkboxes in a list, then they are not counted as tasks:
- - [x] option 1
- - [ ] option 2
- - [ ] option 3
While this may mess up the spacing a bit (conceptually, this is a list of lists then), we could also use this. However, I am also quite happy with the template suggested in this PR.

mattulbrich:
> I find it very annoying if you look on a PR and Github says "5 out of 10 tasks done".

That's the reason for the PR. If we have checkboxes than we should have them for things were 10/10 is the desired goal, and fewer should not be merged.

PR #3447, opened 2024-03-22
Update dependabot configuration

wadoon:
This PR reconfigures *dependenabot* in the following manner:

1. Group incoming updates into a single Pull Request
> Sometimes there are 3 or 4 PRs created by dependabot. Now they are bundled into one combined PR. 

2. Slow down the checking of new versions from weekly to monthly
> We are not very security-critical. This helps to prevent downloading new libs every week.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,16 +8,16 @@ updates:
   - package-ecosystem: "gradle"
     directory: "/"
     schedule:
-      interval: "montly"
+      interval: "monthly"
     groups:
-      all:
+      gradle-deps:
         patterns:
           - "*"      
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -5,7 +5,19 @@
 
 version: 2
 updates:
-  - package-ecosystem: "gradle" # See documentation for possible values
-    directory: "/" # Location of package manifests
+  - package-ecosystem: "gradle"
+    directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "montly"
```
mattulbrich, line 11: should probably be "mont**h**ly"

mattulbrich (approved):

PR #3453, opened 2024-04-10
Fixes a StackOverflow when pretty printing a taclet

unp1:
## Intended Change

Fixes a StackOverflowException when pretty printing taclets, which occurred
for instance while showing a tooltip of program with a modality and blocked
the subsequent taclet application.

## Type of pull request

Bug fix (non-breaking change which fixes an issue)

## Ensuring quality

- Added a test that fails in the unfixed version (i.e., without this PR)
- I have tested the feature as follows: Recreated the scenario where previously a StackOverflow happened
  (selecting the loopscope taclet and pretty printing its taclet)

### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/pp/LogicPrinter.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/pp/LogicPrinter.java
@@ -1669,9 +1669,9 @@ public class LogicPrinter {
                     for (int i = 0; i < phi.arity(); i++) {
                         ta[i] = phi.sub(i);
                     }
-                    JavaBlock jb1 = mod.program();
-                    Modality m = Modality.getModality(mod.kind(), jb1);
-                    Term term = services.getTermFactory().createTerm(m, ta,
+                    final Modality m =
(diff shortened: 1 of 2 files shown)


key.core/src/main/java/de/uka/ilkd/key/pp/LogicPrinter.java
```diff
@@ -1669,9 +1669,9 @@ public void printModalityTerm(String left, JavaBlock jb, String right, Term phi,
                     for (int i = 0; i < phi.arity(); i++) {
                         ta[i] = phi.sub(i);
                     }
-                    JavaBlock jb1 = mod.program();
-                    Modality m = Modality.getModality(mod.kind(), jb1);
-                    Term term = services.getTermFactory().createTerm(m, ta,
+                    final Modality m =
+                        Modality.getModality((Modality.JavaModalityKind) o, mod.program());
```
wadoon, line 1673: This will crash if new languages are added. Ok for now.

wadoon (approved):

PR #3454, opened 2024-04-12
Boyer Moore Majority Vote

mattulbrich:
## Intended Change

Adding a new example to KeY, and to runAllProofs.

## Plan

This PR is ready.

## Type of pull request

-  non-breaking change which adds an example
- There are changes to the deployment/CI infrastructure (gradle, github, ...): added test case

## Ensuring quality

- I have tested the feature as follows: added to runAllProofs
- I have checked that runtime performance has not deteriorated: n/a


Drodt:
I was working on the same example :)

Drodt (approved):

mattulbrich:
Ah. We should have coordinated :) Let's compare solutions!

Drodt:
The code was very similar of course but I wanted to try if I could use the partition approach they originally used. Couldn’t find a good source for it though—thanks to great hotel wifi.

PR #3460, opened 2024-04-20
Add key features for the FM tutorial

wadoon:
## Features

- Support of `\array2seq` and `\seq_put` (and `\seq_upd`) in JML
  (Currently hard coded. It could be better if we add meta-data to functions or have theories)

- Fix `\locset(x.*,a[*])`, merge with \storeref function

Tested with the current of the FM tutorial.

### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/speclang/njml/Translator.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/speclang/njml/Translator.java
@@ -1039,11 +1039,10 @@ class Translator extends JmlParserBaseVisitor<Object> {
             try {
                 return processJmlBuiltInFunction(fullyQualifiedName, accept(ctx.expressionlist()));
             } catch (Exception e) {
-                // weigl: only log error. There is no reason why a method should not start with a \\
-                // in JML.
-                // The next version will have, e.g., \seq and \map classes.
+                // weigl: only log error. There is no reason why a method should not
+                // start with a "\" in JML. The next version will have, e.g., \seq and \map classes.
                 addWarning(ctx,
-                    "%s was not understand as a JML built-in function. I assume it to be a method or class."
+                    "%s is not known as a JML built-in function. I assume it to be a method or class."
                             .formatted(fullyQualifiedName));
             }
         }
@@ -1076,6 +1075,7 @@ class Translator extends JmlParserBaseVisitor<Object> {
         // at some point in life, we may want to have a customizable map here.
         return switch (name) {
         case "\\array2seq" -> termFactory.translateToJDLTerm(name.substring(1), params);
+        case "\\seq_upd" -> termFactory.translateToJDLTerm("seqUpd", params);
         case "\\seq_put" -> termFactory.translateToJDLTerm("seqUpd", params);
         default -> throw new IllegalStateException("Unexpected value: " + name);
         };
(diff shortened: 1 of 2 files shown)


key.core/src/main/java/de/uka/ilkd/key/speclang/njml/Translator.java
```diff
@@ -1031,6 +1033,21 @@ public Object visitPrimarySuffixCall(JmlParser.PrimarySuffixCallContext ctx) {
...
+        // support JML built-ins
+        if (fullyQualifiedName.startsWith("\\")) {
+            try {
+                return processJmlBuiltInFunction(fullyQualifiedName, accept(ctx.expressionlist()));
+            } catch (Exception e) {
+                // weigl: only log error. There is no reason why a method should not start with a \\
+                // in JML.
+                // The next version will have, e.g., \seq and \map classes.
+                addWarning(ctx,
+                    "%s was not understand as a JML built-in function. I assume it to be a method or class."
```
mattulbrich, line 1046: language: "%s is not a built-in JML function name. I assume it is a method or class name."

key.core/src/main/java/de/uka/ilkd/key/speclang/njml/Translator.java
```diff
@@ -1397,20 +1442,22 @@ public SLExpression visitPrimaryCreateLocsetSingleton(
...
-        if (seq.isEmpty()) {
-            raiseError(ctx, "empty!");
-        } else if (seq.size() == 1) {
-            ret = seq.get(0);
-        } else {
-            ret = tb.union(seq);
-        }
-        return new SLExpression(ret, javaInfo.getKeYJavaType(PrimitiveType.JAVA_LOCSET));
-    }
+    /*
```
mattulbrich, line 1445: Why is this unformatted  now? Why not keep the original indentation?

key.core/src/main/java/de/uka/ilkd/key/speclang/njml/Translator.java
```diff
@@ -1054,6 +1071,16 @@ public Object visitPrimarySuffixCall(JmlParser.PrimarySuffixCallContext ctx) {
         return result;
     }
 
+    private SLExpression processJmlBuiltInFunction(String name,
+            ImmutableList<SLExpression> params) {
+        // at some point in life, we may want to have a customizable map here.
+        return switch (name) {
+        case "\\array2seq" -> termFactory.translateToJDLTerm(name.substring(1), params);
+        case "\\seq_put" -> termFactory.translateToJDLTerm("seqUpd", params);
```
mattulbrich, line 1079: That should probably be unified mit `visitSequenceReplace`. I like this method here since it points into a nice configurable direction.

I think something like the following might be nice, ... eventually:
```
functions {
   \jml Seq array2seq(Heap, Object);
   \jml Seq Any head(Seq);
}
```

key.core/src/main/java/de/uka/ilkd/key/speclang/njml/Translator.java
```diff
@@ -1329,57 +1356,75 @@ public SLExpression visitPrimaryReachLocs(JmlParser.PrimaryReachLocsContext ctx)
...
-            }
-        } else {
-            // Array access
-            SLExpression index = accept(ctx.expression());
-            return termFactory.arrayRef(base, fullyQualifiedName, index, null);
-        }
-        assert false;
-        return null;
-    }
+    /*
```
mattulbrich, line 1359: Why is this unformatted now? Why not keep the original indentation?

key.core/src/main/java/de/uka/ilkd/key/speclang/njml/Translator.java
mattulbrich: The tutorial does now not longer contain "seq_put" (well, it did not contain it before?)

key.core/src/main/java/de/uka/ilkd/key/speclang/njml/Translator.java
wadoon: :question:

key.core/src/main/java/de/uka/ilkd/key/speclang/njml/Translator.java
wadoon: > That should probably be unified mit `visitSequenceReplace`. I like this method here since it points into a nice configurable direction.

I would rather prefer the following when we finally achieve theories: 

```
\theory JML {
  \from Seq \import seqUpd \as seq_upd
}
```

Everything imported or exported from JML theory is available in the JML namespace.

PR #3462, opened 2024-04-21
Removal of JUnit4

wadoon:
This PR removes the remaining traces of JUnit 4 completely from KeY. JUnit 5 offers the same functionalities in a simpler matter, therefore there is no need to keep JUnit 4.

I also updated the recoder module thanks to IntelliJ migration support.


WolframPfeifer (approved):
Everything looks good, thanks!

Issue #3407, opened 2024-02-13
Support for JML `TYPE`


flo2702:
KeY should have more support for the JML type `TYPE`.

As far as I know, KeY only supports the `\type` construct in expressions of the exact form `\typeof(e) == \type(C)`. It does not understand the type `TYPE` and the operator `<:` at all.

## Use case 1

Consider a class hierarchy like the following:

```
class A {
    //@ ghost TYPE packed;
    //@ invariant packed <: \type(A) ==> P;
}

class B extends A {
    //@ invariant packed <: \type(B) ==> Q;

    //@ requires this.packed == \type(B);
    //@ ensures P;
    //@ ensures Q;
    void foo() {}
}
```

The method contract of `foo()` should be provable.

## Use case 2

Remembering the type of objects in a `\seq` or container object.

wadoon:
Does this hurt the theory on the meta-level? 

* What is `\type(\type(a))`? 
* Is `\type` only defined for reference datatypes?
* Is `\type(2) well-defined? Or is `\type(2)` a reference type due to autoboxing?
* Is `\TYPE <: \TYPE`? 
* Is `x <: y` similar to `instanceOf`? Then, will it support annotations?

flo2702:
`\type(\TYPE) <: \type(\TYPE)` holds.

`\type` is defined for type names. I.e., `\type(ClassName)`, \type(int)`, and `\type(\TYPE)` are all legal expressions, but `\type(\type(...))` and `\type(2)` are not.

The subtype operator supports only base types without annotations.

PR #3465, opened 2024-04-26
Support for JML`\TYPE`

flo2702:
## Related Issue

This pull request addresses #3407.

## Intended Change

KeY now supports the type `\TYPE`, the subtype operator `<:`, and the function `\type`.

The following method contract illustrates their semantics:

```
class A {

    int f;

    //@ ghost \TYPE packed;
    //@ invariant packed <: \type(A) ==> f > 0;
}

class B extends A {
    //@ invariant packed <: \type(B) ==> f > 1;

    //@ requires this.packed == \type(B);
    //@ ensures f > 0;
    //@ ensures f > 1;
    void subtypes() {}
}
```

## Type of pull request

## Ensuring quality
    

key.core/src/main/java/de/uka/ilkd/key/java/visitor/Visitor.java
```diff
@@ -406,4 +406,6 @@ void performActionOnCcatchBreakWildcardParameterDeclaration(
      * @param jmlAssert the statement to perform the action on.
      */
     void performActionOnJmlAssert(JmlAssert jmlAssert);
+
+    void performActionOnSubtype(Subtype subtype);
```
wadoon, line 410: This does not make any sense to me. Why do you need a Java AST node?

key.core/src/main/java/de/uka/ilkd/key/java/expression/operator/Subtype.java
wadoon, line 1: A Java AST node is unnecessary since the latest updates on set and assert statements.

wadoon (changes requested):

key.core/src/main/java/de/uka/ilkd/key/java/expression/operator/Subtype.java
wadoon: Issue will be created and clean up is for a different PR.

wadoon (approved):

PR #3468, opened 2024-05-02
Nullness Type System activated for `key.ncore`

wadoon:
Nullness type checker activated in `key.ncore`.


wadoon:
@flo2702 It compiles now. On `key.ncore`, the checker framework is misconfigured. 

If I enable the `ENABLE_NULLNESS`, I get an error in `ImmutableMap<K,V>`.

flo2702:
Do we write `@Nullable private T field` or `private @Nullable T field`?

I prefer the second option. Unless you have strong reasons for the first option, I will change this before merging.

wadoon:
> Do we write `@Nullable private T field` or `private @Nullable T field`? (Same question applies to method return types)

I have no preferences. And for me this is also no quality issue. 

Maybe a good reformatter can reformat this.

flo2702 (approved):

PR #3470, opened 2024-05-03
Nullness Type System for `key.core` 

wadoon:
More nullness type system checks. Now for `key.core`.

I want first to merge #3399 into KeY to avoid double work.


wadoon:
I have pushed further and activated Nullness checks for more packages but have not finished everything. Annoying are the situations where the `InitilizationChecker` hits: 

```java
class A {
  public A() { ... 
    setFoo(2);
  }
  void setFoo(int i){...}
}
```

The complain is that `setFoo` is called with `@Unintialized A this` but expects `@Initialized A this`. 

@wmdietl @flo2702 
Is there any trick to get pass this message? For example, an 
```java
assert this instanceof @Initialized A;
```
would be fine. 

Inside KeY, we have many of these moments. It is also not a bad code design and is hard to avoid. Also adding the right annotation leads to ugly code: 


```java
class A {
  public A() { ...; setFoo(2);  }
  void setFoo(@Uninitialized A this, int i){...}
}
```

and this needs to be added to all transitive called methods.

wadoon:
@wmdietl I saw that this error is correct. However, I want to avoid the clutter of receiver types in the source code.

wmdietl:
@wadoon If this only happens in a few constructors, simply suppress the warning there. 
If you never care about warnings from the Initialization Checker, pass `-AassumeInitialized` to turn it off.
You might be able to get something in-between by passing `-AsuppressWarnings=` with the error key from the call... https://eisop.github.io/cf/manual/manual.html#suppresswarnings-command-line

Do you see any other option that we could add to the framework?

wadoon:
For the first, I stick to `-AassumeInitialized`. 

This pattern of calling own methods in constructor is rather common in UI code, e.g., `AbstractAction` in Swing. 

> Do you see any other option that we could add to the framework?

I am thinking about, why is this so easy in Kotlin, and in Java so complicated. 

My first thought was that a simpler annotation would be fine. At least lower the clutter. 

```kotlin
typealias @ActionInit = `@UnknownInitialization(javax.swing.AbstractAction.class) KeYAction this`;
public class KeyAction { @ActionInit public KeyAction setName(...) {...} }
```

My second thought was, why am I not able to stop the "under-initialization phase" by calling `this(..);` or getting an `UnknownInitialization(A.class) A this` by calling `super();`

Here is the excerpt on how we typically build our Swing Actions: 

```java
abstract class javax.swing.AbstractAction  implements ActionListener {
   private Map<> attrib = new HashMap<>();
   public void putValue(Object key, Object value);
}

class KeyAction extends AbstractAction {   
    void setName(String name) { putValue(NAME, name);} 
}

class AboutDialogAction extends KeyAction {
    public AboutDialog() { 
        super(); // normally implicit
        setName("About..."); // <--- should be possible KeyAction is constructed.
    }
}
```

Of course, there might be loopholes (`super()` calls an overridden method) for certain scenarios this should be valid, especially when a class is "encapsulated under inheritance" iff. I could use the parent class also via delegation.

wmdietl:
> I am thinking about, why is this so easy in Kotlin, and in Java so complicated.

Kotlin doesn't check initialization:

Take

````kotlin
fun main() {
    println(Sub().toString())
}

open class Super {
    var f: String
    
    init {
        foo()
        f = ""
    }
    
    open fun foo() {
    }
}

class Sub : Super() {
    init {}
    override fun foo() {
        f.hashCode()
    }
}
````

This does not produce any warnings from the compiler and results in an NPE.

The default in Kotlin to make classes and methods final helps a bit, but for your example you would still need open classes and methods.

The Nullness Checker aims to soundly handle initialization, which requires some specifications.


> My first thought was that a simpler annotation would be fine. At least lower the clutter.

I've filed an issue for this proposal: https://github.com/eisop/checker-framework/issues/1076
As there is perfectly fine syntax in Java 8+, I'm not so convinced this would be super useful.
In your example, only one method declaration, `setName`, would need a receiver. There is no overhead in the calls.


> Of course, there might be loopholes (super() calls an overridden method) for certain scenarios this should be valid, especially when a class is "encapsulated under inheritance" iff. I could use the parent class also via delegation.

The Initialization Checker does precisely keep track of `this()` and `super()` calls and refines the initialization frame accordingly.
Your example could have a further subclass of `AboutDialogAction`, which could override `setName` and cause an NPE.
By documenting the receiver type you would make explicit that the receiver is partially initialized and make this explicit for any overrides.

**Comment by wadoon** [comment]
> > My first thought was that a simpler annotation would be fine. At least lower the clutter.
> 
> I've filed an issue for this proposal: [eisop/checker-framework#1076](https://github.com/eisop/checker-framework/issues/1076) As there is perfectly fine syntax in Java 8+, I'm not so convinced this would be super useful. In your example, only one method declaration, `setName`, would need a receiver. There is no overhead in the calls.

`setName` is the tip of the iceberg. The full example would be [`KeyAction`](https://javadoc.io/doc/org.key-project/key.ui/latest/de/uka/ilkd/key/gui/actions/KeyAction.html) with 23 methods. Or [`SettingsPanel.java`](https://javadoc.io/doc/org.key-project/key.ui/latest/de/uka/ilkd/key/gui/settings/SettingsPanel.html) and [`SimpleSettingsPanel`](https://javadoc.io/doc/org.key-project/key.ui/latest/de/uka/ilkd/key/gui/settings/SimpleSettingsPanel.html) which both follows an inheritance-builder philosophy. (You inherit a panel, and the super-class provides simple methods for constructing the UI in a certain fashion.)

**Comment by lks9** [comment]
Code fixed during null check annotations:
* repaired {ZipFileCollection, DirectoryFileCollection}.opencurrent(FileRepo) accoriding to interface docs

**Comment by as2-0** [comment]
In the package key.core/src/main/java/de/uka/ilkd/key/macros/ there are methods with ~20 implm. Some of the paramters are sometimes only used in a few or maybe even none of the implementations (I didn't had the time to check).
Maybe its better to refactor them.

**Comment by wadoon** [comment]
KaKeY: 
- First: Make the PR fit again. 
- Second: Make the annotation runtime checkable.

**Comment by mattulbrich** [comment]
*Agreed during meeting:* 
- new checkerframework version to avoid errors
- merge main into this
- find a RAC framework for `@NonNull` (I wrote one a while ago, ...)
- then merge this into main (medium term) even if not verifiable to allow for compatibility.

**Comment by wadoon** [comment]
I guess this PR is rather dead now due to being impossible to rebse/merge.

**Comment by lks9** [comment]
At least it was a lot of work, what do you mean by "rather dead"? Not useful anymore?

**Comment by wadoon** [comment]
> At least it was a lot of work, what do you mean by "rather dead"? Not useful anymore?

A long time ago, we already concluded that `main` has diverged so far from this branch that nobody has the power to go through hundreds of merge conflicts, so this PR is de facto dead. 

Of course, we could let this PR just be open, but why? 

If you like to do the merging, just feel free.

**Comment by WolframPfeifer** [comment]
The question is: Is there a way to transfer the results achieved here to the current main (at least partially)? I think that we should not just close this, since it effectively makes the branch not findable any more, and there is even the danger that it is deleted without knowing what it is about.

**Comment by wadoon** [comment]
> The question is: Is there a way to transfer the results achieved here to the current main (at least partially)? 

This is something I discussed with Florian back then in 2024, as it had already gotten worse. The conclusion was that the merge conflicts are not worth resolving: 

1. There are too many of them, 
2. The monotony of the work will just lead to errors, and
3. You also need to look into these classes again if they have been affected by a change indirectly. 

The only resolution strategy was: If there is a change, we drop the version in this PR. 

```
git merge -X theirs origin/main
```

And then you still need to fix several hundred conflicts. 

I looked into the current state of this PR now. Git tells me that **3066 files** are in conflict. Focusing on Java, ignoring recoder, we are ending up with 1760 files. GitHub tells me that this PR touches 1320. 

I plea we should just jump to *acceptance* and not fall into the other four steps. 

If we really want this endeavour, we really need to work on execution.  

> I think that we should not just close this, since it effectively makes the branch not findable anymore, and there is even the danger that it is deleted without knowing what it is about.

I do not care if this branch is findable. The branch is now rotted and will further rot. 
There is no benefit in looking for an old branch for maybe a handful of useable nullness annotations. In the looking-up time, you can just annotate the classes by yourself in the updated version. 

Also: The PR list is not a museum. If PRs are not *actively* followed, we should just close them.

lks9:
> > The question is: Is there a way to transfer the results achieved here to the current main (at least partially)?
> 
> This is something I discussed with Florian back then in 2024, as it had already gotten worse. The conclusion was that the merge conflicts are not worth resolving:

This PR was part of the hackeyton 2025. If it was already in that state in *2024*, it was a mistake to work on it in the first place. A discussion in October 2025 seems more likely. But why mention an undocumented discussion from a few months ago now and not back then? Just a small note here (whenever you reached the conclusion) should be a trivial step. I really do not understand this.

> I looked into the current state of this PR now. Git tells me that **3066 files** are in conflict. Focusing on Java, ignoring recoder, we are ending up with 1760 files. GitHub tells me that this PR touches 1320.

> > I think that we should not just close this, since it effectively makes the branch not findable anymore, and there is even the danger that it is deleted without knowing what it is about.
> 
> I do not care if this branch is findable. The branch is now rotted and will further rot. There is no benefit in looking for an old branch for maybe a handful of useable nullness annotations. In the looking-up time, you can just annotate the classes by yourself in the updated version.

So, in other words, there is no way to transfer the results (partially) to the current main. Then it is reasonable to close this. But I fear that the future work on another branch on nullness annotations might also end up getting rotten.

> Also: The PR list is not a museum. If PRs are not actively followed, we should just close them.

I agree.

wadoon:
> This PR was part of the hackeyton 2025
Did I mess up with the year... 

After this PR disaster, nobody will dedicate work to a PR only to add Nullness annotation anymore.  

If you renovate or update some code pieces, you might add Nullness annotation to the new code, but then looking at this PR is just a waste of time. 

Adding Nullness annotation to your PR takes roughly a few minutes (or you ask your Agent now).

If you see it differently, then you should try to bring this to the main level.

PR #3472, opened 2024-05-22
Realize generic navigation structure

Drodt:
## Related Issue

This pull request depends on #3436.

## Intended Change

Previously, the `SyntaxElement` interface was introduced with the intent of a common supertype and navigation structure for generic AST elements. This PR adds functionality for navigation through ASTs. It serves as a ground for a rewrite of AST visitors, walkers, and matching; each of which will have their separate PR.

## Type of pull request

- Refactoring (behaviour should not change or only minimally change)
- New feature (non-breaking change which adds functionality)
- There are changes to the (Java) code

## Ensuring quality

    
- I made sure that the introduced/changed code is well documented (Javadoc and inline comments).
- I have tested the feature as follows: automatic tests

## Additional information and contact(s)

     

The contributions within this pull request are licensed under GPLv2 (only) for inclusion in KeY.

PR #3478, opened 2024-06-13
Remove SequentFormula class

unp1:
## Intended Change

This PR prepares the ground for generalizing proof and calculus structures.
It removes the superfluous SequentFormula class, which is only a behaviorless
container for a term. The class is a left-over from times when we had formulas with 
constraints on the sequent level

It depends on PR #3472 to be merged first as it is based on that branch

## Plan

* [ ] Pass tests
* [ ] Code cleanup

## Type of pull request

## Ensuring quality

* [ ] I have tested the feature as follows: tests
* [ ] I have checked that runtime performance has not deteriorated.


FliegendeWurst:
One advantage of the SequentFormula wrapper is that it is a bit harder to mix up top-level terms and subterms (newtype pattern).

PR #3482, opened 2024-06-14
RuleCommand can now deal with rules that have schema variables for logical variables

mattulbrich:
Rules in script could not be applied if there were free logical variable schema variables around.

This is the case for class invariant rules and others.

The commit fixes the issue by copying code from how it is done for interactive rule applications.

## Intended Change

More rules can be applied.

## Type of pull request

## Ensuring quality

- I have tested the feature as follows: FM case study

### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/macros/scripts/RuleCommand.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/macros/scripts/RuleCommand.java
@@ -61,8 +61,18 @@ public class RuleCommand extends AbstractCommand<RuleCommand.Parameters> {
         RuleApp theApp = makeRuleApp(args, state);
         Goal g = state.getFirstOpenAutomaticGoal();
 
-        if (theApp instanceof TacletApp) {
-            RuleApp completeApp = ((TacletApp) theApp).tryToInstantiate(g.proof().getServices());
+        if (theApp instanceof TacletApp tacletApp) {
+
+            ImmutableList<TacletApp> ifSeqCandidates =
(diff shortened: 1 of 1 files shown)


mattulbrich:
This PR is relevant for the FM tutorial.

@Drodt @WolframPfeifer @flo2702 @wadoon

wadoon (approved):

key.core/src/main/java/de/uka/ilkd/key/macros/scripts/RuleCommand.java
```diff
@@ -61,8 +61,18 @@ public void execute(AbstractUserInterfaceControl uiControl, Parameters args, Eng
         RuleApp theApp = makeRuleApp(args, state);
         Goal g = state.getFirstOpenAutomaticGoal();
 
-        if (theApp instanceof TacletApp) {
-            RuleApp completeApp = ((TacletApp) theApp).tryToInstantiate(g.proof().getServices());
+        if (theApp instanceof TacletApp tacletApp) {
+
+            ImmutableList<TacletApp> ifSeqCandidates =
+                    tacletApp.findIfFormulaInstantiations(g.sequent(), g.proof().getServices());
```
Drodt, line 67: This call fails if the `ifFormulaInstantiations` have already been instantiated.

Drodt (changes requested):

key.core/src/main/java/de/uka/ilkd/key/macros/scripts/RuleCommand.java
mattulbrich: should be fixed by another if.

Drodt (approved):

PR #3486, opened 2024-06-21
Map with new equalities

tobias-rnh:
## Related Issue

This pull request depends on PR #3459, which should be merged before reviewing this PR.

## Intended Change

This PR adds the new collection ``LinkedHashMapWrapper`` that can be used like the ``LinkedHashMap`` in key.core/de.uka.ilkd.key.util. The difference is, that elements are wrapped internally so that ``equalsModProperty()`` and ``hashCodeModProperty()`` are used instead of the normal ``equals()`` and ``hashCode()`` methods.

For this to work, ``hashCodeModProperty()`` is implemented for all properties introduced in #3459.

## Plan
~~Use general navigation structure instead of TreeWalker for hashCodeModThisProperty~~

## Type of pull request

## Ensuring quality

    
- I have tested the feature as follows: Already existing and new tests

     


Drodt (approved):

Drodt:
There are some minor merge errors. Besides that it looks fine

mi-ki:
Having a quick glance and check after this PR, the changes seem to make the taclet varcond TermLabelCondition non-functional. For example, in well-definedness checks, the according taclets (e.g., _wd_Logical_Op_And_ and _wd_Logical_Op_AndSC_) are not applicable anymore. Could you maybe have a further look at this?

tobias-rnh:
> Having a quick glance and check after this PR, the changes seem to make the taclet varcond TermLabelCondition non-functional. For example, in well-definedness checks, the according taclets (e.g., _wd_Logical_Op_And_ and _wd_Logical_Op_AndSC_) are not applicable anymore. Could you maybe have a further look at this?

That is curious as this PR should have only added hash-methods that were not used before and a HashMap wrapper that is also not yet used in any previously existing code.
I am currently away for also the next week, but I will take a look at this once I am back again.

mi-ki:
Cool, thanks in advance!

tobias-rnh:
As a starting point for me, is there any specific example or problem you looked at where these taclets were not applicable anymore, but they were previously?

wadoon:
As a use case, we had for a long time the verification of the hagrid key-server. For this, we only used Integer and the symbolical equality (=) in JavaDL b/c the map theory was restricted to that. 

Would it now be possible to extend this example to equalities on String or even `.equals()` with the new theory?

mi-ki:
> As a starting point for me, is there any specific example or problem you looked at where these taclets were not applicable anymore, but they were previously?

I just re-checked the example from two weeks ago, and apparently cannot reproduce the problem, as it seems to work fine. Hence, sorry for the false alarm, and thanks for the effort!

tobias-rnh:
> > As a starting point for me, is there any specific example or problem you looked at where these taclets were not applicable anymore, but they were previously?
> 
> I just re-checked the example from two weeks ago, and apparently cannot reproduce the problem, as it seems to work fine. Hence, sorry for the false alarm, and thanks for the effort!

Well that is nice, no problem.

tobias-rnh:
> As a use case, we had for a long time the verification of the hagrid key-server. For this, we only used Integer and the symbolical equality (=) in JavaDL b/c the map theory was restricted to that.
> 
> Would it now be possible to extend this example to equalities on String or even `.equals()` with the new theory?

If I understand your question correctly, then the answer is no. My changes were just meant to finish hashCodeModProperty and wrap the LinkedHashMap to be used as a more efficient cache with equalsModProperty in the future.
So I have not changed things about the map theory.

PR #3494, opened 2024-07-06
Restoration of the formatter for KeY files

wadoon:
Archeology: In the back old days, we reformatted all our KeY files. 

If you need to restore a branch before that time to bring it on the master, you should also reformat the KeY files before the merging. 

This PR brings the formatter written by Wiesler. 

The original MR was [!614](https://git.key-project.org/key/key/-/merge_requests/614).

Currently unclear, whether we want to bring this on the main.

**Has Sonarqube activated for evaluation reason**

* Missing for main 
  * [x] Test cases

### DIFF
--- a/key.format/src/test/java/CheckAll.java
+++ b/key.format/src/test/java/CheckAll.java
@@ -1,16 +1,18 @@
 /* This file is part of KeY - https://key-project.org
  * KeY is licensed under the GNU General Public License Version 2
  * SPDX-License-Identifier: GPL-2.0-only */
+
+import org.antlr.v4.runtime.CharStreams;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.DynamicTest;
+import org.junit.jupiter.api.TestFactory;
+
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.util.stream.Stream;
 
-import de.uka.ilkd.key.nparser.format.KeyFormatFacade;
-
-import org.junit.jupiter.api.Assertions;
-import org.junit.jupiter.api.DynamicTest;
-import org.junit.jupiter.api.TestFactory;
+import static de.uka.ilkd.key.nparser.format.KeyFileFormatter.format;
 
 /**
  * @author Alexander Weigl
@@ -26,7 +28,11 @@ public class CheckAll {
             var sq = s.toList();
             return sq.stream().filter(it -> it.getFileName().toString().endsWith(".key"))
                     .map(it -> DynamicTest.dynamicTest(it.getFileName().toString(),
-                        () -> Assertions.assertTrue(KeyFormatFacade.checkFile(it))));
+                            () -> {
+                                var formatted = format(CharStreams.fromPath(it));
+                                var content = Files.readString(it);
+                                Assertions.assertEquals(content, formatted);
+                            }));
         }
     }
 }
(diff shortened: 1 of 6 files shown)


mattulbrich:
Achieving auto-formatting for `.key` files would be nice since we have the infrastructure for this ...

mattulbrich (changes requested):
There is no documentation / commenting at all.

wadoon:
@mattulbrich Looking in the test log shows, that not all KeY files in the repo are parsable anymore.

wadoon:
CodeBuff could be interesting here: https://www.antlr.org/papers/codebuff.pdf

mattulbrich:
This PR seems to also touch totally unrelated parts of the code. This is why it should not be merged as  is.
Does it make sense to  extract the actual formatter code and add that to the code base?

wadoon:
> This PR seems to also touch totally unrelated parts of the code. This is why it should not be merged as is. Does it make sense to extract the actual formatter code and add that to the code base?

I only see the necessary changes, and a lot of changes that arise by the fact, that the main is running away (spotless indentation changes, etc.)

key.format/src/test/java/CheckAll.java
```diff
@@ -0,0 +1,27 @@
...
+import org.junit.jupiter.api.TestFactory;
+
+/**
+ * @author Alexander Weigl
+ * @version 1 (14.07.24)
+ */
+public class CheckAll {
+    @TestFactory
+    public Stream<DynamicTest> checkAllKeYFiles() throws IOException {
+        var s = Files.walk(Paths.get(".."));
```
mattulbrich, line 22: Does this check all .key files in and under the key main directory? That would seem too strict.

wadoon: is no limited to the rule base.

wadoon:
Formatter works now, but not on Windows. @mattulbrich

key.core/src/test/java/de/uka/ilkd/key/proof/runallproofs/GenerateUnitTests.java
```diff
@@ -116,12 +111,15 @@ private static void createUnitClass(RunAllProofsTestUnit unit)
         Map<String, String> vars = new TreeMap<>();
         vars.put("className", className);
         vars.put("packageName", packageName);
-        vars.put("baseDirectory", settings.getBaseDirectory().getAbsolutePath());
+        vars.put("baseDirectory", settings.getBaseDirectory().getAbsolutePath()
```
wadoon, line 114: Required for Windows. :(

mattulbrich: :( indeed. Why is this?

mattulbrich:
There is no documentation.

Can we add a README on the subproject toplevel explaining what this is for and what to do, how to configure etc.

mattulbrich:
Should it be part of the reformatting tasks? Would that not touch some gradle files in key.core?

key.core/src/test/java/de/uka/ilkd/key/proof/runallproofs/GenerateUnitTests.java
wadoon: I used `generateRAPUnit` on Windows. The different plattform (and Java NIO) resulted into file paths using `\\` and compile errors the backslashes were not escaped. 

Now, the code escapes backslashes.

wadoon:
> Should it be part of the reformatting tasks? Would that not touch some gradle files in key.core?

I wanted to keep the PR small. 

Normally, you want the formatter in spottless, but it is not obvious how todo it. I also fear the dependency when formatting requires compilation.

PR #3498, opened 2024-07-07
Remove version lock for JDT in spotless

wadoon:
Remove the version lock for JDT in the spotless configurations, b/c the new version can handle all classes w/o error. 

Also, apply spotless to the code basis. It seems, that Java in Javadoc is also formatted.


WolframPfeifer (commented):
I agree that since the eclipse formatter does not crash any more with the newer version, we can remove the version lock. The change with the indentation in a few places is good. However, I don't think we should reformat the code inside `<pre>` tags in JavaDoc. Is there are way to switch that off?

Issue #3456, opened 2024-04-14
Control-C may leave a defunct process


mattulbrich:
## Description

When starting KeY from the command line, pressing ctrl-c prior to the main window's appearance leaves KeY in a defunct state that can only be killed from outside. Pressing ctrl-c again does not help.

## Reproducible

sometimes.

It seems the time when ctrl-c is pressed plays a role. It seems  that KeY must be in a state where the signal is registered, yet the exit action does not have its full effect yet.

### Steps to reproduce 

Start KeY from the command line (e.g. using the shadow jar) and press ctrl-c before the window shows up.

It should terminate, but instead it only prints `WARN  WindowUserInterfaceControl - Caught SIGINT, exiting...`to the console and gets stuck.

### Additional information

The signal handler seems to have been introduced in 2ff6a372056.

---

* Commit: 9cc569ccced37e242b3a85779f2afdc42b0031ca

mattulbrich:
Nice issue number btw :smile:

wadoon:
In #3451, I have already removed this part. This class is part of the module `jdk.unsupported`. I think the authors want to tell us something with this name. 🤔

WolframPfeifer:
Closed by #3499.

PR #3499, opened 2024-07-20
Remove Ctrl+C handling, see #3456

wadoon:
Removes the explicit ctrl+c handling in `WindowsUserInterfaceControl`

See bug report: #3456.


mattulbrich:
The test cases seem to fail (on ubuntu at least ...). @WolframPfeifer and @FliegendeWurst are perhaps better reviewers.

FliegendeWurst (approved):
Test failures are unrelated

WolframPfeifer (approved):
The original PR that introduced this feature (#3249) was meant as an experimental idea to recover from some out-of-memory situations. However, it seems that it does more harm than good (OOM should be quite rare hopefully), so I am ok with just removing the feature to restore default shell behavior.

PR #3506, opened 2024-08-22
repairing heatmap updates for inner nodes

mattulbrich:
#3391 has been merged prematurely. This is an addition addressing a request made there by @WolframPfeifer .

All documentation from #3391 applies to here.

> @mattulbrich Do you have a good idea how to fix the problem with the non-goal views? If so, could you update it, please? If not, feel free to merge. The remaining UI bug is minor.

### DIFF
--- a/key.ui/src/main/java/de/uka/ilkd/key/gui/actions/HeatmapToggleAction.java
+++ b/key.ui/src/main/java/de/uka/ilkd/key/gui/actions/HeatmapToggleAction.java
@@ -3,6 +3,7 @@
  * SPDX-License-Identifier: GPL-2.0-only */
 package de.uka.ilkd.key.gui.actions;
 
+import java.awt.*;
 import java.awt.event.ActionEvent;
 import java.beans.PropertyChangeListener;
 import javax.swing.*;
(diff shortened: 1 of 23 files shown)


key.ui/src/main/java/de/uka/ilkd/key/gui/actions/HeatmapToggleAction.java
```diff
@@ -67,5 +69,11 @@ public void actionPerformed(ActionEvent e) {
         // this updates the heatmap highlights
         mainWindow.getCurrentGoalView().getHighlighter().removeAllHighlights();
         mainWindow.getCurrentGoalView().printSequent();
+        Component component = mainWindow.getMainFrame().getContent();
```
WolframPfeifer, line 72: This should now always update twice, right? See my suggestion ...

key.ui/src/main/java/de/uka/ilkd/key/gui/nodeviews/MainFrame.java
```diff
@@ -113,4 +113,8 @@ public boolean isShowTacletInfo() {
     public void scrollTo(int y) {
         scrollPane.getVerticalScrollBar().setValue(y);
     }
+
+    public Component getContent() {
```
WolframPfeifer, line 117: Component is always a SequentView. My suggestion changes the type to remove all the casts.

WolframPfeifer (approved):
Thanks for this. Somehow I did not see that the SequentView was stored in MainFrame.

The fix is fine. However, I attached a small patch with a refactoring as a suggestion that prevents double updates and some unnecessary casting in MainFrame. I would be happy if you could apply it, I cannot, since this is a PR from your repo.

[suggestion.txt](https://github.com/user-attachments/files/16785638/suggestion.txt)

key.ui/src/main/java/de/uka/ilkd/key/gui/actions/HeatmapToggleAction.java
mattulbrich: Don't know. If it works with your suggestion, then ... yes. I was not sure. Thanks for finding out.

mattulbrich:
checker framework is fixed on other branch.

PR #3507, opened 2024-08-26
Repairing the NonNull checker framework exception

mattulbrich:
## Related Issue

For a while the checker framework raised an exception during CI. This adds type annotations to remove the errors-

## Intended Change

No more failing CI.

I refactored the util classes a bit, but there should be no different behaviour.

## Type of pull request

- Refactoring (behaviour should not change change)

## Ensuring quality
   
- I added new test case(s) for new functionality / existing functionality

See https://eisop.github.io/cf/manual/manual.html#map-key-defaults-lowerbound for details on the fix.

Thanks to @wmdietl  and @flo2702 for fixing this with me.


wadoon:
Still errors in the types, but different ones.

PR #3508, opened 2024-08-30
Fix copyright notice

flo2702:
## Intended Change

Changes the copyright notice in the "about" dialog to read "2001 – 2024" instead of "2023"


flo2702:
@WolframPfeifer As we just discussed. The version number in `build.gradle` should not be changed here, but later when we create the 2.12.3 branch.

WolframPfeifer (approved):
Thanks!

PR #3509, opened 2024-08-30
Fix for visual bug with overlapping/unreadable text in color settings

WolframPfeifer:
This PR fixes a visual bug in the Options dialog when changing the colors.
The bug was originally discovered by @BookWood7th (on Windows, but I think that is irrelevant), who also provided this screenshot that shows the problem:

Screenshot 2024-08-23 142549

## Intended Change
* Fix of the visual bug where during editing of a color text field, the text gets unreadable and weirdly overlaps. The fix was to set the field to transparent. I do not know why it works, but somehow opaque components do not work correctly with background colors that also contain alpha.
* Fix: For the preview icons besides the color (in non-edited cells), the alpha channel was not correctly interpreted: It seems that it was interpreted as if the background was black. When merged with the non-alpha part, it resulted in colors that are usually not shown in the GUI (since the background color is not black there). Now, background is interpreted as white, which is much better in my opinion (compare for instance the first two colors between the screenshots: they represent the same color code!).

New screenshot with the fix:
image

## Type of pull request

- Other: 

## Ensuring quality
    
- [x] I have tested the feature as follows: Manual inspection in GUI.

### DIFF
--- a/key.ui/src/main/java/de/uka/ilkd/key/gui/colors/ColorSettingsProvider.java
+++ b/key.ui/src/main/java/de/uka/ilkd/key/gui/colors/ColorSettingsProvider.java
@@ -75,7 +75,10 @@ public class ColorSettingsProvider extends SimpleSettingsPanel implements Settin
             }
 
             private Icon drawRect(Color c, int size) {
-                BufferedImage bi = new BufferedImage(size, size, BufferedImage.TYPE_INT_RGB);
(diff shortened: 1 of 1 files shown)


WolframPfeifer:
@BookWood7th Could you maybe check whether this fixes the bug also on Windows?

wadoon (approved):

WolframPfeifer:
Thanks for approving. I also got the confirmation by @BookWood7th that it fixes the original bug.

PR #3519, opened 2024-10-21
fixes Location#compareTo(Location)

### LINKED ISSUE #3518: Location#compareTo(Location) breaks with undefined Location
## Description

The Location class may be instantiated without an associated file. This causes fileUri to be null. Due to this Location::compareTo causes a NullPointerException, when used by a Location without an associated file

## Reproducible
 
always

### Steps to reproduce 

1. Create a Location without an associated fileUri (Location.UNDEFINED / any constructor values with fileUri set to null)
2. Create a second Location object
3. call compareTo on the first Location comparing it to the second

Expected to sort by their Position objects. Actually throws NullPointerException.

### Additional information

---

* Commit: 
[8e28439](https://github.com/KeYProject/key/commit/8e284394d2e0d70dc54fe684ec9c707b6f60a7df)

BookWood7th:
## Related Issue

This pull request addresses #3518.

## Intended Change

Adds explicit nullsLast comparators using the natural order of URI and Position when comparing locations.
This avoids a NullPointerException, when fileUri field is null in one of the locations being compared

## Type of pull request

## Ensuring quality

    
- I have tested the feature as follows: KeY test suite

### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/parser/Location.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/parser/Location.java
@@ -94,9 +94,7 @@ public record Location(URI fileUri, Position position) implements Comparable<Loc
 
     @Override
     public int compareTo(@NonNull Location o) {
-        return Comparator
-                .<Location, URI>comparing(l -> l.fileUri, Comparator.nullsLast(Comparator.naturalOrder()))
-                .thenComparing(Location::getPosition, Comparator.nullsLast(Comparator.naturalOrder()))
-                .compare(this, o);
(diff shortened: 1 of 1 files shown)


BookWood7th:
There appears to be a test that fails unexpectedly sometimes
de.uka.ilkd.key.rule.merge.MergeRuleTests#testLoadClosedGcdProofWithMergePointStatements().
I believe @WolframPfeifer and I witnessed this test failing before when conducting windows unit tests, then not failing when running them again.

WolframPfeifer (approved):

PR #3520, opened 2024-10-23
Add conversion rules for Float/Double negation

WolframPfeifer:
It seems that there are two rules missing which convert from javaUnaryMinusFloat/Double to negFloat/negDouble. Without these rules, KeY (and also the SMT float conversion) can not prove very much when a negation occurs, since the function will just be left uninterpreted.
 
This PR just adds the two missing rules.

Caveat: Somebody with more knowledge about floating points should double-check this PR (I think it should be sound but am not 100% sure).

## Type of pull request

-[x] Bug fix (non-breaking change which fixes an issue)
-[x] There are changes to the taclet rule base
- Other: 

## Ensuring quality
  
### DIFF
--- a/key.core/src/main/resources/de/uka/ilkd/key/proof/rules/floatRulesAssumeStrictfp.key
+++ b/key.core/src/main/resources/de/uka/ilkd/key/proof/rules/floatRulesAssumeStrictfp.key
@@ -8,6 +8,11 @@
 }
 
 \rules(programRules:Java, floatRules:assumeStrictfp) {
+    translateJavaUnaryMinusFloat {
+        \find(javaUnaryMinusFloat(f1))
+        \replacewith(negFloat(f1))
+        \heuristics(javaFloatSemantics)
+    };
 
     translateJavaAddFloat {
         \find(javaAddFloat(f1, f2))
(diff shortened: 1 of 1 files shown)


WolframPfeifer:
@JonasKlamroth @samysweb I just created this PR for our small fix.

samysweb:
@mattulbrich so far the `strictfp` mode did not have a rule to convert `-x` (for `float x`) into the SMT solver floating point operation.
This pull request would add these two Taclets for float and double. Do you agree that these rules currently missing is a bug rather than a feature?

mattulbrich (commented):
@samysweb Indeed, this looks like an oversight and the rules appear to be very sensible.

mattulbrich (approved):

samysweb:
@WolframPfeifer is the qodana test meant to fail? If yes, then I think we can merge this now.

mattulbrich:
It is not "meant" to fail, but fails consistently and can be ignored. Go ahead and merge.

wadoon:
> It is not "meant" to fail, but fails consistently and can be ignored. Go ahead and merge.

I thought we had deactivated it, didn't we?

mattulbrich:
> I thought we had deactivated it, didn't we?

Thought so, too. Perhaps accidently readded to main in a merge?
Let's re-deactivate it.

PR #3525, opened 2024-10-30
Fix #3524

### LINKED ISSUE #3524: Soundness bug: Assignable treatment is wrong
## Description

false can be proved exploiting a problem in the faulty assignable treatment of loops.

## Reproducible

always. Example attached

### Steps to reproduce 

Load the attached proof for the follwoing example:

[problem.zproof.gz](https://github.com/user-attachments/files/17568572/problem.zproof.gz)
(github does no longer accept .zproof files, so you have to gunzip this.)

```java
class A {
    
    /*@ normal_behaviour
      @  requires a.length == 2;
      @  ensures \result == 0;
      @ also
      @ normal_behaviour
      @  requires a.length == 2;
      @  ensures \result == 1;
      @*/
    int problem(int[] a) {
        int i = 0;
        a[1] = 0;
        /*@ loop_invariant 0<=i<=1;
          @ loop_invariant i == 0 ==> a[1] == 0;
          @ loop_invariant a[0] == \old(a[0]);
          @ decreases 2-i;
          @ assignable a[i];
          @*/
        while(i < 1) {
            i++;
            a[1] = 1;
        }
        return a[1];
    }

    /*@ normal_behaviour
      @  ensures b;
      @*/
    void proveAnything(boolean b) {
        int r = problem(new int[2]);
        //@ assert r == 0;
        //@ assert r == 1;
    }

    /*@ normal_behaviour
      @  ensures false;
      @*/
    void proveFalse() {
        proveAnything(false);
    }
}
```

### Additional information

The problem is that in the PO for loop preservation the wrong variable value is used in the evaluation of assignable clauses.

---

* Commit: 80779516745d33e7eaa471616bb47a3b68f2fd09

Drodt:
## Related Issue

This pull request addresses #3524.

## Intended Change

For all loopScopeInv rules: Declares for each local variable `i` a new PV `i_before`, stores the value of `i` before the loop in `i_before`, and applies `{i := i_before}` to the frame condition.

## Type of pull request

- There are changes to the taclet rule bas

## Ensuring quality

    
- I have tested the feature as follows: Some local tests, including the example of #3524.

### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/rule/conditions/NewLocalVarsCondition.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/rule/conditions/NewLocalVarsCondition.java
@@ -1,0 +1,66 @@
+/* This file is part of KeY - https://key-project.org
+ * KeY is licensed under the GNU General Public License Version 2
+ * SPDX-License-Identifier: GPL-2.0-only */
+package de.uka.ilkd.key.rule.conditions;
+
+import de.uka.ilkd.key.java.Services;
+import de.uka.ilkd.key.java.Statement;
+import de.uka.ilkd.key.java.abstraction.KeYJavaType;
+import de.uka.ilkd.key.logic.op.LocationVariable;
+import de.uka.ilkd.key.logic.op.SchemaVariable;
+import de.uka.ilkd.key.rule.MatchConditions;
+import de.uka.ilkd.key.rule.VariableCondition;
+import de.uka.ilkd.key.rule.inst.LocationVarMap;
+import de.uka.ilkd.key.rule.inst.SVInstantiations;
+import de.uka.ilkd.key.util.MiscTools;
+
+import org.key_project.logic.SyntaxElement;
+import org.key_project.util.collection.DefaultImmutableMap;
+import org.key_project.util.collection.ImmutableMap;
+
+import org.jspecify.annotations.NonNull;
+
+/**
+ * Collects all local variables written to in {@code bodySV} and instantiates {@code varsSV} with a
+ * map
+ * from these variables to their (newly created, unique) "before" version.
+ */
+public class NewLocalVarsCondition implements VariableCondition {
+    private final SchemaVariable varsSV;
+    private final SchemaVariable bodySV;
+
+    public NewLocalVarsCondition(SchemaVariable vars, SchemaVariable body) {
+        this.varsSV = vars;
+        this.bodySV = body;
+    }
+
+    @Override
+    public MatchConditions check(SchemaVariable var, SyntaxElement instCandidate,
+            MatchConditions matchCond, Services services) {
+        SVInstantiations svInst = matchCond.getInstantiations();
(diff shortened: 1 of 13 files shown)


mattulbrich:
Guys, ... you are quick! :clap: @Drodt @unp1

key.core/src/main/java/de/uka/ilkd/key/rule/conditions/NewLocalVarsCondition.java
```diff
@@ -0,0 +1,66 @@
...
+import org.key_project.logic.SyntaxElement;
+import org.key_project.util.collection.DefaultImmutableMap;
+import org.key_project.util.collection.ImmutableMap;
+
+import org.jspecify.annotations.NonNull;
+
+/**
+ * Collects all local variables written to in {@code bodySV} and instantiates {@code varsSV} with a
+ * map
+ * from these variables to their (newly created, unique) "before" version.
```
mattulbrich, line 26: Can you elaborate a bit more. I did not really understand this.

key.core/src/main/java/de/uka/ilkd/key/rule/metaconstruct/DeclareLocalLoopVars.java
```diff
@@ -0,0 +1,54 @@
...
+import de.uka.ilkd.key.java.declaration.VariableSpecification;
+import de.uka.ilkd.key.java.reference.TypeRef;
+import de.uka.ilkd.key.logic.op.LocationVariable;
+import de.uka.ilkd.key.logic.op.ProgramSV;
+import de.uka.ilkd.key.rule.inst.SVInstantiations;
+
+import org.key_project.util.collection.ImmutableMap;
+
+/**
+ * A transformer that takes a ProgramSV, which must be instantiated with a map from variable to
```
mattulbrich, line 23: Are ProgramSVs not instantiated by ProgramElements? How can that be matched to a map?

mattulbrich (commented):
It seems that the fix requires a larger change on the loop rule. There seem to be new "_Before" variables now. But ... are there not already _Before-Variables available on the sequent?

key.core/src/main/java/de/uka/ilkd/key/rule/metaconstruct/DeclareLocalLoopVars.java
Drodt: Yeah, this is a bit of a hack. It would probably be better to directly create the necessary variable declarations and updates in \newLocalVars().

Drodt:
> It seems that the fix requires a larger change on the loop rule. There seem to be new "_Before" variables now. But ... are there not already _Before-Variables available on the sequent?

Not necessarily for possibly modified local variables wrt. the loop scope rule. The old loop rule did add those variables, the loop scope rule does not.

mattulbrich (approved):
Loooks a lot cleaner without the maps in the schema variable. Thanks

PR #3529, opened 2024-11-23
Removal of Triple, and Quadruple

wadoon:
## Refactoring 

This PR starts the removal of generic data classes in favor of Java records (aka. named tuples). In this PR, only `Triple` and `Quadruple` are removed. 

In general, generic parametric data classes (`Tuple`, `Union`, ...) lead to unreadable, incomprehensible code, e.g., what does the entity `Triple<StatementBlock, URI, Integer>` express? Answer: It is the key in a map to find block contracts. Using the record `BlockContractKey` expresses this better than triple of something. 

Therefore, records give you the chance of defining useful variable names, and documentation,  and also avoid auto-boxing (hence NPE). 

## Intended Change

No use of `Triple<…>` and `Quadruple<…>`.

## Type of pull request

## Ensuring quality
    
- **X** by TEST CASES!!! and compiler


key.core/src/main/java/de/uka/ilkd/key/speclang/jml/translation/JMLSpecFactory.java
```diff
@@ -1233,9 +1231,9 @@ public Contract createJMLDependencyContract(KeYJavaType kjt, LocationVariable ta
         var context = Context.inClass(kjt, false, tb);
 
         // translateToTerm expression
-        Triple<IObserverFunction, Term, Term> dep =
-            new JmlIO(services).context(context).translateDependencyContract(originalDep);
-        return cf.dep(kjt, targetHeap, dep, dep.first.isStatic() ? null : context.selfVar());
+        var dep = new JmlIO(services).context(context).translateDependencyContract(originalDep);
```
WolframPfeifer, line 1234: I think this is a bad use of var ...

WolframPfeifer (changes requested):
Thanks! In my opinion, it is a very good idea to remove these two classes, and use records with good names instead.

I only have some minor remarks:
- In some comments, "first" and "second" have been replaced too aggressively ...
- Some uses of var should be removed in my opinion.
- Some record parameters could use better names instead of "first", "second", ...

PR #3533, opened 2024-12-13
Fixing the broken automode

mattulbrich:
## Intended Change

The auto-mode in KeY does no longer work: It merely loads a file but does not try to prove it.
This PR partially reverts changes introduced in 8b6df72e14ac9ed620961d93db0a779fc65b7a6a.

## Plan

* [x] Fix `ConsoleUserInterfaceControl`

## Type of pull request

## Ensuring quality

- I have tested the feature as follows: Run it on the CLI

It re-establishes CLI behaviour that we used to have in KeY.

### DIFF
--- a/key.ui/src/main/java/de/uka/ilkd/key/ui/ConsoleUserInterfaceControl.java
+++ b/key.ui/src/main/java/de/uka/ilkd/key/ui/ConsoleUserInterfaceControl.java
@@ -148,35 +148,40 @@ public class ConsoleUserInterfaceControl extends AbstractMediatorUserInterfaceCo
                 printResults(openGoals, info, result2);
             }
         } else if (info.getSource() instanceof ProblemLoader) {
-            LOGGER.debug("{}", result2);
-            System.exit(-1);
-        }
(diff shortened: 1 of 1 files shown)


key.ui/src/main/java/de/uka/ilkd/key/ui/ConsoleUserInterfaceControl.java
```diff
@@ -131,52 +131,56 @@ public void taskFinished(TaskFinishedInfo info) {
...
-                LOGGER.debug("", e);
                 System.exit(-1);
             }
-        } else if (macroChosen()) {
-            applyMacro();
-        } else {
-            finish(proof);
+            if (loadOnly || openGoals == 0) {
+                LOGGER.info("Number of open goals after loading: {}", openGoals);
+                System.exit(0);
```
wadoon, line 159: This is normally not good. The `CUIC` is what you use for KeY as a library. `System.exit` is very dangerous here, but not your fault.

wadoon (approved):

PR #3537, opened 2025-01-17
Proposing a more flexible lexing framework

mattulbrich:
## Intended Change

The syntax highlighting framework in the issue dialog makes the cursor jump around a lot. It does not support KeY files for the moment.

## Plan

* [x] Make the "JavaDocument" language-independent
* [x] Implement a corr. Java lexer
* [x] Implement a corr. KeY parser
* [x] Document the changes
* [ ] (optional) Make sure the realtime colour update works with the Java version since that may be configured at runtime

The code is there can be reused from other places.

## Type of pull request

- Refactoring (behaviour should not change or only minimally change). Actually improvements in the error editor windows.
- There are minor improvements in the UI

## Ensuring quality

- I have tested the feature as follows: tried it out.


wadoon:
There was a branch (`weigl/editor` or so), where I had robust editors on the `RSyntaxArea` component. 

@mattulbrich Should/Does the highlighting respect the current JML active (conditional) tags?

wadoon:
One thing I could find here: 

https://git.key-project.org/key/key/-/merge_requests/341

But there should also be an integrative swing editor with support PSDBG.

mattulbrich:
> @mattulbrich Should/Does the highlighting respect the current JML active (conditional) tags?

The highlighting machinery is the one currently used by the component to display java source code in KeY.
I must admit that I do not know. ... Moreover, perhaps syntax highlighting should not depend on that after all, at least debatable.

mattulbrich:
> One thing I could find here: https://git.key-project.org/key/key/-/merge_requests/341
> I also found the integrative swing editor with support PSDBG on branch `weigl/Scripts`.

The editor looks nice from the picture! However, on Friday we were pretty sure that we did *not* want to integrate an editor into the current swing UI. If an editor ... then as a plugin to whatever VSCode or such with decent feedback into the prover UI.

This is really lightweight syntax highlighting for a few bits within KeY.

The current solution lacked highlighting for KeY and made the cursor jump annoyingly in the exception recovery dialog.

wadoon:
> > @mattulbrich Should/Does the highlighting respect the current JML active (conditional) tags?
> 
> The highlighting machinery is the one currently used by the component to display java source code in KeY. I must admit that I do not know. ... Moreover, perhaps syntax highlighting should not depend on that after all, at least debatable.

If it is inside of the KeY. I see the potential for recognizing errors quickly. For standalone editors, I would say it should be configurable.

mattulbrich:
> @mattulbrich Should/Does the highlighting respect the current JML active (conditional) tags?

After checking a bit more, it turns out that it did already support this before the change.

key.ui/src/main/java/de/uka/ilkd/key/gui/sourceview/JavaJMLEditorLexer.java
```diff
@@ -0,0 +1,544 @@
...
+        StyleConstants.setForeground(keyword, JAVA_KEYWORD_COLOR.get());
+        StyleConstants.setForeground(comment, COMMENT_COLOR.get());
+        StyleConstants.setForeground(javadoc, JAVADOC_COLOR.get());
+        // StyleConstants.setForeground(string, JAVA_STRING_COLOR);
+        StyleConstants.setForeground(jml, JML_COLOR.get());
+        StyleConstants.setForeground(jmlkeyword, JML_KEYWORD_COLOR.get());
+        StyleConstants.setBold(jmlkeyword, true);
+    }
+
+    private void checkAt() {
```
wadoon, line 337: Please use `JMLMarkerDecision` here for a unified experience.

key.ui/src/main/java/de/uka/ilkd/key/gui/sourceview/JMLEditorLexer.java
```diff
@@ -0,0 +1,160 @@
+/* This file is part of KeY - https://key-project.org
```
wadoon, line 1: Remove this file.

wadoon (changes requested):

key.ui/src/main/java/de/uka/ilkd/key/gui/sourceview/JMLEditorLexer.java
mattulbrich: done

PR #3541, opened 2025-02-01
Make the proof tree (or the KeY GUI in general) responsive during automatic proof search

### LINKED ISSUE #3415: Make the proof tree (or the KeY GUI in general) responsive during automatic proof search

At the moment, during the automatic proof search the GUI of KeY "freezes", i.e. does not respond to input events except for a click on the stop button. However, depending on the step limit the automatic proof search might take quite a long time, in which the user can not do anything. It would be a very nice feature to be able to browse the proof (and maybe also other parts of the interface, for example the source code). While the proof search is ongoing, KeY should present the proof state from before the automatic search was started. 

## Additional context

Note that this feature request is probably not that easy to implement: The current solution for "freezing" the UI (except from the stop button) is to lay an invisible panel over the KeY GUI that blocks all clicks and input events. For this feature request, a different and more sophisticated solution will be needed.

wadoon:
Fixes #3415. 

State: 

* [x] Glasspane removed
* [x] `KeyAction` classes extended for enabledness updates
* [x] Look through actions and decide whether they are safe or not to be executed during auto-mode.

### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/pp/SequentPrintFilter.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/pp/SequentPrintFilter.java
@@ -82,6 +82,10 @@ public abstract class SequentPrintFilter {
      * entries.
      */
     protected void filterIdentity() {
+        if(originalSequent==null) {
+            return;
+        }
+
         antec = ImmutableSLList.nil();
         Iterator<SequentFormula> it = originalSequent.antecedent().iterator();
         while (it.hasNext()) {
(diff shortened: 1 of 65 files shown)


wadoon:
### State of this PR: 

In this PR the `GlassPane` blocking the user input was removed. Instead of this overall deactivation, each action must be de-/activated. A new facility was introduced to `KeyAction`. `KeyAction` holds a predicate function `() -> Boolean` which defines its enability. `KeyAction#updateEnabledness` updates `Action#setEnable` accordingly to this function. This method should be used in listeners (property change listeners, etc...) to update the value. 

`MainWindowAction` exploits this facility with two methods, which setup the listener and predicate function accordingly: 

* `#enabledOnAnActiveProof();`: Action is enabled if a proof is selected
* `#enabledWhenNotInAutoMode();`: Action is not enabled if the main window is in "auto mode". 

Note, that this PR requires manual testing and clicking within the GUI.

PR #3549, opened 2025-02-15
Cleanup: Removal of the write and read settings using Properties

wadoon:
This PR removes the deprecated methods for `readSettings(Properties)` and `writeSettings(Properties)`. 

**This will break old KeY files.**

Consequences are a lot of dead code, and some simplifications in the UI. 

"Show all Settings" is now tree-based instead of tables. 

image


wadoon:
Question: Could it be that proof settings can changed by the proof replayer?

wadoon:
#3974

PR #3551, opened 2025-02-15
Cleanup: Use Standard Tech for Test Fixtures

wadoon:
This PR removes the `LineProperties.java` in favor of YAML test fixture. 

Pros: 

- Support by JUnit `YamlSource` in parameterized test
- Support for YAML in browser
- Flexibility and type safety

This PR brings two dependencies for tests: 

- AssertJ for writing simpler assertion in test with better messages.
- Jackson for reading YAML


mattulbrich:
Unfortunately, this PR does not follow the PR summary template of the project. Can you please provide the information of the template?

mattulbrich:
Thanks for replacing the parser. Actually, I think it is more understanable and clearer with the test cases in individual files. Can that be arranged?

wadoon:
> Thanks for replacing the parser. Actually, I think it is more understanable and clearer with the test cases in individual files. Can that be arranged?

done

mattulbrich:
> > Thanks for replacing the parser. Actually, I think it is more understanable and clearer with the test cases in individual files. Can that be arranged?
> 
> done

thanks. Is it on purpose that the files are in a different directory now?

wadoon:
> thanks. Is it on purpose that the files are in a different directory now?

Avoidance of one more folder level in the hierarchy. It seemed to be unnecessary for script fixtures.

mattulbrich:
I have little experience with yml, and, hence, it seems to me that this change actually adds complexity rather than reducing it. I cannot appreciate the improvements that this PR bring. Perhaps someone more familiar w/ the language can see this and do the review.

Drodt:
Yaml is a pretty common format. I like this change; tests are more compact and readable.

Drodt (approved):

PR #3552, opened 2025-02-15
Cleanup: Making the build script fit for the future.

wadoon:
The best practices in Gradle have been moved further. This PR tries to follow. 

It introduced: 

* One common `libs.versions.toml` file for a global handling of dependencies. 
  This dependency management tries to avoid using the same Maven coordinates in various places. 
  A different solution would be using a BOM. 

* It introduces `java-convention` plugin, which assimilates the old `subproject { ... }` configuration

* Build scripts are transferred to kts (Kotlin) for more type safety.


wadoon:
I guess in the meantime, we would rather do it again using Agents than start a merge now.

PR #3556, opened 2025-02-19
repair the axiom generation for adts defined in JavaDL

mattulbrich:
## Related Issue

This pull request addresses #3427, but does not solve it.

## Intended Change

The axiom taclet created for adt was not applicable. It is applicable now.

## Plan

* [x] remove the "notFreeIn" clause from the generated taclet
* [x] remove the find clause from the taclet.

## Type of pull request

## Ensuring quality

- I have tested the feature as follows: Looked at the translation of a few examples

It is a small change in the creation of a taclet.

### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/nparser/builder/TacletPBuilder.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/nparser/builder/TacletPBuilder.java
@@ -371,9 +371,9 @@ public class TacletPBuilder extends ExpressionBuilder {
     }
 
 
-    private RewriteTacletBuilder<RewriteTaclet> createAxiomTaclet(
+    private TacletBuilder<NoFindTaclet> createAxiomTaclet(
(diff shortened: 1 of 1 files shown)


mattulbrich:
Addressed during the 3rd HacKeYthon.

key.core/src/main/java/de/uka/ilkd/key/nparser/builder/TacletPBuilder.java
```diff
@@ -371,9 +371,9 @@ private TacletGoalTemplate createGoalDtConstructor(KeYParser.Datatype_constructo
     }
 
 
-    private RewriteTacletBuilder<RewriteTaclet> createAxiomTaclet(
+    private TacletBuilder createAxiomTaclet(
```
wadoon: raw usage of parameterized class

wadoon (approved):

Issue #3563, opened 2025-02-20
Automatic solver fail to proceed with two equal set singletons


frereit:
## Description

The default solve strategy does not automatically apply "equalityToElementOf". This prevents it from automatically proving a simple case where two set singletons are equal.

## Reproducible
  
always

### Steps to reproduce 

Attempt to prove the following .key file automatically:

```
\functions {
    java.lang.Object o1;
    java.lang.Object o2;
    Field f1;
    Field f2;
}


\problem {
    singleton(o2, f2) = singleton(o1, f1)
    ==>
    o2 = o1
}
```

This does not work. The automatic solver does not make any progress. After manually applying the "equalityToElementOf" rule, the automatic prover is able to prove this successfully.

---

* Commit: 51cf67e4591ca205fde0893f62a1d757b41a5114

PR #3565, opened 2025-02-20
Add a stricter dependency contract proof obligation

frereit:
## Intended Change

The current proof obligation (`DependencyContractPO`) only verifies that the return value of a method only depends on the locations in the accessible clause. In contrast, this new proof obligation additionally verifies that any objects on the heap (except those newly created) only depend on the locations in the accessible clause.

The formula that is generated by this proof obligation for a method `m` with contract $(\mathit{pre}, \mathit{post}, \mathit{mod}, \mathit{dep})$ is roughly

$$\mathit{pre} \wedge \mathit{freePre} \wedge \langle \mathtt{res = m(args...)@C; } \rangle (h_1 = \mathtt{heap} \wedge r_1 = \mathtt{res}) \wedge \\{u\\} \langle \mathtt{res = m(args...)@C; } \rangle (h_2 = \mathtt{heap} \wedge r_2 = \mathtt{res}) \rightarrow \mathit{mod} \subseteq \mathit{dep} \wedge r_1 = r_2 \wedge \forall o; \forall f; ((o,f) \in \mathit{mod} \rightarrow o.f@h_1 = o.f@h_2) $$

While testing, you may run into #3563, so beware.

A simple example which proves succesfully with the current proof obligation, but cannot be proven with this new one is as follows:

```java
/*@ public normal_behavior
@ accessible x;
@ assignable x;
@ ensures \result == x; 
@*/
public int foo() {
  if (this.y > 0) {
    this.x += 1;
    return x - 1;
  }
  return this.x;
}
```

## Plan

* [ ] Code cleanup
* [ ] Document the changes
* [ ] Add tests
* [ ] Discussion
* [ ] UI changes?

## Type of pull request

## Ensuring quality

    
### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/proof/init/DependencyContractPO.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/proof/init/DependencyContractPO.java
@@ -23,14 +23,14 @@ import org.key_project.util.collection.ImmutableList;
 /**
  * The proof obligation for dependency contracts.
  */
-public final class DependencyContractPO extends AbstractPO implements ContractPO {
+public class DependencyContractPO extends AbstractPO implements ContractPO {
 
-    private Term mbyAtPre;
+    protected Term mbyAtPre;
 
-    private final DependencyContract contract;
+    protected final DependencyContract contract;
 
-    private InitConfig proofConfig;
-    private TermBuilder tb;
+    protected InitConfig proofConfig;
+    protected TermBuilder tb;
 
 
     // -------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


Drodt:
This PR is the result of the 2025 HacKeYthon

mattulbrich:
Thanks, guys! Looks cool.

Contracts are used to conservatively abstract from method invocations. The (old) dependency contracts are used to show that the value of calls to pure methods does not change since the dependency set remains untouched.

Is there a similar use case for the stronger dependency contracts that exploit the proofs? The method-invocation rule would probably have to ensure that the read variables of the called method are a subset of the accessible locations in the calling method. ...

mattulbrich:
Is planned to have a taclet option to switch between the old and new PO?

Drodt:
> Thanks, guys! Looks cool.
> 
> Contracts are used to conservatively abstract from method invocations. The (old) dependency contracts are used to show that the value of calls to pure methods does not change since the dependency set remains untouched.
> 
> Is there a similar use case for the stronger dependency contracts that exploit the proofs? The method-invocation rule would probably have to ensure that the read variables of the called method are a subset of the accessible locations in the calling method. ...

The motivation (for me) was first to enable dependency contracts for void methods and second to see if it could be done. We did briefly consider if it would be useful but I think we need more experimentation for a proper use case.

For that reason, a taclet option is not planned. (And we should not make it a taclet option as it does not affect the semantics of the Kripke structure.)

PR #3569, opened 2025-02-23
Use a custom Github action for setting up the SMT solvers

wadoon:
## Intended Change

Exchange of the `dlsmt.sh` scripts in favor of a real Github action that uses Github's API for setting up the SMT solvers. 

This action is self-written and available here: https://github.com/KeYProject/setup-smt

*Benefit:* The pipeline becomes OS-independent and now we can run macOS and windows pipelines with SMT  solvers. This PR also enables SMT solvers on windows *unit-test* pipelines!

On the road: Fixes for double artifact names on 8our broad release tests](https://github.com/KeYProject/key/actions/workflows/tests_winmac.yml).

## Type of pull request

- **No code changed.**
- There are changes to the deployment/CI infrastructure (github)
  - [board release tests needs to be triggered manually](https://github.com/KeYProject/key/actions/runs/13485598748) [Broad Release Tests](https://github.com/KeYProject/key/actions/workflows/tests_winmac.yml)
     
    **broad release test** is currently broken as it uses `--release 21` on a JRE 17. 

## Ensuring quality
- I have tested the feature as follows:  CI 

----- 

I currently have only added z3 and CVC5 to `setup-smt` as these were also in `dlsmt.sh`. Looking at [`java-smt3`](https://github.com/sosy-lab/java-smt) the following exists:

* [Bitwuzla](https://bitwuzla.github.io/) 
* [Boolector](https://boolector.github.io/) 
* [CVC4](https://cvc4.github.io/) 
* [CVC5](https://cvc5.github.io/) 
* [MathSAT5](http://mathsat.fbk.eu/)
* [OpenSMT](https://verify.inf.usi.ch/opensmt) 
* [OptiMathSAT](http://optimathsat.disi.unitn.it/) 
* [Princess](http://www.philipp.ruemmer.org/princess.shtml) 
* [SMTInterpol](https://ultimate.informatik.uni-freiburg.de/smtinterpol/) 
* [Yices2](https://yices.csl.sri.com/) 
* [Z3](https://github.com/Z3Prover/z3) 

Please add a PR at [keyproject/setup-smt](/keyproject/setup-smt) if some of these are required.


wadoon:
From KaKeY: 

- `#testDiv5` and '#testDiv6' are deactivated as their runtime hit the timeout (total of 10 min. runtime). They are probably the cause of the unexpected shutdown problems on Linux runners.

Issues: #3575

wadoon:
@mattulbrich @WolframPfeifer waiting for review ...

.github/workflows/tests.yml
```diff
@@ -30,6 +30,18 @@ jobs:
       GH_TOKEN: ${{ github.token }}
     steps:
       - uses: actions/checkout@v4
+
+      - name: Setup SMT solvers
+        uses: keyproject/setup-smt@v0
+        with:
+          z3Version: 4.13.0
```
mattulbrich, line 37: Why is the version fixed here, but not below? According to the setup-smt, this only works with 4.14.0
Should CVC5 not also be fixed?
If this is because it would fail otherwise, pls add a comment -- and it should be investigated.

mattulbrich (approved):
Please check the version there, add a comment if appropriate or change the line.

.github/workflows/tests.yml
wadoon: Unset the version. Maybe `4.14.0` does the job now after disabling the two JUnit tests.

PR #3579, opened 2025-03-05
Improve JavaDocs for Qualifier, SyntaxElement

Drodt:
## Intended Change

Improve some JavaDocs in `Qualifier` and `SyntaxElement`. Also fixes a minor issue in `SortDependingFunction`, where we compared objects of disjoint types, leading to inefficient comparisons.

Thanks to @mattulbrich for pointing it out!

## Type of pull request

## Ensuring quality

    
### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/logic/op/Qualifier.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/logic/op/Qualifier.java
@@ -7,6 +7,20 @@ import java.util.WeakHashMap;
 
 import org.key_project.logic.TerminalSyntaxElement;
 
+/**
+ * This class is primarily used in logic operations where qualifiers (e.g., types or sorts)
+ * are referenced multiple times, such as in {@link ObserverFunction} and
+ * {@link SortDependingFunction}.
+ * It helps to ensure that qualifiers maintain identity-based equality when reused.
+ * <p>
+ * As a qualifier is part of the structure of a symbol, e.g., function symbols, it has to be its own
+ * AST node. Hence, it is part of the structure reached by navigating through
+ * {@link org.key_project.logic.SyntaxElement} and
+ * {@link org.key_project.logic.SyntaxElementCursor}.
+ * </p>
+ *
+ * @param <T> The type of the qualifier object being wrapped.
+ */
 public class Qualifier<T> implements TerminalSyntaxElement {
     private final T qualifier;
 
@@ -16,6 +30,13 @@ public class Qualifier<T> implements TerminalSyntaxElement {
         this.qualifier = qualifier;
     }
 
+    /**
+     * Get the instance for this qualifier. If none exist yet, create one.
+     *
+     * @param qualifier the qualifier for which we create an instance
+     * @return new instance
+     * @param <T> The type of the qualifier object being wrapped.
+     */
     synchronized static <T> Qualifier<T> create(T qualifier) {
         if (INSTANCES.containsKey(qualifier)) {
             return (Qualifier<T>) INSTANCES.get(qualifier);
--- a/key.core/src/main/java/de/uka/ilkd/key/logic/op/SortDependingFunction.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/logic/op/SortDependingFunction.java
@@ -102,7 +102,7 @@ public final class SortDependingFunction extends JFunction {
      * @return the variant for the given sort
      */
     public synchronized SortDependingFunction getInstanceFor(Sort sort, TermServices services) {
-        if (sort == this.sortDependingOn) {
+        if (sort == this.sortDependingOn.getQualifier()) {
             return this;
         }
 
--- a/key.ncore/src/main/java/org/key_project/logic/SyntaxElement.java
+++ b/key.ncore/src/main/java/org/key_project/logic/SyntaxElement.java
@@ -4,8 +4,14 @@
 package org.key_project.logic;
 
 /**
- * This interface declares the methods common to all logic related (terms, formulas, programs)
- * AST nodes.
+ * This interface declares the methods common to all logic related (terms, formulas, and programs)
+ * AST nodes. All tree-like structures should implement this interface.
+ * <p>
+ * For navigating the syntax elements, see {@link SyntaxElementCursor}.
+ * </p>
+ * <p>
+ * If a class has no children, it should implement {@link TerminalSyntaxElement}.
+ * </p>
  */
 public interface SyntaxElement {
     /**
@@ -23,6 +29,9 @@ public interface SyntaxElement {
      */
     int getChildCount();
 
+    /**
+     * @return a new cursor for the subtree with the current node as root.
+     */
     default SyntaxElementCursor getCursor() {
         return new SyntaxElementCursor(this);
     }


key.core/src/main/java/de/uka/ilkd/key/logic/op/Qualifier.java
```diff
@@ -7,6 +7,20 @@
...
+ * As a qualifier is part of the structure of a symbol, e.g., function symbols, it has to be its own
+ * AST node. Hence, it is part of the structure reached by navigating through
+ * {@link org.key_project.logic.SyntaxElement} and
+ * {@link org.key_project.logic.SyntaxElementCursor}.
+ * </p>
+ *
+ * @param <T> The type of the qualifier object being wrapped.
+ */
 public class Qualifier<T> implements TerminalSyntaxElement {
     private final T qualifier;
```
mattulbrich, line 25: Why does a class called Qualifier have a field called qualifier? Is this not self-referential?

key.core/src/main/java/de/uka/ilkd/key/logic/op/Qualifier.java
```diff
@@ -7,6 +7,20 @@
 
 import org.key_project.logic.TerminalSyntaxElement;
 
+/**
+ * This class is primarily used in logic operations where qualifiers (e.g., types or sorts)
+ * are referenced multiple times, such as in {@link ObserverFunction} and
+ * {@link SortDependingFunction}.
+ * It helps to ensure that qualifiers maintain identity-based equality when reused.
+ * <p>
```
mattulbrich, line 15: I must admit that I could not follow this explanation.

It seems that this class is used to make sure that sort annotations are used with canonical representatives such that == can be used, not equals. But sorts are are already organised this way.

Can you perhaps prefix what this class stands for before mentioning its main use case?

key.core/src/main/java/de/uka/ilkd/key/logic/op/SortDependingFunction.java
```diff
@@ -102,7 +102,7 @@ public static SortDependingFunction getFirstInstance(Name kind, TermServices ser
      * @return the variant for the given sort
      */
     public synchronized SortDependingFunction getInstanceFor(Sort sort, TermServices services) {
-        if (sort == this.sortDependingOn) {
+        if (sort == this.sortDependingOn.getQualifier()) {
```
mattulbrich, line 105: Why Qualifier here? It feels oddly out of place ...

key.ncore/src/main/java/org/key_project/logic/SyntaxElement.java
```diff
@@ -4,8 +4,14 @@
 package org.key_project.logic;
 
 /**
- * This interface declares the methods common to all logic related (terms, formulas, programs)
- * AST nodes.
+ * This interface declares the methods common to all logic related (terms, formulas, and programs)
```
mattulbrich, line 7:
```suggestion
 *  This interface defines the base class for abstract syntax tree structures regardless of the underlying language.
 *
 *  It declares the methods common to all logic related (terms, formulas, and programs)
```

key.core/src/main/java/de/uka/ilkd/key/logic/op/Qualifier.java
```diff
@@ -16,6 +30,13 @@ private Qualifier(T qualifier) {
...
     }
 
+    /**
+     * Get the instance for this qualifier. If none exist yet, create one.
+     *
+     * @param qualifier the qualifier for which we create an instance
+     * @return new instance
+     * @param <T> The type of the qualifier object being wrapped.
+     */
     synchronized static <T> Qualifier<T> create(T qualifier) {
```
mattulbrich, line 40: Why is this called create? It does not necessarily create a new object. `get` would be a more fitting name.
Again you create a qualifier from a qualifier, how can that be?

key.core/src/main/java/de/uka/ilkd/key/logic/op/Qualifier.java
mattulbrich: Is that perhaps rather a QualifierReference than a Qualifier, like a type reference is one concrete mentioning of a type. Usually references are not identity comparable. So why is it important that references to qualifiers are kept unique?

mattulbrich (changes requested):
The qualifier class still leaves some question marks as to its nature.

key.core/src/main/java/de/uka/ilkd/key/logic/op/Qualifier.java
wadoon: I would not leave the param `T` unbound. 

As an implementor, I would have questions on the identity and shareability `T`-instances. 

@mattulbrich: Coming from jmlparer, it is normal to have an `Modifier` ast class, which also carries the positional information, which contains a reference to an enum defining the set of possible identifiers.

Drodt: > Why does a class called Qualifier have a field called qualifier? Is this not self-referential?

I renamed the class to make the distinction clearer.

Drodt: > I would not leave the param `T` unbound.

What would you choose as a bound? From the wrapper's POV, T needs no restrictions.

> As an implementor, I would have questions on the identity and shareability `T`-instances.

Noted.

key.core/src/main/java/de/uka/ilkd/key/logic/op/Qualifier.java
Drodt: Changed to `get`

key.core/src/main/java/de/uka/ilkd/key/logic/op/Qualifier.java
Drodt: Changed

key.core/src/main/java/de/uka/ilkd/key/logic/op/SortDependingFunction.java
mattulbrich: @wadoon: I do not think type annotations are the point here.

mattulbrich: @Drodt I wonder: If we can always easily obtain the unique sort wrapper object from a sort, would it not be clearer to use Sort here instead of Qualifier<Sort> since this is on the more concrete level where the AST abstraction is not so present.

An Operator, e.g., also has a result sort, not a qualifier.

Why is it wrapped here?

Drodt: > @Drodt I wonder: If we can always easily obtain the unique sort wrapper object from a sort, would it not be clearer to use Sort here instead of Qualifier since this is on the more concrete level where the AST abstraction is not so present.
> 
> An Operator, e.g., also has a result sort, not a qualifier.
> 
> Why is it wrapped here?

That's certainly possible. But then, every sort would be a SyntaxElement and we would traverse over them, which we haven't done before. Typically, we skipped over all type information, unless it was part of the syntax.

That's also a part of it: Typically, a sort is not explicitly part of the syntax---unless it's part of a sort depending function.

mattulbrich: I had the diff I sent you still in my sources so I quickly committed it.

mattulbrich (approved):

PR #3587, opened 2025-03-29
AST for KeY Scripts

wadoon:
## Synopsis

JMLScripts and KeY scripts use the same backend, but the ProofScriptEngine is currently bound to the KeY language's parser. This PR introduces a simple AST class to abstract the source of a script, which is a simple record containing the command name, positional and keyword arguments, sub-commands, and positional information. 

```mermaid
classDiagram
  direction TB

  class ScriptCommandAst { 
  name : String 
  positionalArgs : List<Object>
  keywordArgs    : Map<String,Object>
  loc            : Location
 }

ScriptCommandAst "*" <-- ScriptCommandAst : sub-commands
```

Along the road, the interface of script commands is simplified: There is no need for preparing the parameter class instance anymore. Instead, `execute(..)` is called directly, and any command implementation can deal with the AST as they like. 

```mermaid
classDiagram 
  direction TB

   class ProofScriptCommand {
    execute(uiControl, cmdAst, engineState) 

    getArguments() : List<ProofScriptArgument>
    getName() : String
    getAlias() : List<String>
    getDocumentation()
  }
```

Also, proof script commands can announce a list of aliases. This allows the implementation of multiple (similar) commands with one implementation. The implementation can adapt its behavior using the given AST representation. For example, the `let` command is also callable via `letf`, the latter forces/overwrites the abbreviation. 

Using `List<Object>` and `Map<String,Object>` for the representation inside of the AST allows the use within JML scripts, as custom objects can be provided within the AST. Later, the `ValueInjector` should know the conversion rules to get from the custom object to the required types, e.g., how to come from a `JMLTerm` to a `JTerm`.

Builds upon: #3021

## Intended Change

`ProofScriptEngine` is free of ANTLR classes. 
An AST for the execution of proof scripts exists, which is useable

## Plan

* [x] Do it.
* [x] Test it.
* [x] Ship it.

## Type of pull request

  - **Breaking change**: 
    breaks the internal API of proof script execution, and script commands.

## Ensuring quality


build.gradle
```diff
@@ -101,6 +101,7 @@ subprojects {
         testImplementation ("org.junit.jupiter:junit-jupiter-params")
         testRuntimeOnly ("org.junit.jupiter:junit-jupiter-engine")
         testRuntimeOnly ("org.junit.platform:junit-platform-launcher")
+        testImplementation ("org.assertj:assertj-core:3.26.3")
```
mattulbrich, line 104: Approved added dependency:
1) It is a dependency for testcases only and does increase runtime dependencies
2) It adds clarity to test cases

key.core/src/main/java/de/uka/ilkd/key/scripts/ScriptCommandAst.java
```diff
@@ -0,0 +1,73 @@
...
+                + positionalArgs.stream().map(ScriptCommandAst::humanString).collect(joining(" "))
+                + (commands != null
+                        ? " {"
+                            + commands.stream().map(ScriptCommandAst::asCommandLine)
+                                    .collect(joining("\n"))
+                            + "\n}"
+                        : ";");
+    }
+
+    public static String humanString(Object value) {
```
mattulbrich, line 67: What does `human string` mean? Is this perhaps better named `toReadableString`? Making this receive an Object seems very general. Could you add some doc when this public method is best used?

key.core/src/main/java/de/uka/ilkd/key/scripts/ScriptCommandAst.java
```diff
@@ -0,0 +1,73 @@
...
+        List<Object> positionalArgs,
+        @Nullable List<ScriptCommandAst> commands,
+        @Nullable Location location) {
+
+    public ScriptCommandAst(String commandName, Map<String, Object> namedArgs,
+            List<Object> positionalArgs) {
+        this(commandName, namedArgs, positionalArgs, Collections.emptyList(), null);
+    }
+
+    public String asCommandLine() {
```
mattulbrich, line 52: What does `command line` mean? We do not have a CLI. Should that produce a string that would be parsed to the same (or an somehow equivalent) AST? would `toScriptCode` or `toScriptLine` or so be a better name?

mattulbrich: I think this deserves a short doc.

key.core/src/main/java/de/uka/ilkd/key/scripts/ScriptCommandAst.java
```diff
@@ -0,0 +1,73 @@
...
+        return commandName + ' ' +
+                namedArgs.entrySet().stream()
+                        .map(it -> it.getKey() + ": " + humanString(it.getValue()))
```
mattulbrich, line 55: In line 26 the keys and values are separated by a "=", here by a ":". One seems to be wrong.

mattulbrich: Checking the grammars: `(COLON|EQUALS)`. But better be consistent within one file ...

key.core/src/main/java/de/uka/ilkd/key/scripts/ScriptCommandAst.java
```diff
@@ -0,0 +1,73 @@
...
+/// @param commands a nullable block of proof script arguments (represents "higher-order proof
+/// scripts").
+/// If null, the block was omitted syntactically in contrast to an empty list.
+/// @param location the location of this command for error reporting.
+/// @author Alexander Weigl
+/// @version 1 (14.03.25)
+@NullMarked
+public record ScriptCommandAst(
+        String commandName,
+        Map<String, Object> namedArgs,
```
mattulbrich, line 42: I do not feel so well with only having Object here. There feels to be more type safety achievable ...
But perhaps that would be overengineered. ...

key.core/src/main/java/de/uka/ilkd/key/scripts/ScriptCommandAst.java
```diff
@@ -0,0 +1,73 @@
...
+/// ```
+///
+/// @param commandName the name of the command, e.g., "macro" for `macro auto;`
+/// @param namedArgs a map of the given named arguments and values.
+/// If a named argument is not given, the entry should be missing in the map. Null-values are not
+/// allowed.
+/// @param positionalArgs the list of given positional arguments
+/// @param commands a nullable block of proof script arguments (represents "higher-order proof
+/// scripts").
+/// If null, the block was omitted syntactically in contrast to an empty list.
```
mattulbrich, line 35: Important point!

key.core/src/main/java/de/uka/ilkd/key/scripts/ScriptCommandAst.java
```diff
@@ -0,0 +1,73 @@
...
+/// ```
+/// <commandName> key_1=value_1 ... key_m=value_m positionalArgs_1 ... positionalArgs_n {
+/// commands_0; ...; commands_k;
+/// }
+/// ```
+///
+/// @param commandName the name of the command, e.g., "macro" for `macro auto;`
+/// @param namedArgs a map of the given named arguments and values.
+/// If a named argument is not given, the entry should be missing in the map. Null-values are not
+/// allowed.
```
mattulbrich, line 31: Could one write `Map<String, @NonNull Object>` to emphasise this point? Still worth mentioning here.

key.core/src/main/java/de/uka/ilkd/key/nparser/KeyAst.java
mattulbrich, line 1: It's not changed by this PR, but the comment in (now) line 41 seems to be wrong:
 To obtain an KeYAst use the {@link ParsingFacade#getParseRuleContext(KeyAst)}.

It is more likely: To obtain an KeYAst use the methods in the facade and to obtain the raw ANTLR parse tree use the above method. Right?

key.core/tacletProofs/seqPerm2/Taclet_schiffl_thm_1.proof
mattulbrich, line 1: Why change this? Is ":" the new symbol to be prefered? Seems reasonable if there are no more " around arguments

mattulbrich (approved):
Nice.

Thanks for the comments. Appreciate that.
Made a few suggestions. Mostly on naming and documentation.

Have not actually run it, but since we plan to extend this to JML scripts, it will be tested thoroughly soon.

key.core/src/main/java/de/uka/ilkd/key/scripts/ScriptCommandAst.java
mattulbrich: Since it is used in the above method, it seems to actually produce a parsable string, too?!

wadoon: The existence is to translate a `ParseRuleContext` into something readable.

key.core/src/main/java/de/uka/ilkd/key/scripts/ScriptCommandAst.java
wadoon: The type safety is ensured on executing of the command. 

For backwards compat., you are allowed to use a String also if an Integer is expected in the command. 

Another solution would to define an interface for values of AST:

```mermaid
classDiagram
    ScriptValue <|-- SVInt
    ScriptValue <|-- SVDouble
    ScriptValue <|-- SVTerm
    ScriptValue <|-- SVSequent
    ScriptValue <|-- SVBool
    ScriptValue <|-- SVRule
    ScriptValue : +intValue()
    ScriptValue : +strValue()
    ScriptValue : +termValue()
    ScriptValue : +sequentValue()
    ScriptValue : +ruleValue()
    ScriptValue : +boolValue()
```

key.core/src/main/java/de/uka/ilkd/key/scripts/ScriptCommandAst.java
wadoon: But the class is annotated with `@NullMarked`, hence this would be redundant.

key.core/src/main/java/de/uka/ilkd/key/nparser/KeyAst.java
wadoon: IMHO, the javadoc captures my intention. Sometimes you need the `ParseRuleContext` (e.g., reformatting), but you should be careful to use such methods. You should avoid using parse trees or ANTLR classes directly (to keep a separation of concern).

PR #3590, opened 2025-04-06
Rewriter of old to new settings format.

wadoon:
This PR brings a tool for rewriting the settings in KeY files to the new JSON-like format. A requirement to get rid of the old property-based interface. 

Changes along the road:

* The KeYLexer can be configured whether `\proof` is also treated as `EOF` or not. In normal mode, the lexer stops at `\proof` so that the s-expr of the proof are not going through the parser. 

## Plan 
* [x] Rewrite all KeY files. This happens in separate PR #3688

## Type of pull request

  * Lexer became the possibility to *not stop at `\proof`*.

## Ensuring quality

- Before overwriting the file, the reformatter checks the parseability. 
  The rewriting itself is done via the settings classes (load/store methods), as we are not morons, this should ensure settings equality.

### DIFF
--- a/key.core/src/main/antlr4/KeYLexer.g4
+++ b/key.core/src/main/antlr4/KeYLexer.g4
@@ -40,6 +40,10 @@ lexer grammar KeYLexer;
    }
 
     private Token tokenBackStorage = null;
+    private boolean proofIsEOF = true;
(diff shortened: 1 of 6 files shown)


mattulbrich:
*Agreed during meeting:* Move to `scripts` directory. Reminder: When removing from the parser, make sure to give a descriptive error message.

PR #3593, opened 2025-04-11
Fixed obtaining results from CVC4 and cvc5

WolframPfeifer:
This small fix removes the delimiters in CVC4.props and also sets CVC4 and also cvc5 to `--no-interactive`, which fixes the (at least most?) crashes with the `Stream closed` error message.

image


WolframPfeifer:
@mattulbrich This should fix the issue with CVC4, and also some with cvc5 which occurred for me.

wadoon (approved):

PR #3597, opened 2025-04-14
KeY/JSON config for SMT solvers

wadoon:
## Intended Change

The SMT solvers are defined in unstructured untyped properties files. In the meantime, we have updated our KeY config to a JSON-like format, we should also update our SMT solvers definition files. 

This comes with this PR. 

## Plan 

* [x] rewriting history

## Type of pull request

## Ensuring quality

- CI


wadoon:
KaKeY: Go ahead.

mattulbrich:
*Agreed at meeting*: Good idea, but let's have individual json files (to allow for additions)

wadoon:
Two errors should remain: 

1. `spotlessApply` does not fix the formatting locally. 
2. EOFException b/c Stream is closed when sending `(get-unsat-core)` to Z3.

flo2702 (approved):

PR #3599, opened 2025-04-17
Using FlatLAF for a modern look and feel

wadoon:
This PR brings FlatLAF as the default LaF for KeY, supporting both light/dark modes.  

To make this works, it required cleaning in the current UI code. 

New (old) rules for UI code: 

1. No stupid default colors like `Color.white` or `Color.black` for foreground/background of UI components. Instead, you should use: `UIManager.getColor(...)` to get the appropriate current color. 
2. Every developer-picked color should exist in a dark/light mode variant. 
3. UI code needs to be reactive for LaF changes. Either by listening with a `PropertyChangeListener` on the `ViewSettings` or by overriding `Component#updateUI()`. Later, one is triggered after the LaF is set. 
4. LaF-dependent objects (like colors, icons) should not be final or static.

image

image

## Plan

* [x] use `SwingUtillities.updateComponentTreeUi` to update the UI after laf switch
* [x] overwrite `Component.updateUI()` if necessary to update the icons. 

* [x] fix settings menu in dark mode:
  
  image

* [x] Tooltip settings dialog
  
  image

* [x] Proof Tree renderer
  
  image

* [x] Testgen window
  
  image

* [x] Color for syn. highlighting in Source / Sequent view

  image

* [x] Color settings support for light and dark colors.

* Not happened: Detecting the current light/dark mode settings from OS. 
  * This requires: CLI calls for Linux, registry access on Windows using JNA, or native libs on macOS using JFA. There is a library that encapsulates everything but the deps are too strong. 
  

## Type of pull request

  - removal of hard coded colors in some components 
  - support for dark/light switches (two icon colors, etc.)

## Ensuring quality
- Looking at the UI for a long time.

### DIFF
--- a/key.ui/src/main/java/de/uka/ilkd/key/gui/MainWindow.java
+++ b/key.ui/src/main/java/de/uka/ilkd/key/gui/MainWindow.java
@@ -410,6 +410,7 @@ public final class MainWindow extends JFrame {
 
         if (instance == null) {
             updateLookAndFeel();
+            instance = new MainWindow();
             final var viewSettings = ProofIndependentSettings.DEFAULT_INSTANCE.getViewSettings();
             final PropertyChangeListener propertyChangeListener = (evt) -> updateLookAndFeel();
             viewSettings.addPropertyChangeListener(
@@ -417,7 +418,6 @@ public final class MainWindow extends JFrame {
             viewSettings.addPropertyChangeListener(
                 ViewSettings.PROP_LOOK_AND_FEEL, propertyChangeListener);
 
-            instance = new MainWindow();
             if (ensureIsVisible) {
                 instance.setVisible(true);
             }
@@ -439,6 +439,14 @@ public final class MainWindow extends JFrame {
             // enable custom window decorations
             JFrame.setDefaultLookAndFeelDecorated(viewSettings.isDefaultLookAndFeelDecorated());
             JDialog.setDefaultLookAndFeelDecorated(viewSettings.isDefaultLookAndFeelDecorated());
+        }
+
+        for (Window w : Window.getWindows()) {
+            SwingUtilities.updateComponentTreeUI(w);
+        }
+
+        if (instance != null) {
+            // SwingUtilities.updateComponentTreeUI(instance);
         }
     }
 
--- a/key.ui/src/main/java/de/uka/ilkd/key/gui/MainWindow.java
+++ b/key.ui/src/main/java/de/uka/ilkd/key/gui/MainWindow.java
@@ -64,7 +64,10 @@ import de.uka.ilkd.key.settings.ViewSettings;
 import de.uka.ilkd.key.smt.SolverTypeCollection;
 import de.uka.ilkd.key.smt.solvertypes.SolverType;
 import de.uka.ilkd.key.ui.AbstractMediatorUserInterfaceControl;
-import de.uka.ilkd.key.util.*;
+import de.uka.ilkd.key.util.KeYConstants;
+import de.uka.ilkd.key.util.KeYResourceManager;
+import de.uka.ilkd.key.util.PreferenceSaver;
+import de.uka.ilkd.key.util.ThreadUtilities;
 
 import org.key_project.logic.Name;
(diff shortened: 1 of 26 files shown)


key.ui/src/main/java/de/uka/ilkd/key/gui/MainWindow.java
```diff
@@ -404,6 +407,15 @@ public static MainWindow getInstance(boolean ensureIsVisible) {
             return instance;
         }
         if (instance == null) {
+            //FlatDarculaLaf.setup();
+            FlatLightLaf.setup();
+
+            if( SystemInfo.isLinux ) {
+                // enable custom window decorations
+                JFrame.setDefaultLookAndFeelDecorated( true );
+                JDialog.setDefaultLookAndFeelDecorated( true );
```
Drodt, line 416: Fix formatting. Then I (and spotless) are happy

Drodt (dismissed):

FliegendeWurst (dismissed):
https://github.com/KeYProject/key/blob/main/key.ui/src/main/java/de/uka/ilkd/key/gui/settings/StandardUISettings.java#L32-L40 should be updated as well

wadoon:
> https://github.com/KeYProject/key/blob/main/key.ui/src/main/java/de/uka/ilkd/key/gui/settings/StandardUISettings.java#L32-L40 should be updated as well

Do not need to update. The mechanism is in the constructor.

unp1 (approved):
Thanks!

mattulbrich:
Updating the looks is generally a good idea; even for KeY. However, I agree that there are some glitches that should have been fixed before merging.
* Why are there boxes around the tree elements in the proof tree?
* The background of inner and leaf proof nodes is the same
* Most text fields have a grey background. That is not good. Incl. the source pane
* The default colour setup is not really good. The dark mode is not usable.
* For me: the foot steps button is never highlighted when hovering.

Let's look at it in the next meeting.

wadoon:
> In particular:
> 
> * low contrast in proof tree (selected step is with black text in dark blue background)

I tested the colors on my screen, with the i-do-not-know color model. Make a suggestion of a better color.  

> * lower contrast in SourceView (on the right): grey background

In the current LaF themes, white symbolizes an active/editable component. `SourceView` is not editable. 

Also, you should use the appropriate background color from the LaF, and not set it to white.
This also destroyed the use of other LaFs, esp. dark ones. 

> * No visual difference between SequentView for closed and open nodes (formerly had white background when open)

Again, `setBackgroundColor(WHITE)` is not the solution to the problem. 

> * worse syntax highlighting of SequentView (black bold instead of blue bold for most keywords)

The only change was that the colors are now configurable and inserted into the CSS string.

> * no visual separation between tabs, menus etc.

Comes from the LaF, resembles the current style of many OS. 

These are rather minor issues.

> I really don't understand why this was merged (in particular, without getting feedback from more of the KeY developers).

Because we have different opinions on what should be on `main` and what should not. Some of us are more liberal and see `main` also as an incubator. 

 
> Note that chaning the LaF back to Metal does **not** bring back the old KeY, but looks really different as well now (even after restarting KeY).

Mhm... the window decoration setting in MainWindow. 

> I vote for reverting that, or the design should be fixed (I think that you can set all the colors manually, but we really should ship KeY with good default design)!

wadoon:
> * Why are there boxes around the tree elements in the proof tree?

Yes, the borders were always there and set by some renderer. 

At some past point, there is/was the need to display too much information in the tree, e.g., exploration steps, symbolic execution, ... 

> * The background of inner and leaf proof nodes is the same

The background follows the LaF. In both cases, the component is not editable.
`setBackground(WHITE)` is not the answer, the right value is: `TextField.background ---> javax.swing.plaf.ColorUIResource[r=70,g=73,b=75]`. 

> * Most text fields have a grey background. That is not good. Incl. the source pane

Not editable, is not editable. 

> * The default colour setup is not really good. The dark mode is not usable.

Works for me. 

> * For me: the foot steps button is never highlighted when hovering.

Works also for me. 

image

PR #3600, opened 2025-05-05
Warn on missing SMT solvers if flag is set

wadoon:
Currently, if SMT solvers are not available during testing, these test cases will be silently ignored---a behavior acceptable on local machines, but not on CI. 

On CI, this is firstly ensured by executing `z3` and `cvc5` on command-line. This ensures installation, but not that KeY finds the executable. 

This PR changes the behavior of SMT test cases to failing if SMT solvers are not present, and an additional flag is set.

This flag is set iff the environment variables sets `failIfSmtSolverIsUnavailable` or `GITHUB_ACTIONS` to `true`. 

## Type of pull request

## Ensuring quality

- I have tested the feature as follows: CI


wadoon:
Migration to cvc5 as far as possible.

mattulbrich:
_Agreed at meeting_: Retire CVC4 support. Migrate test cases to CVC5 is possible, drop otherwise.
@WolframPfeifer can migrate test cases from other PR #3592. Then merge that onto master, that might resolve issues here.

wadoon:
wait for https://github.com/KeYProject/key/pull/3592

WolframPfeifer:
#3592 is merged into main, I updated this PR. Also, I migrated the CVC4 test case to cvc5, so this PR should be merge-ready now.

WolframPfeifer:
We need an additional reviewer, since @wadoon and I both have committed to this PR ...
Can somebody else please review? It contains only small changes and should be very easy to review.

Drodt (approved):

PR #3578, opened 2025-03-05
Extract Taclets and Prover Main Loop to ncore

Drodt:
## Intended Change

This pull request is part of the modularization effort in KeY. It contains the first building blocks for generalizing the prover infrastructure, this includes the sequent data structure, as well as proof object, goal, rule, and the prover main loop.

## Type of pull request

- Refactoring (behaviour should not change or only minimally change)
- Breaking change (fix or feature that would cause existing functionality to change)
- There are changes to the (Java) code

## Ensuring quality

    
- I made sure that introduced/changed code is well documented (javadoc and inline comments).
- I have checked that runtime performance has not deteriorated.

## Additional information and contact(s)

     

The contributions within this pull request are licensed under GPLv2 (only) for inclusion in KeY.

PR #3605, opened 2025-05-14
Generalize Taclet Implementation

Drodt:
## Related Issue

This pull request builds on #3578.

## Intended Change

Generalize `Taclet` class in `ncore` to be used for other target languages and logics. In particular, we do not enforce different subclasses, e.g., `FindTaclet`, `AntecTaclet`, `RewriteTaclet`. Instead we use a (nullable) `SyntaxElement` for the find expression. This may be a `Term` (e.g., for "rewrite taclets") or a `Sequent` (e.g., for antecedent taclets).

This PR does not yet move the matching, application, or execution logic. This will be done in a future PR.

## Plan

* [x] Wait on #3578 to be merged.

## Type of pull request

## Ensuring quality

    
- I have tested the feature as follows: The code is used implicitly in JavaKeY and tested by the relevant test cases.

@unp1 
     

### DIFF
--- a/key.core.example/src/main/java/org/key_project/Main.java
+++ b/key.core.example/src/main/java/org/key_project/Main.java
@@ -3,7 +3,9 @@
  * SPDX-License-Identifier: GPL-2.0-only */
 package org.key_project;
 
-import java.io.File;
+
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.*;
 
 import de.uka.ilkd.key.control.KeYEnvironment;
@@ -39,7 +41,7 @@ public class Main {
      */
     public static void main(String[] args) {
         LOGGER.info("Starting KeY example application.");
-        File location = args.length == 1 ? new File(args[0]) : new File("example");
+        Path location = Paths.get(args.length == 1 ? args[0] : "example");
         // Path to the source code folder/file or to a *.proof file
         try {
             // Ensure that Taclets are parsed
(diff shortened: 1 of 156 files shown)


key.core/src/main/java/de/uka/ilkd/key/logic/OpCollectorJavaBlock.java
```diff
@@ -15,10 +16,10 @@
  */
 public class OpCollectorJavaBlock extends OpCollector {
     @Override
-    public void visit(Term t) {
+    public void visit(org.key_project.logic.Term t) {
         super.visit(t);
-        if (t.javaBlock() != JavaBlock.EMPTY_JAVABLOCK) {
-            var collect = new JavaASTCollector(t.javaBlock().program(), LocationVariable.class);
+        if (t.op() instanceof Modality mod && !mod.program().isEmpty()) {
+            var collect = new JavaASTCollector(mod.program().program(), LocationVariable.class);
```
WolframPfeifer, line 22: `mod.program().program()` looks really weird. It should be correct, but I think the names are not good here.

key.core/src/main/java/de/uka/ilkd/key/proof/TermProgramVariableCollector.java
```diff
@@ -34,9 +36,9 @@ public void visit(Term visited) {
             result.add(variable);
         }
 
-        if (!visited.javaBlock().isEmpty()) {
+        if (visited.op() instanceof Modality mod && !mod.program().isEmpty()) {
             ProgramVariableCollector pvc =
-                new ProgramVariableCollector(visited.javaBlock().program(), services);
+                new ProgramVariableCollector(mod.program().program(), services);
```
WolframPfeifer, line 41: Same as in the other remark

WolframPfeifer (approved):
Thanks, I think this PR is good and ready. I have some minor remarks, which you might want to change.

Btw.: I only learned about markdown comments during this review, and think it is a good idea to use them. Does our spotless formatter support them? Sometimes it seems that we got a bit weird formatting (e.g. lines with only a single word) ... Also, a separate PR would have made it much easier to review, since now there are lots of files touched with only converted comments.

key.ncore.calculus/src/main/java/org/key_project/prover/rules/Taclet.java
```diff
@@ -430,26 +424,106 @@ public String toString() {
...
         return (TacletExecutor<@NonNull G, ?>) executor;
     }
 
-    public abstract Taclet setName(String s);
+    public abstract Taclet setName(String name);
 
     public ImmutableList<RuleSet> getRuleSets() {
         return ruleSets;
     }
+
```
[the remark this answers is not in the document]
WolframPfeifer: Sorry, marked wrong line. The comment is about the record `ApplicationRestriction`.

PR #3606, opened 2025-05-16
Fix SMT ApplyAction not pruning when undoing

### LINKED ISSUE #3601: Can't undo SMT Solver Apply Action
## Description

The SMT Solver Apply Action creates additional nodes for all goals to be closed. When trying to undo the Apply Action these additional nodes are not removed.

## Reproducible
  
always

### Steps to reproduce 

1. Load Contraposition example
2. Run SMT Solver
3. Apply Result
4. Press Undo

Expect proof state to return to original state before calling SMT solvers.
Actually does not remove the nodes added by the apply action. Instead the goal after the added nodes are internally reopened. The proof tree only reflects this when inspecting another node.

### Additional information

Image

---

* Commit: 
[2208fad](https://github.com/KeYProject/key/commit/2208fad878b291b0f616c0f7e19fb1c34e9dc436)

BookWood7th:
## Related Issue

This pull request resolves #3601 .

## Intended Change

The ProofSMTApplyAction now performs an undo by pruning the proof instead of reopening goals. This ensures there are no leftover nodes in the proof tree from performed apply actions.

## Type of pull request

## Ensuring quality

    
- I have tested the feature as follows: KeY Test Suite

### DIFF
--- a/key.ui/src/main/java/de/uka/ilkd/key/gui/actions/useractions/ProofSMTApplyUserAction.java
+++ b/key.ui/src/main/java/de/uka/ilkd/key/gui/actions/useractions/ProofSMTApplyUserAction.java
@@ -39,6 +39,12 @@ public class ProofSMTApplyUserAction extends UserAction {
      * The number of goals that will be closed by this action.
      */
     private final int numberOfGoalsClosed;
+    /**
+     * The original nodes of the proof. Used to undo this action
+     */
+    private final Collection<Node> originalProofNodes;
+
+    private final Node originalSelectedNode;
 
     public ProofSMTApplyUserAction(KeYMediator mediator, Proof proof,
             Collection<SolverListener.InternSMTProblem> smtProblems) {
@@ -48,6 +54,8 @@ public class ProofSMTApplyUserAction extends UserAction {
                 .filter(p -> p.getProblem().getFinalResult()
                         .isValid() == SMTSolverResult.ThreeValuedTruth.VALID)
                 .count();
+        this.originalProofNodes = proof.openGoals().stream().map(Goal::node).toList();
+        this.originalSelectedNode = mediator.getSelectedNode();
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


key.ui/src/main/java/de/uka/ilkd/key/gui/actions/useractions/ProofSMTApplyUserAction.java
```diff
@@ -39,6 +39,12 @@ public class ProofSMTApplyUserAction extends UserAction {
      * The number of goals that will be closed by this action.
      */
     private final int numberOfGoalsClosed;
+    /**
+     * The original nodes of the proof. Used to undo this action
+     */
+    private final Collection<Node> originalProofNodes;
```
wadoon, line 45: I have my problem this storing the nodes:

1. Nodes point to Proof. Is it ensured that the lifetime of this class is shorter than that of the proof? Otherwise, we have a memory leak.

2. Normally, Actions are UI global and shared across proofs. Is that also the case here? Have you tested with multiple proofs? Have you tested with sequential multiple SMT applications?

BookWood7th: I do not believe storing the nodes is a problem. Currently the canUndo check is a little counterintuitive since it checks if the nodes of the goals closed by this action are present. This is convoluted but should evaluate to finding all of the originalProofNodes stored here. I'm open to changing this to finding the nodes directly.

1. The Proof class provides a dispose method, which should prevent memory leaks caused by storing references to any instances of proof. Any memory leaks are therefore not caused by this class holding onto references to Proof, but an incorrect usage of the referenced Proof class. Furthermore, this would cause an issue for all inheritors of UserAction, as they all hold references to proof and a Getter for this reference.
2. While this may be the case in other programs, KeY does not store actions across proofs. All actions are proof specific, as is mentioned in the tooltip when hovering over the undo button. This can also be seen in the history dropdown. The implementation for this is found in the ActionHistoryExtension class, which stores a Map of Proofs and the List of their respective actions.

Undoing across multiple proofs only undoes the last action on the selected proof. This works as intended.

I am not entirely sure, what you mean by "sequential multiple SMT applications", but it does work both for applying SMT to different goals sequentially and undoing them afterwards, as well as applying and undoing repeatedly on the same goal.

WolframPfeifer: I think there is a confusion here. In our code base we have `UserAction` and all its descendants, which do **not** inherit `javax.swing.AbstractAction`, but represent an action that the user can do in the GUI and which can be undone. Maybe we should find another name for this, since it is inconsistent with swing actions ...

wadoon: Are the undo steps deleted after the proof is closed?

PR #3608, opened 2025-05-24
Move Taclet matching infrastructure to ncore

unp1:
## Intended Change

This PR generalizes the taclet matching logic and moves common infrastructure up to ncore.

It also improves upon 
* cohesion for the VM based program logic as the concerns of match program creation and program interpretation are now separated
* VM instructions are simpler and more focused on one functionality (e.g., cursor placement not convoluted with match instruction; or dynamic case distinctions in match instructions reduced by creating a more specific program for these cases)

This PR is part of the modularization effort for supporting other target languages.

## Plan

* [x] Wait for PR #3605 to be merged upon which this PR is based

## Type of pull request

## Ensuring quality

    
- I have tested the feature as follows: standard tests without failure (these test the extracted infrastructure as well)

@Drodt 

### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/nparser/builder/TacletPBuilder.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/nparser/builder/TacletPBuilder.java
@@ -517,8 +517,8 @@ public class TacletPBuilder extends ExpressionBuilder {
             rs.forEach(b::addRuleSet);
         }
 
-        if (ctx.DISPLAYNAME() != null) {// last entry
-            b.setDisplayName(accept(ctx.dname));
+        if (ctx.DISPLAYNAME() != null && !ctx.DISPLAYNAME().isEmpty()) {// last entry
+            b.setDisplayName(Objects.requireNonNull(accept(ctx.dname)));
         }
 
         if (ctx.HELPTEXT() != null) { // last entry
--- a/key.core/src/main/java/de/uka/ilkd/key/rule/tacletbuilder/TacletBuilder.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/rule/tacletbuilder/TacletBuilder.java
@@ -31,6 +31,7 @@ import org.key_project.util.collection.ImmutableList;
 import org.key_project.util.collection.ImmutableSLList;
 import org.key_project.util.collection.ImmutableSet;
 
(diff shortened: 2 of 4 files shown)


wadoon:
> @wadoon: Do you have an idea why? Something wrong with the merge to main. I checked TestMiscTools and MiscTools they seem to be the same.

I am not sure why this test case was negative on my branch and is now positive. 

I have looked into the test case and would say it's wrong.

The path `\abc\def` or `/abc/def` is valid on Windows. The missing drive letter will be added from the current default path.    

```
   @Test
    public void testMakeFilenameRelativeWindows() {
        // run only on Windows systems
        if (File.separatorChar != '\\') {
            return;
        }

        // test windows delimiters
        String s = "C:\\Windows";
        String t = "c:\\";
        String u = MiscTools.makeFilenameRelative(s, t);
        assertEquals("Windows", u);
        // do stupid things
        try {
            t = File.separator + "home" + File.separator + "daniel";
            u = MiscTools.makeFilenameRelative(s, t);
            fail();
        } catch (RuntimeException e) {
            assertTrue(true);
        }

    }
```

I switched the string-based implementation for a java.nio one.

wadoon:
> > I switched the string-based implementation for a java.nio one.
> > Thanks!

fix #3619

wadoon (approved):
Would it be possible to create a new module `key.ncore.matching`?

wadoon (changes requested):
There is an error in the checkerframework configuration. You need to adapt or remove the package constraint. 

```diff
diff --git a/key.ncore.calculus/build.gradle b/key.ncore.calculus/build.gradle
index fb7fa96b0f..0a75120fe6 100644
--- a/key.ncore.calculus/build.gradle
+++ b/key.ncore.calculus/build.gradle
@@ -23,7 +23,7 @@ checkerFramework {
                 "org.checkerframework.checker.nullness.NullnessChecker",
         ]
         extraJavacArgs = [
-                "-AonlyDefs=^org\\.key_project\\.logic",
+                //"-AonlyDefs=^org\\.key_project\\.prover",
                 "-Xmaxerrs", "10000",
                 "-Astubs=$rootDir/key.util/src/main/checkerframework:permit-nullness-assertion-exception.astub",
                 "-AstubNoWarnIfNotFound",
```

PR #3610, opened 2025-05-26
Fix Checker Framework Errors

Drodt:
## Intended Change

Currently, the Checker Framework fails, causing all CI tests to fail. This PR fixes some annotations in key.util, which makes the framework pass.

## Type of pull request

## Ensuring quality

- I have tested the feature as follows: Ran checker framework on my machine


wadoon:
> But they it failed? The annotation were correct on main.

Drodt:
The annotations were not entirely correct. They did not allow for null values even where they are fine and where we test for null values being handled correctly.

PR #3611, opened 2025-05-26
SMT Refactoring

BookWood7th:
## Intended Change

This PR aims to refactor the SMT parts of KeY. Among other things this PR will replace the current homemade solutions for solver scheduling with methods of ThreadPool classes. This should cut down on thread creation overhead.

## Plan

* [x] Implement ThreadPool scheduling
* [x] Overhaul SMTResult to avoid requests to SMTSolver pertaining to its result
* [x] Overhaul Sockets
* [x] Overhaul SimplePipe
* [ ] Overhaul SolverCommunication
* [ ] Overhaul SMTProblem
* [ ] Overhaul SolverListener
* [ ] Document Class Responsibilities 
* [ ] Testing
* [ ] Code Formatting

## Type of pull request

## Ensuring quality

    
### DIFF
--- a/key.core.testgen/src/main/java/de/uka/ilkd/key/smt/testgen/AbstractTestGenerator.java
+++ b/key.core.testgen/src/main/java/de/uka/ilkd/key/smt/testgen/AbstractTestGenerator.java
@@ -35,6 +35,7 @@ import de.uka.ilkd.key.settings.ProofIndependentSMTSettings;
 import de.uka.ilkd.key.settings.ProofIndependentSettings;
 import de.uka.ilkd.key.settings.TestGenerationSettings;
 import de.uka.ilkd.key.smt.*;
+import de.uka.ilkd.key.smt.communication.AbstractCESolverSocket;
 import de.uka.ilkd.key.smt.model.Model;
 import de.uka.ilkd.key.smt.solvertypes.SolverType;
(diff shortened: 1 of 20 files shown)


BookWood7th:
> I am not sure if this is the right time to do this. In Java, the jep for structured concurrency is on the way and truly the thing we want.

Thanks for bringing this to my attention!
As far as I'm aware, the SMT Translation subtasks are fairly independent and should be treated as such. While they do spawn additional threads in the main branch version, this behavior is no longer present in this PR.
Each `SMTSolver` instance implements Runnable/Callable and is handled in a purely sequential manner (with the exception of the creation of an external process, which falls outside the scope of Java concurrency to my knowledge)

StructuredConcurrency would offer advantages in the case of unexpected failures due to bugs in the code of KeY, such as null pointer exceptions when not adding a listener to instances of `SMTSolverImplementation`, where this PR checks if any Future completed with an exception after it reports that it is "done". How KeY should behave in cases like these is unclear to me. Current main branch behavior is not suitable, as it fails without any notice in either log or UI (only as uncaught exceptions in the console).
Another point where this could be added is the `SMTInvokeAction`, which creates a new Thread to perform its task.

Both of these changes would still benefit from a refactoring and would be easier to implement afterwards in my opinion.

PR #3621, opened 2025-06-16
Migrate from ST4 to Apache Freemarker in keyext.proofmanagement

wadoon:
## Related Issue

ST4 is blocking from module-info usage, see #3451. 
This PR swaps ST4 against Apache Freemarker. 

## Type of pull request

## Ensuring quality
    

WolframPfeifer:
On my branch `pfeifer/rmst` I fixed the report generation. I reverted some changes you did, for example moving `CheckerDate.ProofEntry` from an inner (non-static!) class into a top-level class. It is not pretty, but seems to work there now.

WolframPfeifer:
> On my branch `pfeifer/rmst` I fixed the report generation. I reverted some changes you did, for example moving `CheckerDate.ProofEntry` from an inner (non-static!) class into a top-level class. It is not pretty, but seems to work there now.

Ah no, spoke too soon! Some things are still wrong in the report (filtering of messages, settings entries, CSS) ...
But at least there is no Freemarker exception any more.

wadoon:
> Ah no, spoke too soon! Some things are still wrong in the report (filtering of messages, settings entries, CSS) ... But at least there is no Freemarker exception any more.

With f3459addbff90f6c2b9fbe237a03581b1b551692, the report is generated successfully w/o exception.
I cherry-picked your branch,  made some smaller adjustments, and already added #3622. It should be fine now.

wadoon:
Added svg-based diagram, that also shows entries with value 0.

image

image

PR #3623, opened 2025-06-19
module-info.java, 2nd attempt

wadoon:
Second run for module-info in KeY after the drop of ST4. Builds-upon #3621, should merged afterwards.  

This pull request adds module support in KeY by adding `module-info.java` and doing the necessary stuff. There are three main issues with modules in KeY: 

- clashing Java packages (packages are sealed)
  (solved by renaming some packages)

- Docking Frames is not-modularized 
  (Baked a new version, with renamed packages. **This version is on central snapshot repo, and should then be deployed to Maven central)

- Caching extension in wrong service file

Additional: 

- Removal of ANTLR2+3 from KeY. 

Hint:

- Isabelle translation added a Scala library, which breaks the Java module convention, but it seems not to break the JVM module system (invalid module name). 
 

## Type of pull request


Drodt:
Is this really ready for review?

wadoon:
> Is this really ready for review?

A review should wait for #3621. This PR builds upon it (removal of ST4).

unp1 (changes requested):
The Proof Caching extension is no longer there, when I start it via ./gradlew :key.ui:run
 - we need most likely a mainModule in addition/instead of mainClass

But the problem then could be that ServiceLoader.load(...) for proof macros, scrips, PO, etc. have no declared uses in module-infos no opens, which means the on the module path ...util.reflection thorws InaccessibleObjectExceptions which is aproblem when the application is run via classpath

Shoul dwe expose in  key-core the requires directives as transitive, so that depending modules don;t have to redeclare them?

wadoon:
We should rather give up on modules. In KaKeY, the predecessor PR #3845 was also not favoured.

wadoon:
@unp1: 
> Should we expose in key-core the requires directives as transitive, so that depending modules don;t have to redeclare them?

That is something I have on different repos, I also already made a proposal with #3638.

PR #3639, opened 2025-07-19
Fix multiple SLF4J providers

BookWood7th:
## Intended Change

Excludes transitive dependency on slf4j-simple, which was included by scala-isabelle before.

## Type of pull request

- There are changes to the deployment/CI infrastructure (gradle)

## Ensuring quality

    
- I have tested the feature as follows: Test suite, Manual testing of Isabelle Translation


WolframPfeifer (changes requested):
If it works, it is fine. Without having tested, I assume logging is still done, just through the main logger of KeY then (?).

My only request would be to have a line of comment with a short explanation, why this is necessary. Otherwise, we will for sure not remember in a few months/years.

BookWood7th:
> If it works, it is fine. Without having tested, I assume logging is still done, just through the main logger of KeY then (?).
> 
> My only request would be to have a line of comment with a short explanation, why this is necessary. Otherwise, we will for sure not remember in a few months/years.

I added a comment to explain this. When slf4j-simple is excluded, Gradle falls back to the logback dependency of KeY for slf4j.

WolframPfeifer (approved):
Thanks!

PR #3643, opened 2025-07-30
Fix NPE using KeY command line interface caused by Java NIO

wadoon:
## Related Issue

Using KeY one CLI with a relative path w/o any parent results into a NPE. Resolving CLI arguments to absolute path resolves this issues. 

Problem arise from `java.nio.Path.getParent()` return `null` for paths likes `A.java` in any working directory. 

Accordingly to javadoc, this is also the behavior of `java.io.File.getParentFile()`, but it did not fail back then. 

## Intended Change

The following works:

```sh
cd /tmp/k/
java -jar $KEY contraposition.key
``` 

## Plan

* [x] Fix it. 

## Type of pull request

## Ensuring quality

- Manually tested.


key.ui/src/main/java/de/uka/ilkd/key/util/CommandLine.java
```diff
@@ -402,7 +402,7 @@ public ArrayList<Path> getFileArguments() {
 
             File tmp = new File(fileArg);
             if (tmp.exists()) {
-                ret.add(tmp.toPath());
+                ret.add(tmp.toPath().toAbsolutePath());
```
mattulbrich, line 405: Does this help on Windows? Does `X:\keyfile.key` have a parent path?
(I cannot check unfortunately)

key.ui/src/main/java/de/uka/ilkd/key/util/CommandLine.java
```diff
@@ -402,7 +402,7 @@ public ArrayList<Path> getFileArguments() {
 
             File tmp = new File(fileArg);
```
mattulbrich, line 403: Why create a File first and then transform to a Path?

mattulbrich (approved):

key.ui/src/main/java/de/uka/ilkd/key/util/CommandLine.java
wadoon: Minimal invasive fix.

key.ui/src/main/java/de/uka/ilkd/key/util/CommandLine.java
wadoon: This is not the crucial case. `parentOf("X:\xxx.key")` is `X:\\`. It is the same problem as in Linux: `parentOf("xxx.key")` is `null`.

Therefore, it should also solve this problem on all platforms.

PR #3652, opened 2025-08-13
Add Polymorphic Sorts and Functions

Drodt:
## Related Issue

Based on #3384. Thanks, @mattulbrich @wadoon 

## Intended Change

Adds polymorphic sorts as proposed by Ulbrich and Weigl and adds polymorphic functions as an obvious extension. This PR also adds matching logic, which has been tested in RustyKeY and JavaKeY.

The syntax is very verbose right now. Generic parameters are written as `<[E1, E2, E3]>`. As an example, see polymorphic sequences:
```key
\sorts {
    \generic E;
    PSeq<[E]>;
}

\functions {
    // getters
    E pseqGet<[E]>(PSeq<[E]>, int);
}
```
In the future, the syntax may be simplified and parameters may be partially inferred.

Additionally, polymorphic datatypes are supported:
```
\datatypes {
	List<[E]> = Nil | Cons(E head, List<[E]> tail);
}
```
The taclet generation for data types is changed accordingly.

To allow for taclets where an input is required (e.g., induction over `List<[int]>`), we had to modify the input for terms and instantiations in proof files to allow sort specification, i.e., `x:List<[int]>`. Consequently, the sort is also now emitted for all instantiations in all proof files.

## Plan

* [x] Test changes
* [x] Documentation
* ~Remove SortDependingFunction? (Also simplifies grammar!)~ (Future work)
* ~Make sequences, sets, etc. polymorphic?~ (Future work)

## Type of pull request

## Ensuring quality

    
  `.github/workflows/tests.yml`


Drodt:
I tested some of the runAllProofs tests that fail on the CI. They work on my machine. I don't know what the issue is

Drodt:
All tests pass now. This PR is ready for review.

wadoon (approved):

WolframPfeifer:
> Not sure if I understand exactly what you mean but generic sorts are fine in taclets, even in parametric sorts and functions. See the examples I added in this PR.

Yes, I have seen these. But these are not user files, right? That is, if you try to load them in the KeY GUI, you get an exception (Sth like "Generic sort not allowed"), because the generic sorts have been removed already. So it is not possible to have a \problem block that refers to a polymorphic sort (only to concrete instances), and likewise with a \rules block in a **user** file. That is what I mean by adding rules later.

PR #3653, opened 2025-08-15
FIX: URGENT: Changes in Gradle 9 results into non-execution of custom test tasks.

wadoon:
**URGENT: ** Test results are not trustworthy anymore. Custom test tasks are not executed. 

Gradle 9 change the default configuration for test tasks. No prominent clue on the changelog was given. The Gradle wrapper is on version 9 since the last version bump. 

The result is that custom test tasks like `testProveRules`, `testRunAllProofs`, ... are unable to find test classes anymore. Therefore, execution succeeds w/o error and w/o execution of the test cases. Hence, the PR checks are useless now. 

This fix needs to be on `main` fast: 

1. It sets the two necessary paths
2. To avoid regression, Gradle fails for all tests if no test cases were executed. 

Manually tested.


mattulbrich:
Recently, we spent more time fixing refactorings that made it to the master. It seems that we should be more restrictive with refactoring PRs and test them a lot more thoroughly before they enter the main branch.

In this case: Why did we upgrade to gradle 9 without this being noticed? Why did our review culture not catch this?

mattulbrich:
Nice catch, Richard :-)
Thanks for the fix, Alexander.

What was the reason to upgrade to Gradle 9? If it is necessary to update the setup: Perhaps such an infrastructural change should be a PR on its own where the submitter and the reviewer check that the infrastructure still behaves as expected?

PR #3654, opened 2025-09-02
Higher Order Proof Scripts

wadoon:
## Intended Change

This PR generalizes the usage of proof script blocks further. 

**Currently,** you are allowed to have *one* proof script block at the end of a command:
```
onAll { auto;}
```

With this PR, proof script blocks can appear as a regular argument anywhere (positional and keyworded):
```
integer-split 
  "<0" : { auto;  }
  "=0" : { instantiate; } 
  ">0" : { tryclose; }
```

This allows for various constructs, e.g., 

```
if "cond" { .. then ..} {.. else ..}
```

## Plan

* [x] Do it
* [x] Test it

## Type of pull request

- New feature (non-breaking change which adds functionality)
- There are changes to the (Java) code

## Ensuring quality
    
- I made sure that introduced/changed code is well documented (javadoc and inline comments).

PR #3657, opened 2025-09-08
Proof Scripts in JML

mattulbrich:
## Related Issue

For quite some time, we have a prototype for JML proof scripts which we finally want to bring to the mater branch.

## Intended Change

Proof scripts can be written into JML comments and can be obeyed during automatic verification.
The major benefit is that the localisation of the proof node to which a proof script should be applied is natural.
The other benefit is that one can use JML expressions and thus does not have to leave the realm of JML to do script-guided proofs.

This relies on #3587 and needs #3654 merged.

## Plan

* [x] Refine the syntax
* [x] Adapt lexer and parser
* [x] Reimport the implementation that was available on an older branch
* [x] Make it more convenient
* [x] Add examples
* [x] Code cleanup
* [x] Document the changes

## Type of pull request

- There are changes to the taclet rule base (possibly small adaptations)

## Ensuring quality

WIP:
- I made sure that introduced/changed code is well documented (javadoc and inline comments). :heavy_check_mark: 
- I made sure that new/changed end-user features are well documented (https://github.com/KeYProject/key-docs). :heavy_check_mark: 
- I added new test case(s) for new functionality. :heavy_check_mark: 
- I have tested the feature as follows: ... examples and test cases
- I have checked that runtime performance has not deteriorated. not for non-script proofs.

Planned to be done Mid 09/25


wadoon:
I have written an integer split command. 

An induction command is on the way with support for int, user-data types, bsum(?). 

Commit: 4644331608e051e73884ef2c11b66490d189fca5

mattulbrich:
The history of this PR was totally changed while I was working on it.
I will have to force-push my changes of the last few days.
In case it is still needed: e841235708a1099a7e0b74c9af68c3c7bf8d5369

wadoon (approved):

mattulbrich:
I believe I fixed the functional issues that were left. It remains:
* Documentation
* Rebasing or merging with main resolving the above conflicts [which are not 100% clear]

There is a chance that this can be done by the end of the month.

mattulbrich:
It seems there is one open test case that has little to do with scripts but with error reporting:

The test case is `(true & true` (missing closing paren), where the error message was:

    ')' (closing parenthesis) expected, but found '}'

but now reads:

    one of '\if', '\ifEx', 'true', 'false', UTF_EMPTY, '(' (opening parenthesis), ... expected, but found '}'

which is obviously wrong. It turns out the parser bails out at the `(` and reports what it expects there and says that it found '}' instead. So these two locations seem to refer to different places in the parsed text.

The last commit that reworked this error messaging is rather recent:  a3480186e9e9e217e820a15636169bdae551eb12

@unp1, perhaps you can look into it?

mattulbrich:
> @unp1, perhaps you can look into it?

While this case shows that something is still amiss with the error reporting, this particular instance could be resolved by commenting out one line ... but considering a parenthesised term as a literal was not ideal anyway.

So: no immediate action needed here. Sorry if I caused inconvenience.

unp1:
> > @unp1, perhaps you can look into it?
> 
> While this case shows that something is still amiss with the error reporting, this particular instance could be resolved by commenting out one line ... but considering a parenthesised term as a literal was not ideal anyway.
> 
> So: no immediate action needed here. Sorry if I caused inconvenience.

no problem. Sorry for not answwering earlier, was preparing for a journal deadline today and did not see it until now.

mattulbrich:
grafik
grafik

I have no idea why the Windows test cases fail. I even used my Windows PC at home to run the tests: And they succeed. ...

unp1:
> I have no idea why the Windows test cases fail. I even used my Windows PC at home to run the tests: And they succeed. ...

Apparently (at least C's analysis) line endings on Windows. It suggests two fixes:

- ScriptLineParser so that users do not run into the problem:
   replace "case '\r' -> { }"  with "case '\r' -> { continue; }" to ignore '\r' as otherwise it ends up in 
   if (state != State.IN_COMMENT) {
    cmdBuilder.append((char) c);   // <-- runs for '\r' too
}

and which I am not such a fan of:

- in .gitattributes (to make tests more robust)
+# The embedded JML proof scripts in these inputs must parse identically on all
+# platforms. Force LF on checkout so Windows does not introduce CR characters
+# that the proof-script tokenizer (ScriptLineParser) is sensitive to.
+*.java text eol=lf

mattulbrich:
> > I have no idea why the Windows test cases fail. I even used my Windows PC at home to run the tests: And they succeed. ...

Thanks for investigating, Richard @unp1 

> Apparently (at least C's analysis) line endings on Windows. It suggests two fixes:

Who is C? Or is it Claude?

This branch is on a reduced github workflow. It seems that the parameters are parsed identically. (see details below). In the current version, all `\r` are filtered out anyways.

What definitely **is** different is the order in which the files are investigated. I added a sort routine to make this order deterministic.

However, for some reason there are now again 8 merge conflicts even though I merged only last week; that is why the tests do not run. I do not know if I will find the time to do this merge in the next days. :(

<details>
<summary>Grep results showing that parameters are identically parsed.</summary>
AutoOnly2 fails on Windows, not on ubuntu.

```
linux:2026-06-27T06:25:25.3047846Z     !!! Parameters for /home/runner/work/key/key/key.core/build/resources/test/de/uka/ilkd/key/scripts/jml/AutoOnly2.java: Parameters{shouldClose=false, method='null', exception='null', deleteTmpDir=true, settings=null}
windows:2026-06-27T06:26:33.9999167Z     !!! Parameters for D:\a\key\key\key.core\build\resources\test\de\uka\ilkd\key\scripts\jml\AutoOnly2.java: Parameters{shouldClose=false, method='null', exception='null', deleteTmpDir=true, settings=null}
```

</details>

unp1:
> > > I have no idea why the Windows test cases fail. I even used my Windows PC at home to run the tests: And they succeed. ...
> 
> Thanks for investigating, Richard @unp1
> 
> > Apparently (at least C's analysis) line endings on Windows. It suggests two fixes:
> 
> Who is C? Or is it Claude?
yupp, at the moment I have still the license (not for long anzmore :-)
> 
> This branch is on a reduced github workflow. It seems that the parameters are parsed identically. (see details below). In the current version, all `\r` are filtered out anyways.
> 
> What definitely **is** different is the order in which the files are investigated. I added a sort routine to make this order deterministic.
> 
> However, for some reason there are now again 8 merge conflicts even though I merged only last week; that is why the tests do not run. I do not know if I will find the time to do this merge in the next days. :(
> 
> Grep results showing that parameters are identically parsed.

If you want I can try to get Claude doing the resolution and we both check afterwards that everything is correct?
Only problem is I don't have access to a Windows machine, so tests will run on CI.

mattulbrich:
If C could do the merge and you review it, that would be splendid.
The history on this branch is a bit broken since I modified the github workflows for quicker testing. So history rewrite would be required here. I was able to partially merge intermediate masters into the branch, but it is rather painful.

My bet and hope is that c0a7ae8e5feb025451008c6cbe021f0377eb2bcb will solve the problem. ..

mattulbrich:
> Only problem is I don't have access to a Windows machine, so tests will run on CI.

I do have a Windows machine: And it goes through there :face_with_open_eyes_and_hand_over_mouth: . But not on the CI machine. Therefore, I assume the indeterministic file order may be the problem.

This of course raises another issue: Why does this order matter? At least on could investigate then. ... if that is the reason.

wadoon:
> I do have a Windows machine: And it goes through there 🫢 . 

Windows != Windows. On Github, it is a Windows Server (2023 or 2025, I think.) 
Behavior on these machines can slightly differ.

In the old days, e.g., JVM decided for `-server` mode on Windows Server resulting into different behavior on volatile fields.

unp1:
> I do have a Windows machine: And it goes through there 🫢 . But not on the CI machine. Therefore, I assume the indeterministic file order may be the problem.
> 
> This of course raises another issue: Why does this order matter? At least on could investigate then. ... if that is the reason.

The reason is actually quite interesting as a reminder of the subtleties between the platforms:

The problem was in SymbolicExecutionMacro: 

 ```
ADMITTED_RULES = Arrays.asList(
                Streams.toString(SymbolicExecutionOnlyMacro.class
                        .getResourceAsStream("SymbolicExecutionOnlyMacro.admittedRules.txt"))
                        .split("\n"));
ADMITTED_RULE_SETS = Arrays.asList(
                Streams.toString(SymbolicExecutionOnlyMacro.class
                        .getResourceAsStream("SymbolicExecutionOnlyMacro.admittedRuleSets.txt"))
                        .split("\n"));
```

depending on the line endings "simplify\n" --> "simplify" and a later equals check to "simplify" when deciding whether to apply the taclet or not succeeded. While when the line endings in the txt file were 
"simplify\r\n" --> "simplify\r" and the later equals check with "simplify" became false and no rules were applied.

The difference between the CI and your machine is most likely the git configuration that converts / does not convert the line endings in the files. Even so, not necessary, the fix for the deprecated ScriptLineParser is also in as additional hardening.

The merge and the commit that fixes the test are now on the branch.

PR #3661, opened 2025-09-18
small fixes to datatype handling

mattulbrich:
## Related Issue

#3662 Parsing problem in \datatypes with no 'self' selectors

## Intended Change

Small changes in data type treatment:
- error messaging improved
- consistent naming scheme
- *corrected EQ taclet in correct order (a=b became b=a at one point)*
- *fixed bug in the Axiom taclet for datatypes* which crashed KeY while loading.
- added generated deconstructor rules to "simplify" ruleset 

## Plan

Small fixes: done.

## Type of pull request

- There are changes to the taclet rule base [only to taclets created at load time for datatype declarations]

## Ensuring quality

- Pair programming
- I added new test case(s) for new functionality. Manually.

Co-authored by @WolframPfeifer 

### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/nparser/builder/FunctionPredicateBuilder.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/nparser/builder/FunctionPredicateBuilder.java
@@ -68,17 +68,18 @@ public class FunctionPredicateBuilder extends DefaultBuilder {
                 args[i] = argSort;
                 var argName = argNames.get(i).getText();
                 SortedOperator alreadyDefinedFn = dtNamespace.lookup(argName);
-                if(alreadyDefinedFn == null) {
+                if (alreadyDefinedFn == null) {
                     alreadyDefinedFn = namespaces().functions().lookup(argName);
                 }
-                if(alreadyDefinedFn == null) {
+                if (alreadyDefinedFn == null) {
                     alreadyDefinedFn = namespaces().programVariables().lookup(argName);
                 }
                 if (alreadyDefinedFn != null
                         && (!alreadyDefinedFn.sort().equals(argSort)
                                 || !alreadyDefinedFn.argSorts().equals(ImmutableList.of(sort)))) {
-                    semanticError( argNames.get(i), "Name already in namespace: %s" +
-                            ". Identifiers in datatype definitions must be unique (also wrt. global functions).", argName);
+                    semanticError(argNames.get(i), "Name already in namespace: %s" +
+                        ". Identifiers in datatype definitions must be unique (also wrt. global functions).",
+                        argName);
                 }
                 Function fn = new JFunction(new Name(argName), argSort, new Sort[] { sort }, null,
                     false, false);
--- a/key.core/src/main/java/de/uka/ilkd/key/nparser/builder/FunctionPredicateBuilder.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/nparser/builder/FunctionPredicateBuilder.java
@@ -78,8 +78,10 @@ public class FunctionPredicateBuilder extends DefaultBuilder {
                         && (!alreadyDefinedFn.sort().equals(argSort)
                                 || !alreadyDefinedFn.argSorts().equals(ImmutableList.of(sort)))) {
                     // The condition checks whether there is already a function with the same name
-                    // but different signature. This is necessarily true if there is a globally defined function
-                    // of the same name and may or may not be true if there is another constructor argument of the
+                    // but different signature. This is necessarily true if there is a globally
+                    // defined function
+                    // of the same name and may or may not be true if there is another constructor
+                    // argument of the
                     // same name.
                     semanticError(argNames.get(i), "Name already in namespace: %s" +
                         ". Identifiers in datatype definitions must be unique (also wrt. global functions).",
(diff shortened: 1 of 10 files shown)


key.core/src/main/java/de/uka/ilkd/key/nparser/builder/FunctionPredicateBuilder.java
```diff
@@ -65,11 +67,19 @@ public Object visitDatatype_decl(KeYParser.Datatype_declContext ctx) {
...
+                if (alreadyDefinedFn == null) {
+                    alreadyDefinedFn = namespaces().programVariables().lookup(argName);
+                }
                 if (alreadyDefinedFn != null
                         && (!alreadyDefinedFn.sort().equals(argSort)
-                                || !alreadyDefinedFn.argSort(0).equals(sort))) {
-                    throw new RuntimeException("Name already in namespace: " + argName);
+                                || !alreadyDefinedFn.argSorts().equals(ImmutableList.of(sort)))) {
+                    semanticError(argNames.get(i), "Name already in namespace: %s" +
+                        ". Identifiers in datatype definitions must be unique (also wrt. global functions).",
```
wadoon, line 81: If you want better error message, you can access `#getOrigin()` to get an human readable definition position, i.e. `filename#line`.

key.core/src/main/java/de/uka/ilkd/key/nparser/builder/FunctionPredicateBuilder.java
```diff
@@ -65,11 +67,19 @@ public Object visitDatatype_decl(KeYParser.Datatype_declContext ctx) {
...
+                    alreadyDefinedFn = namespaces().functions().lookup(argName);
+                }
+                if (alreadyDefinedFn == null) {
+                    alreadyDefinedFn = namespaces().programVariables().lookup(argName);
+                }
                 if (alreadyDefinedFn != null
                         && (!alreadyDefinedFn.sort().equals(argSort)
-                                || !alreadyDefinedFn.argSort(0).equals(sort))) {
-                    throw new RuntimeException("Name already in namespace: " + argName);
+                                || !alreadyDefinedFn.argSorts().equals(ImmutableList.of(sort)))) {
```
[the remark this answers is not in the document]
mattulbrich: Sorry, I do not think I understand this. Which condition? below what?

The changes here are to not only fail on other selectors but also on global functions and progvars.
The old line 71 raised an IOBE in one run.

key.core/src/main/java/de/uka/ilkd/key/nparser/builder/FunctionPredicateBuilder.java
mattulbrich: The filename and line and column seem to be set correctly (judging from the error window). There is no "getOrigin()" method in `FunctionPredicateBuilder`. Can you make the suggestion more concrete?

key.core/src/main/java/de/uka/ilkd/key/nparser/builder/FunctionPredicateBuilder.java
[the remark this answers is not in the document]
Drodt: Either that or add an explaining comment. As is, it requires anyone reading to know quite a lot about the implicit assumptions.

key.core/src/main/java/de/uka/ilkd/key/nparser/builder/TacletPBuilder.java
```diff
@@ -452,7 +459,7 @@ private JTerm createQuantifiedFormula(KeYParser.Datatype_constructorContext cont
         }
 
         if (ind.isEmpty()) {
-            return tb.all(qvs, tb.func(fn, args));
+            return tb.all(qvs, tb.subst(qvX, tb.func(fn, args), phi));
```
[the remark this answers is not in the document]
mattulbrich: done.

key.core/src/main/java/de/uka/ilkd/key/nparser/builder/FunctionPredicateBuilder.java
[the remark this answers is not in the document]
mattulbrich: I added a comment explaining this.

key.core/src/main/java/de/uka/ilkd/key/nparser/builder/FunctionPredicateBuilder.java
Drodt: Since you pass the parser context, the location is calculated by the BuildingException. I see nothing else you should do.

Drodt (approved):

key.core/src/main/java/de/uka/ilkd/key/nparser/builder/FunctionPredicateBuilder.java
wadoon: The `ParserRuleContext` has the position of the duplicate definition, but the user may want to know where it was originally added. 

In KeY, we have an interface [`HasOrigin`](https://github.com/KeYProject/key/blob/f2b5fdcc0f5f3214f15c2791c0b7ddacc4dfc10e/key.ncore/src/main/java/org/key_project/logic/HasOrigin.java#L12) for such a reason. But it is currently only implemented by Sorts and Taclets. I guess that I did not want to have more complicated constructors back then.

mattulbrich: I see. Would a Location not be more appropriate for HasOrigin instead of a string?
At any rate, I would not extend the reach of HasOrigin in this PR and close this conversation.

PR #3663, opened 2025-09-20
Fix proof task tree forced white background.

wadoon:
## Related Issue and Solution

The proof task tree (left upper component in the UI) has a forced white background if it is empty initially. This does not respect LaFs. 

To be more precise, the logic was as follows: 

1. The scroll pane was empty initally (view port not set)
2. The default view port was requested and the background was set to white. 
3. If finally a proof was loaded, the view port was set to the appropriate `JList`.

New logic:

1. Just set the `JList` directly. 

The rest is clean-up of unused methods. 

## Type of pull request
 
## Ensuring quality

- Visual inspection in Dark Theme and Metal. Not coverable by automatic tests.

### DIFF
--- a/key.ui/src/main/java/de/uka/ilkd/key/gui/MainWindow.java
+++ b/key.ui/src/main/java/de/uka/ilkd/key/gui/MainWindow.java
@@ -10,6 +10,7 @@ import java.awt.event.KeyListener;
 import java.awt.event.MouseEvent;
 import java.beans.PropertyChangeEvent;
 import java.beans.PropertyChangeListener;
+import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.nio.file.Path;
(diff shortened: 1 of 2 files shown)


key.ui/src/main/java/de/uka/ilkd/key/gui/MainWindow.java
```diff
@@ -481,36 +476,12 @@ public TermLabelVisibilityManager getVisibleTermLabels() {
...
             awtAppClassNameField.setAccessible(true);
             awtAppClassNameField.set(xToolkit, "KeY");
-        } catch (Exception e) {
-        }
-    }
-
-    /**
-     * Tries to set the configured look and feel if the option is activated.
-     */
-    private void setLaF() {
```
wadoon, line 496: Unused private function

key.ui/src/main/java/de/uka/ilkd/key/gui/MainWindow.java
```diff
@@ -317,11 +315,12 @@ private MainWindow() {
         currentGoalView = new CurrentGoalView(this);
         emptySequent = new EmptySequent(this);
         sequentViewSearchBar = new SequentViewSearchBar(emptySequent);
-        proofListView = new JScrollPane();
         autoModeAction = new AutoModeAction(this);
         mainFrame = new MainFrame(this, emptySequent);
         sourceViewFrame = new SourceViewFrame(this);
         proofList = new TaskTree(mediator);
+        proofListView = new JScrollPane(proofList);
```
wadoon, line 322: Main change, part 1.

Drodt (approved):

PR #3676, opened 2025-10-24
fixes #3670: KeY does not know that enum constants are non-null

### LINKED ISSUE #3670: KeY does not know that enum constants are non-null
## Description

KeY seemingly doesn't have a rule that makes enum constants non-null.

## Reproducible
  
always

### Steps to reproduce 

Try to prove the contract below.

File `Severity.java`
```java
public enum Severity {
    INFO,
    WARNING,
    CRITIQUE	
}
```

File `Checker.java`
```java
public class Checker {
    /*@ public normal_behaviour
      @ ensures \result != null;
      @*/
    Severity checkIt() {
        return Severity.INFO;
    }
}
```

### Additional information

---

* Commit: c95c570785107aafa959651dc5b534aeaff64e8c

mattulbrich:
## Related Issue

This pull request resolves #3670.

## Intended Change

Enum constants are finally supported again (after initial support between 2006 and 2011).

## Features
* Constants are unique
* Constants are never null
* their value of `.ordinal()` can be retrieved.

## Type of pull request

- New feature (non-breaking change which adds functionality): *Bringing back functionality that was once there.*
- There are changes to the (Java) code: *rather locally in some varcond and meta function classes.*
- There are changes to the taclet rule base: *java5.key has been touched*

## Ensuring quality
   
- I made sure that introduced/changed code is well documented (javadoc and inline comments). *yes*

I would like to leave this to @FliegendeWurst and @WolframPfeifer who need this feature in a case study and can study this there, find bugs and add test cases from the study.

Case study by @FliegendeWurst, @WolframPfeifer 


mattulbrich:
I tested it on this code. The methods can be verified. Alas, `different` has an open branch because `Severity.INFO` must be proven to be created. Perhaps a rule similar to the other new rules can be added.

```java
public enum Severity {
    INFO,
    WARNING,
    CRITIQUE	
}

public class Checker {
    /*@ public normal_behaviour
      @ ensures \result != null;
      @*/
    Severity checkIt() {
        return Severity.INFO;
    }

    /*@ public normal_behaviour
      @  ensures \result == 1;
      @*/
    int ord() {
        return Severity.WARNING.ordinal();
    }

    /*@ public normal_behaviour
      @  ensures \result != Severity.CRITIQUE;
      @*/
    Severity different() {
        return Severity.INFO;
    }
}
```

mattulbrich:
Please check for logical soundness. It felt correct, but I have not really checked.

key.core/src/main/java/de/uka/ilkd/key/nparser/varexp/TacletBuilderManipulators.java
```diff
@@ -377,7 +377,7 @@ public IsLabeledCondition build(Object[] arguments, List<String> parameters,
             FREE_3, FREE_4, FINAL_TYPE,
             FREE_5, NEW_TYPE_OF, NEW_DEPENDING_ON, FREE_LABEL_IN_VARIABLE, DIFFERENT, FINAL,
             ENUM_CONST, LOCAL_VARIABLE, ARRAY_LENGTH, ARRAY, REFERENCE_ARRAY, MAY_EXPAND_METHOD_2,
-            MAY_EXPAND_METHOD_3, STATIC_METHOD, THIS_REFERENCE, REFERENCE, ENUM_TYPE,
+            MAY_EXPAND_METHOD_3, STATIC_METHOD, THIS_REFERENCE, REFERENCE, /* ENUM_TYPE, */
```
wadoon: I do not understand this change.

wadoon (commented):
As far as I would tell is that the support is incomplete: 

1. Is `e <: java.lang.Enum` valid for all enums instances `e`?
2. Some methods are missing of [`java.lang.Enum`](https://docs.oracle.com/javase/8/docs/api/java/lang/Enum.html)
3. The static generated `E.values()` method is missing. 

Do you think, we should rather downcompile to Java with final fields?

key.core/src/main/java/de/uka/ilkd/key/nparser/varexp/TacletBuilderManipulators.java
mattulbrich: I can also delete the element.

wadoon:
In which sense, does this break for errors during initialization? 

The JLS gives no direct answer. The nearest case would be ["Example 8.9.2-1. Restriction On Enum Constant Self-Reference"](https://docs.oracle.com/javase/specs/jls/se7/html/jls-8.html#jls-8.9).

mattulbrich:
> This breaks if the enum constructor throws an exception.
> `JMERuleErrorSeverity.java`

Thanks for reporting this point. However, I think this is not an issue with enums but with static initialisation. If one activated static intialisation, then this would/should report the error. If explicitily formulating this as a class with static fields the proof would succeed, too.

wadoon:
1. Make this PR ready (being in non-conflict with main)
2. Generated Enum classes should receive a constructor contract with `signals false;`.

Drodt:
It might be good to look at the changes I made on #3761. For instance, the conversion from JavaParser to KeY ASTs does not actually create an `EnumClassDeclaration` at all. As far as I see, this PR is therefore without effect.

(Some of the enum-related things on #3761 are rather hacky)

wadoon:
@mattulbrich @Drodt 

I have heavily refactored `EnumClassBuilder`. `enum Abc {A,B,C;}` now becomes a bit larger (see below). 

I am still not able to automatically verify the `Checker#ord` from above, but the reason is  that the prover get stuck on `enumOrdinal` symbol. 

But I am able to verify it if I inline `#ordinal()` call. This method is now generated. 

@mattulbrich Do we really need your Taclets and varcond? Can we not solve this by a JML method contract?

The non-null invariant below is not necessary! KeY adds them automatically, but you really needs to add `\static_invariant_for`. I added some invariants for asserting the ADTness of an enum. 

---

```
public final class SimpleEnum extends Enum {
    //@ static invariant A != null && B != null && C != null;
    //@ static invariant A != B;
    //@ static invariant A != C;
    //@ static invariant B != C;
    //@ static invariant (\forall SimpleEnum x; A == x || B == x || C == x);

    public static final SimpleEnum A = new SimpleEnum();

    public static final SimpleEnum B = new SimpleEnum();

    public static final SimpleEnum C = new SimpleEnum();

    private static final String[] $enumConstantNames = { "A", "B", "C" };

    private static final SimpleEnum[] values = { A, B, C };

    public static void values() {
        return values;
    }

    public static SimpleEnum valueOf(String name) {
        if ("A".equals(name))
            return A;
        if ("B".equals(name))
            return B;
        if ("C".equals(name))
            return C;
        throw new IllegalArgumentException();
    }

    public String name() {
        return $enumConstantNames[ordinal()];
    }

    public int ordinal() {
        if (this == A)
            return 0;
        if (this == B)
            return 1;
        if (this == C)
            return 2;
        return 0;
    }
}
```

PR #3678, opened 2025-10-30
Fix Information Flow Proof Loading for KeY Jar File

Drodt:
## Intended Change

Currently, loading a proof obligation for, e.g., information flow, fails if you are using the generated shadow JAR, because no proof obligation loader can be found.

The problem is (apparently), that the symbolic execution module overwrites the list of proof obligation loaders in the META-INF folder.

This PR fixes the issue by simply adding all relevant loaders to the file in `key.core.symbolic_execution` as well.

## Type of pull request

- Other: 

## Ensuring quality

    
- I have tested the feature as follows: Loading and closing a few Inf Flow proofs


wadoon (changes requested):
Hej Daniel, 

the service file is not the right one---it belongs to symbex module. 

In the correct service file, the implementations are already there: 
https://github.com/KeYProject/key/blob/main/key.core/src/main/resources/META-INF/services/de.uka.ilkd.key.proof.init.loader.ProofObligationLoader


If the entries are missing in the shadow jar, then the problem lies with the construction of the shadow jar.

wadoon (approved):

PR #3686, opened 2025-11-18
Add support for Annotation Processors in the Javac Extension

PiIsRational:
## Changes

Extended the Javac Extension to be able to load annotation processors, while the java code is getting checked
in KeY. This includes creating settings for the Javac Extension, to be able to set the checkers. 
Also, the there are changes to the run step of the gradle script
to add (add-export and add-opens) directives to the java runtime such that annotations processors can access the internal java compiler data structures. 
This is required to be done by hand when using KeY as a jar and would require the following (or similar) command:

```bash
java \
    --add-exports jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED \
    --add-exports jdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED \
    --add-exports jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED \
    --add-exports jdk.compiler/com.sun.tools.javac.main=ALL-UNNAMED \
    --add-exports jdk.compiler/com.sun.tools.javac.model=ALL-UNNAMED \
    --add-exports jdk.compiler/com.sun.tools.javac.processing=ALL-UNNAMED \
    --add-exports jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED \
    --add-exports jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED \
    --add-opens jdk.compiler/com.sun.tools.javac.comp=ALL-UNNAMED \
    -jar key-2.12.4-dev-exe.jar
```

## Type of pull request

## Ensuring quality

    
  - I used it with multiple annotation processors
  - I loaded KeY with various settings to check if and when they are deleted

@WolframPfeifer is involved in this pull request.
     


key.ui/src/main/java/de/uka/ilkd/key/gui/plugins/javac/JavaCompilerCheckFacade.java
```diff
@@ -58,69 +51,37 @@ private JavaCompilerCheckFacade() {
...
-            out.save(new OutputStreamWriter(System.err), null);
+                result.stream().map(it -> {
+                    var map = Map.of(
+                        "message", it.text,
+                        "kind", it.getKind().toString(),
+                        "line", it.location.getPosition().line(),
+                        "column", it.location.getPosition().column(),
+                        "additionalInfo", it.getAdditionalInfo());
+                    if (it.location.fileUri() != null) {
+                        map = new HashMap<>(map);
```
wadoon: You should rather use `TreeMap` for determinstic key order. Makes things easier to debug.

WolframPfeifer (changes requested):
Thanks for this PR, which will allow running Checker Framework checkers (e.g. for Nullness of Ownership) from the KeY GUI directly.

Also, thanks to @wadoon for the initial draft of the external checker, and @PiIsRational for polishing.

I have a few minor complaints about the use of `var`. When these are fixed, the PR would be ready from my point of view. Also, there is code smell in the unit test (check without assert), but we could also keep it like that.

WolframPfeifer (approved):

PR #3692, opened 2025-11-26
Fix deadlock when showing JML warnings dialog

FliegendeWurst:
## Intended Change

It seems this method is called on the worker thread, which may lead to a deadlock when creating the dialog.

## Type of pull request

## Ensuring quality

- I have tested the feature as follows: loaded a file with JML warnings (specifically, missing decreases). Now works fine, previously it got stuck on "Reading project.key".

### DIFF
--- a/key.ui/src/main/java/de/uka/ilkd/key/gui/WindowUserInterfaceControl.java
+++ b/key.ui/src/main/java/de/uka/ilkd/key/gui/WindowUserInterfaceControl.java
@@ -592,7 +592,7 @@ public class WindowUserInterfaceControl extends AbstractMediatorUserInterfaceCon
 
     @Override
     public void reportWarnings(ImmutableSet<PositionedString> warnings) {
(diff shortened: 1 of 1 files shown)


wadoon:
I would say, that this is not the patch we would want in just a case.

Could you provide a stack trace showing the location from which the method is invoked?

WolframPfeifer (dismissed):
> Now the warnings dialog will be shown in parallel to the contract chooser, which is a bit odd.

In our meeting, we agreed that this is very difficult to prevent (probably involves writing a class that has an explicit queue of dialogs to show), so we decided to ignore that for the moment.
Thanks, @FliegendeWurst and @wadoon.

WolframPfeifer (approved):
> Now the warnings dialog will be shown in parallel to the contract chooser, which is a bit odd.

In our meeting, we decided to ignore that shortcoming, since a fix would be difficult (probably requires writing a class that keeps an explicit queue of dialogs to show).

Thanks, @FliegendeWurst and @wadoon!

PR #3702, opened 2025-12-18
Check for applicability of Antec-/SuccTaclets in Proof Replay

### LINKED ISSUE #3701: Unsound Proof Replay
## Description

I think this bug is interesting enough to merit an issue. Fix is ready.

The proof replayer does not properly check whether a taclet is actually applicable at the given position. For example, a SuccTaclet like ifElseSplit should only be applied at the top level, but this is not checked at all. Therefore, KeY will accept a proof for the following problem:
```
   x = 0
-> \<{
     x = 1;
     if (false) {}
     else {}
   }\> x = 0
```
We use ifElseSplit below the modality, thereby dropping the update to `x`.

## Reproducible
  
> Is the issue reproducible?
> Select one of: always, sometimes, random, have not tried, n/a 

Always.

### Steps to reproduce 

1. Unzip the attached file
2. Load the unzipped proof file

[if.key.zip](https://github.com/user-attachments/files/24233859/if.key.zip)

Expected: Proof is rejected.
Actual: KeY accepts the proof; formula is shown as proven correct.

---

* Commit: 012874a264

Drodt:
## Related Issue

This pull request resolves #3701.

## Intended Change

Checks whether Antec and SuccTaclets are correctly applied (i.e., the position is top-level).

## Type of pull request

## Ensuring quality

    
- I have tested the feature as follows: Loaded the test file in #3701

### DIFF
--- a/key.core/src/test/java/de/uka/ilkd/key/proof/runallproofs/ProofCollections.java
+++ b/key.core/src/test/java/de/uka/ilkd/key/proof/runallproofs/ProofCollections.java
@@ -942,6 +942,9 @@ public class ProofCollections {
         // There are two possible candidates to instantiate the assumes sequent.
         g.notloadable("proofLoadRepair/insufficient-manipulated.proof");
 
+        // SuccTaclet is applied at a non-top-level position
+        g.notloadable("proofLoadRepair/if.key.proof");
+
         // Verify that taclet instantiations read from proof files are checked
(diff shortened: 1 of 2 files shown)


Drodt:
Entirely unclear to me why testgen tests fail on Windows but not Linux. Most likely no connection to this PR.

PR #3719, opened 2026-01-19
JavaRedux: Integer methods + List::remove + NoSuchElementException

FliegendeWurst:
## Intended Change

The `Integer` class is a bit useless without providing at least `valueOf` and `intValue`, so I added specifications for those methods. And the field `value` to store the int value.

The `List::remove` method to remove elements by index is useful to have.

The `NoSuchElementException` class is required to implement iterators.

## Type of pull request

- Other: change to JavaRedux (bootstrap classpath)

## Ensuring quality

- I have tested the feature as follows: used the newly specified methods in a simple example

### DIFF
--- a/key.core/src/main/resources/de/uka/ilkd/key/java/JavaRedux/java/lang/Integer.java
+++ b/key.core/src/main/resources/de/uka/ilkd/key/java/JavaRedux/java/lang/Integer.java
@@ -10,6 +10,8 @@ public final class Integer implements java.lang.Comparable
    public final static java.lang.Class TYPE;
    public final static int SIZE = 32;
 
+   private /*@ spec_public @*/ final int value;
+
    public static java.lang.String toString(int arg0, int arg1);
    public static java.lang.String toHexString(int arg0);
    public static java.lang.String toOctalString(int arg0);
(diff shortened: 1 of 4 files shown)


FliegendeWurst:
I would like to add a test that checks the below is provable, how would I go about that?

```java
    /*@ public normal_behavior
      @ ensures \result;
      @ assignable \nothing;
      @*/
    static boolean testRemove() {
        List list = new ArrayList();
        list.add("a");
        list.add("b");
        list.add("c");
        Object removed = list.remove(0);
        return removed == "a" && list.get(0) == "b" && list.get(1) == "c";
    }
```

key.core/src/main/resources/de/uka/ilkd/key/java/JavaRedux/java/lang/Integer.java
```diff
@@ -19,11 +21,19 @@ public final class Integer implements java.lang.Comparable
    public static int parseInt(java.lang.String arg0) throws java.lang.NumberFormatException;
    public static java.lang.Integer valueOf(java.lang.String arg0, int arg1) throws java.lang.NumberFormatException;
    public static java.lang.Integer valueOf(java.lang.String arg0) throws java.lang.NumberFormatException;
+   /*@ public normal_behavior
+     @ ensures \result.value == arg0;
+     @ assignable \nothing;
+     @*/
    public static java.lang.Integer valueOf(int arg0);
    public Integer(int arg0);
```
mattulbrich, line 33: should there not also be an assignment to value or at least a contract with the effect in the constructor?

Similar to the static method above.

key.core/src/main/resources/de/uka/ilkd/key/java/JavaRedux/java/lang/Integer.java
```diff
@@ -19,11 +21,19 @@ public final class Integer implements java.lang.Comparable
...
    public Integer(java.lang.String arg0) throws java.lang.NumberFormatException;
    public byte byteValue();
    public short shortValue();
+    /*@ public normal_behavior
+      @ ensures \result == value;
+      @ assignable \strictly_nothing;
+      @*/
    public int intValue();
    public long longValue();
```
mattulbrich, line 47: at least the long conversion can be formulated in JML too.

mattulbrich (approved):
Thanks! I like the additions, when introducing value in Integer, perhaps be a bit more complete. ...

key.core/src/main/resources/de/uka/ilkd/key/java/JavaRedux/java/lang/Integer.java
wadoon: There is still more to say about this conversion. One advantage of `valueOf` over `new Integer` is caching important values. 

The Important values are -128 .. 127, normally. (Configuration options: `VM.getSavedProperty("java.lang.Integer.IntegerCache.high")`) 

```
ensures -128 <= arg0 <= 127 ==> valueOf(arg0) == valueOf(arg0);
```

PR #3732, opened 2026-02-04
Fix nullchecker

Drodt:
## Intended Change

Somehow, null checker fails on main. This PR fixes the error.

## Type of pull request


unp1:
We discussed it before approving the patch and Daniel checked that option. But next() is allowed to return null, if no (still applicable) rule is more in the queue. Returning null is then used to check whether rules are still applicable.

PR #3735, opened 2026-02-10
Optimize UI updates of slicing panel

### LINKED ISSUE #3734: Proof Slicing is automatical active during loading proofs and macros
## Description

During proof loading/replay and macro application, the Proof Slicing UI is constantly updated.

This might be a potential drawback for performance. 

## Reproducible

Always

## Steps to reproduce

Loading a KeY file with a proof is enough:
  
https://github.com/user-attachments/assets/69bab55f-bdb3-4224-99b1-8c13a13a6ec3

> What is your expected behavior, and what was the actual behavior?

Slicing, as well as nearly all UI updates, should be paused during proof loading or macro runs for the sake of performance.

FliegendeWurst:
## Related Issue

This pull request resolves #3734.

## Intended Change

- reduce UI update frequency of slicing panel to once every 500 ms
- reduce work done per rule application to the bare minimum

Alternative: add a listener that listens for macro finish and proof loaded events. Though I think we can afford updating the UI twice a second.

## Type of pull request

## Ensuring quality

- I have tested the feature as follows: UI still updates as expected

### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/proof/Node.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/proof/Node.java
@@ -64,7 +64,7 @@ public class Node implements Iterable<Node> {
 
     private final ArrayList<Node> children = new ArrayList<>(1);
 
-    private RuleApp appliedRuleApp;
+    private @Nullable RuleApp appliedRuleApp;
 
     private NameRecorder nameRecorder;
 
@@ -221,7 +221,8 @@ public class Node implements Iterable<Node> {
         return renamings;
     }
 
-    public RuleApp getAppliedRuleApp() {
+    /// Get the applied rule (null for open/closed goals).
+    public @Nullable RuleApp getAppliedRuleApp() {
         return appliedRuleApp;
     }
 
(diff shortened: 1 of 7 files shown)


wadoon:
I thought the `RuleAppListener`s are disabled during the proof task. In general, we should rather put additional workload after the proof search. 

In this case, the main workload is not to query the graph size of the slicing, rather to calculate the dependency at all.

WolframPfeifer:
> I thought the `RuleAppListener`s are disabled during the proof task. In general, we should rather put additional workload after the proof search.
> 
> In this case, the main workload is not to query the graph size of the slicing, rather to calculate the dependency at all.

I don't think that the RuleAppListeners are disabled in auto mode. Therefore, I think my comment is still valid, and I would prefer to completely get rid of the RuleAppListener for performance reasons. Can we maybe implement the update of the GUI in `selectedNodeChanged` (without a timer)? On the other hand, this would also recalculate when just "scrolling" the proof, right? But probably still better than doing something (even if just starting/restarting a timer) with every rule application.

WolframPfeifer:
> In the extension settings, there is an option "Always track dependencies". If it is not checked, the dependency graph will be created when needed later.

Maybe this should be the default setting then? Also, I noticed that with this flag switched off, the "Dependency Graph" part in the Slicing panel always shows 0 nodes and edges, even after the analysis was run (so the graph for sure is present and non-empty).

FliegendeWurst:
I now removed both RuleAppListeners. Not running the dependency tracker during the proof improves performance by 3-5%.

The UI updates should be fixed.

PR #3758, opened 2026-03-11
Record support by transformation

wadoon:
## Related Issue

This pull request is part of #3774.

## Intended Change

Add a `JavaTransformer` to handle `RecordDeclaration`s by down-compiling to regular Java classes. The regular pattern looks as: 

```java
#modifier record #name(#mod1 #type1 #field1, ...
                       #modN #typeN #fieldN) {
   #bodydeclarations 
}
```

becomes to: 

```java
#modifier #(static when inner-record) final class #name extends java.lang.Record {
   private final #mod1 #type1 #fieldn; 
   ...
   private final #mod1 #typeN #fieldN; 

   public #type1 #name1() { return #name1;} ...
   public #typeN #nameN() { return #nameN;}

   #bodydeclarations 
 
   //# if equals(Object) \not\in #bodydeclaration 
   public final boolean equals(Object o) {
      if(o instanceOf #name n) {
        return java.lang.Objects.equals(#name1, n.#name1) && ... && 
                 java.lang.Objects.equals(#nameN, n.#nameN); 
   }

   //# if hashCode() \not\in #bodydeclaration 
   public final int hashCode() {
      return java.lang.Objects.hash(#name1, ..., #nameN);
   }

   //# if toString() \not\in #bodydeclaration 
   public final String toString() {
      return #name + "["+ 
         ("#name1:" + #name1+",") + ... +
         ("#nameN:" + #nameN+",") + "]";
   }
}
```

## Side Quests:

* Added a printing of JML into the printing of `KeYPipelineTests`. 
  (Not yet completed for all of the thousands JML constructs, but enough for now.)
* Brought back the backticks for JML-escape to KeY.
* Fixes in `IssueDialog` and the search for positions. 
* NPE in ExampleDialog
* It is now possible to simple generate and attach JML to `Node`s inside the transformation pipeline. 

## Plan

* [x] Implement the Java rewriting 
* [x] Contracts?!

## Type of pull request

## Ensuring quality

- I added new test case(s) for new functionality in run-all-proofs
- I added an example for "Load Examples"
- Generated constructor was verified.


key.core/pipelineTests/records/expected/02_RecordClassBuilder/CompactConstructor.java
```diff
@@ -0,0 +1,78 @@
+@javax.annotation.processing.Generated("RecordClassBuilder")
+final class Mapping extends Record {
+    //@ public invariant (\forall Mapping a, b; a.from == b.from && a.to == b.to; a.equals(b));
+    //@ public static invariant (\forall Mapping a; a.equals(a));
+    //@ public static invariant (\forall Mapping a, b; a.equals(b); b.equals(a));
+    //@ public static invariant (\forall Mapping a, b, c; a.equals(b) && b.equals(c); a.equals(c));
+    //@ public static invariant (\forall Mapping a, b; a.equals(b); a.hashCode() == b.hashCode());
+    //@ public static invariant (\forall Mapping a, b; a.hashCode() == b.hashCode(); !a.equals(b));
+
```
unp1, line 3: Rephrasing the static invariants to instance invariants would reduce the number of quantifier nestings and may improve proof search.

key.core/pipelineTests/records/expected/02_RecordClassBuilder/CompactConstructor.java
```diff
@@ -0,0 +1,78 @@
+@javax.annotation.processing.Generated("RecordClassBuilder")
+final class Mapping extends Record {
+    //@ public invariant (\forall Mapping a, b; a.from == b.from && a.to == b.to; a.equals(b));
+    //@ public static invariant (\forall Mapping a; a.equals(a));
+    //@ public static invariant (\forall Mapping a, b; a.equals(b); b.equals(a));
+    //@ public static invariant (\forall Mapping a, b, c; a.equals(b) && b.equals(c); a.equals(c));
+    //@ public static invariant (\forall Mapping a, b; a.equals(b); a.hashCode() == b.hashCode());
+    //@ public static invariant (\forall Mapping a, b; a.hashCode() == b.hashCode(); !a.equals(b));
```
unp1: the last invariant seems not to be correct. 

a.hashCode __ != __ b.hashCode ==>  !a.equals(b)

But this is already the contraposition of the above invariant and hence redundant

key.core/pipelineTests/records/expected/02_RecordClassBuilder/InnerRecord.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @javax.annotation.processing.Generated("RecordClassBuilder")
+    final class MyRecord extends Record {
+        //@ public invariant (\forall MyRecord a, b; a.test == b.test; a.equals(b));
+        //@ public static invariant (\forall MyRecord a; a.equals(a));
+        //@ public static invariant (\forall MyRecord a, b; a.equals(b); b.equals(a));
+        //@ public static invariant (\forall MyRecord a, b, c; a.equals(b) && b.equals(c); a.equals(c));
+        //@ public static invariant (\forall MyRecord a, b; a.equals(b); a.hashCode() == b.hashCode());
+        //@ public static invariant (\forall MyRecord a, b; a.hashCode() == b.hashCode(); !a.equals(b));
+
```
unp1, line 7: similar comments as above for the invariants

key.core/pipelineTests/records/expected/02_RecordClassBuilder/CompactConstructor.java
wadoon: **For now, we could simply deactivate these invariants. For the test cases there are not needed. But as a showcase for automatically generating JML they are quite sufficient.**

But you also want to use these invariants outside of the scope of this record. 

I would still prefer class axioms, which should become Taclets. 

Also better, the introduction of named (static) invariants, which are then not automatically added, and the extension of `\static_invariant_for(obj, name)` to put them on the sequent. 

My final suggestion would be:
```
//@ public explicit static invariant_free taut1: 1+1==2;
```

The explicit modifier (present in C++) marks that the invariant is not implicitly added to `$inv` (only possible with free inv), and need to grasp explicitly.

I would also say, we should a warning or synonym on `axiom` (and treat thi as an `invariant_free`).

wadoon: I removed all invariants, but we should track the idea above.

PR #3773, opened 2026-03-16
Purge sort depending functions in favor of parametric functions

Drodt:
## Intended Change

Replace all sort depending functions from KeY and replace them with parametric functions, e.g., int::select -> select<[int]>.

This PR does not affect behavior. In particular, sequences are not (yet) made parametric and the grammar is not really modified.

`SortDependingFunction` is removed.

## Plan

* [ ] The grammar cannot be improved, because the sort depending function syntax is also used for fields, especially `java.lang.Object::<created>`
* [x] Fix all the tests that are sure to break

## Type of pull request

## Ensuring quality

    
  `.github/workflows/tests.yml`

### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/taclettranslation/assumptions/GenericTranslator.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/taclettranslation/assumptions/GenericTranslator.java
@@ -58,7 +58,7 @@ class GenericTranslator {
                 + " because there are not enough different sorts. " + generics + " " + sorts);
         }
 
-        if (list.size() > 0) {
+        if (!list.isEmpty()) {
             for (JTerm gt : list) {
                 result.add(AssumptionGenerator.quantifyTerm(gt, services));
 
@@ -85,7 +85,6 @@ class GenericTranslator {
      *         <code>term</code> is of type {@link ParametricFunctionInstance} and
      *         <code>instantiation</code> is of type {PrimitiveSort}.
      */
-
     private JTerm instantiateGeneric(JTerm term, GenericSort generic, Sort instantiation, Taclet t)
             throws IllegalArgumentException, IllegalTacletException {
         JTerm[] subTerms = new JTerm[term.arity()];
@@ -123,9 +122,6 @@ class GenericTranslator {
                   // that should have isReference-Condition, but
                   // they don't
                   // have the condition.
-
-                // TODO(DD): Extend to more complex parametric functions; ask RB what is expected
-                // here
                 if (func.getArgs().size() == 1 && func.getArgs().head().sort().equals(generic)) {
                     if (instantiation.extendsTrans(services.getJavaInfo().nullSort())) {
                         return null;


Drodt:
Per the KeY meeting: This PR is truly ready. Eventual syntax improvements are left for future PR (but should be discussed soon!)

key.core/src/main/java/de/uka/ilkd/key/pp/SelectPrinter.java
```diff
@@ -238,7 +238,7 @@ private void printArraySelect(LogicPrinter lp, JTerm heapTerm, JTerm objectTerm,
 
     /*
      * Print a select-term of the following form: T::select( ... , ... , java.lang.Object::<...>)
```
WolframPfeifer, line 240:
```suggestion
     * Print a select-term of the following form: select<[T]>( ... , ... , Classname::#fieldname)
```
I think this comment should also be updated.

key.core/src/main/java/de/uka/ilkd/key/rule/metaconstruct/ArrayBaseInstanceOf.java
WolframPfeifer, line 19:
```suggestion
 * Creates an <tt>instance<[T]>(..)</tt> term for the component type "T" of the array. The component
```
more comment updating

WolframPfeifer, line 31:
```suggestion
    * returns an <tt>instance<[G]>(term.sub(1))</tt> term for the element sort of the given array.
    * It is assumed that <tt>term.sub(0)</tt> is either a term of reference array sort or a term
    * with an <tt>exactInstance</tt> symbol as top level parameterized with the reference array sort.
```
I think this comment can be updated as well

key.ui/examples/standard_key/staticInitialisation/objectOfErroneousClass.key
```diff
@@ -41,45 +41,45 @@
 
 \problem {
      wellFormed(heap)
-   &   boolean::select(heap,
+   &   select<[boolean]>(heap,
```
WolframPfeifer, line 44: Aren't all these `classInitialized` etc. here field names? They are rewritten into the syntax of parametric sorts in this file ...

I am confused: Why are these **also** parametric functions (see HeapLDT)?

key.core/src/main/java/de/uka/ilkd/key/pp/FieldPrinter.java
```diff
@@ -149,8 +149,8 @@ protected boolean isJavaFieldConstant(JTerm fieldTerm) {
     }
 
     /*
-     * Determine whether the field constant is a generic object property. Those are surrounded by
-     * angle brackets, e.g. o.$created
+     * Determine whether the field constant is a generic object property. Those are prefixed by the
+     * separator `#` and the implicit symbol `$`, e.g. o.$created
```
WolframPfeifer, line 153:
```suggestion
     * separator `::`, the field prefix `#`, and the marker for implicit symbols `$`, e.g. `java.lang.Object::#$created`
```
I think this comment mixed in a weird way the input that is checked (that is, the field name), and the output that is printed. The latter does not play any role in this method. If at all, an output example should be given in the caller. My suggestion changes the example here such that it refers to the input.

key.core/src/main/antlr4/KeYLexer.g4
```diff
@@ -382,16 +382,15 @@ GREATEREQUAL
...
+OPENTYPEPARAMS : '<' '[';
+CLOSETYPEPARAMS : ']' '>';
 
 WS:  [ \t\n\r\u00a0]+ -> channel(HIDDEN); //U+00A0 = non breakable whitespace
 STRING_LITERAL:'"' ('\\' . | ~( '"' | '\\') )* '"' ;
 LESS: '<';
 LESSEQUAL: '<' '=' | '\u2264';
 LGUILLEMETS: '<' '<' | '«' | '‹';
 RGUILLEMETS: '>''>' | '»' | '›';
-IMPLICIT_IDENT: '<' '$'? (LETTER)+ '>' ('$lmtd')? -> type(IDENT);
```
WolframPfeifer, line 394: Until doing this review, I was not aware that we have 11 ANTLR .g4 files now (3 * lexer, 8 * parser), instead of 4 which I remember from a while ago. How many different parsers are there, and how do they work together? Is this documented or at least outlined anywhere?

In my opinion, this makes it more difficult to find token/rule declarations, and also opens possibilities for inconsistencies: Here, IMPLICIT_IDENT was removed because it was already in the ncore KeYLexer, right?

key.core/src/main/java/de/uka/ilkd/key/ldt/HeapLDT.java
```diff
@@ -178,6 +179,11 @@ public record SplitFieldName(String className, String attributeName) {
      * @return the split field name
      */
     public static @Nullable SplitFieldName trySplitFieldName(Named symbol) {
+        if (symbol instanceof ParametricFunctionInstance pfi) {
+            // e.g., $classErroneous<[A]>
```
WolframPfeifer, line 183: Is this not one of our implicit field names? Why is it used as an example for a parametric function here?

key.core/src/main/java/de/uka/ilkd/key/ldt/HeapLDT.java
```diff
@@ -428,11 +448,13 @@ public Function getFieldSymbolForPV(LocationVariable fieldPV, Services services)
             }
             final Name kind = new Name(name.toString().substring(index + 2));
 
-            SortDependingFunction firstInstance =
-                SortDependingFunction.getFirstInstance(kind, services);
-            if (firstInstance != null) {
+            final String nameWithoutFieldPrefix = kind.toString().substring(1);
```
WolframPfeifer, line 451: Instead of `substring(1)`, I think something like `substring(JavaDLFieldNames.FIELD_PREFIX.length())` would be more robust.

key.core/src/main/java/de/uka/ilkd/key/taclettranslation/assumptions/GenericTranslator.java
```diff
@@ -116,24 +117,26 @@ private JTerm instantiateGeneric(JTerm term, GenericSort generic, Sort instantia
...
+        if (term.op() instanceof ParametricFunctionInstance func) {
             try { // Try block is necessary because there are some
                   // taclets
                   // that should have isReference-Condition, but
                   // they don't
                   // have the condition.
 
-                if (func.getSortDependingOn().equals(generic)) {
+                // TODO(DD): Extend to more complex parametric functions; ask RB what is expected
+                // here
```
WolframPfeifer, line 128: Did you want to fix or remove this TODO?

key.core/src/main/java/de/uka/ilkd/key/ldt/HeapLDT.java
Drodt: You weren't there at the KeY meeting when I explained just how amazing this implementation is and was: Yes, it is an implicit field but on the logic side, these fields where always sort depending functions, e.g., `A::<classErroneous>`. It just happened to look like the field syntax.

Now that sort depending functions are removed, we use parametric functions and cannot just split.

key.core/src/main/antlr4/KeYLexer.g4
Drodt: It was removed because it is no longer needed. We have no more of these `<id>` identifiers with this PR

Drodt: For the rest, let's discuss at the next meeting

key.core/src/main/java/de/uka/ilkd/key/ldt/HeapLDT.java
Drodt: Not possible. It is a char, which has no length method. I added a comment

PR #3787, opened 2026-03-23
MetaSpace for documentation and origin

wadoon:
## Intended Change

Introduces `MetaSpace` in alignment with `Namespace<T>`. A `MetaSpace` captures the non-proof-relevant information of signature entities, i.e., documentation and origin. 

This makes the `ncore` definitions simpler and enables `MetaSpace` to be shuttable for performance.

The `MetaSpace` is filled by the `DeclarationBuilder` & Co using the `DOC_COMMENT` given in KeY files. Therefore, the description of some entities in XML becomes obsolete. The relevant descriptions were migrated. 

## Fixed side bugs

It was detected, that it is possible to register two taclets with the same name, programmatically. This was detected for generated class axioms taclets. The `IntermediateProofReplayer` now throws an exception if two Taclets with the same name are available. The earlier implementation just used one of them. 

## Type of pull request

## Ensuring quality

- I have tested the feature as follows: unit-tests


key.core/src/main/java/de/uka/ilkd/key/logic/NamespaceSet.java
```diff
@@ -18,33 +19,52 @@
...
+    public NamespaceSet(Namespace<QuantifiableVariable> varNS,
+            Namespace<Function> funcNS,
+            Namespace<Sort> sortNS,
+            Namespace<SortAlias> sortAliases,
+            Namespace<RuleSet> ruleSetNS,
+            Namespace<ParametricSortDecl> parametricSortNS,
+            Namespace<ParametricFunctionDecl> parametricFuncNS,
+            Namespace<Choice> choiceNS,
+            Namespace<IProgramVariable> programVarNS,
+            MetaSpace documentation) {
```
unp1: nothing todo with this PR, but we should find a more maintainance friendly way to register namespaces

key.ncore/src/main/java/org/key_project/logic/HasMetaSpaceKey.java
unp1, line 1: Can we name it HasMetaDoc (or something more speaking, I would like to prevent confusion with MetaConstucts, a name I would also like to retire)

key.core/src/main/java/de/uka/ilkd/key/proof/io/IntermediateProofReplayer.java
```diff
@@ -507,6 +523,9 @@ private TacletApp constructTacletApp(TacletAppIntermediate currInterm, Goal curr
         }
 
         ourApp = constructInsts(ourApp, currGoal, currInterm.getInsts(), services);
+        if (ourApp == null) {
+            System.out.println("§sdsfs");
+        }
```
unp1: Debug left over
-> for error reporting we should construct a TacletAppConstructionException here.

key.core/src/main/java/de/uka/ilkd/key/proof/TacletIndex.java
```diff
@@ -262,23 +263,16 @@ public Object clone() {
...
-        }
-
-        for (ImmutableList<NoPosTacletApp> tacletApps : succList.values()) {
-            tacletApps.forEach(result::add);
-        }
-
-        noFindList.forEach(result::add);
+    public Set<NoPosTacletApp> allNoPosTacletApps() {
+        return new HashSet<>(allNoPosTacletAppsStream().collect(Collectors.toUnmodifiableSet()));
+    }
```
unp1, line 267: this must be a LinkedHashSet otherwise we introduce indeterminsim down the line for example in the OSS where it is used

unp1: Further the current solution allocates a set twice, which is a performance issue in proof search. Maybe collect directly into the the LinkedHashSet

key.core/src/main/java/de/uka/ilkd/key/proof/io/IntermediateProofReplayer.java
```diff
@@ -456,9 +456,25 @@ private TacletApp constructTacletApp(TacletAppIntermediate currInterm, Goal curr
...
+            var possibleApps = currGoal.indexOfTaclets().allNoPosTacletAppsStream()
+                    .filter(it -> it.taclet().name().equals(nTacletName))
+                    .toList();
+
+            if (possibleApps.isEmpty()) {
+                ourApp = null;
+            } else if (possibleApps.size() == 1) {
+                ourApp = possibleApps.getFirst();
+            } else {
+                var taclets = possibleApps.stream().map(TacletApp::toString)
```
unp1: error reporting would be better if we use `MetaSpace` information and `list docs().findOrigin(taclet) (file:line)`
for each duplicate

key.core/src/main/java/de/uka/ilkd/key/util/ExceptionTools.java
```diff
@@ -357,14 +351,14 @@ private static URI parseFileName(String filename) throws MalformedURLException {
...
-            // ANTLR has 0-based column numbers
-            return new Location(parseFileName(exc.input.getSourceName()), exc.position);
-        }
-        return null;
+    private static @Nullable Location getLocation(InputMismatchException exc) {
+        var token = exc.getOffendingToken();
+
+        return token == null ? null
+                : new Location(
+                    // FIXME weigl: This does not work on Windows. Illegal characters.
```
unp1: we should fix that here
This method is actually no longer used but the public getLocation(Throwable)which does URI resolution correctly also for Windows. So I think we can remove that one?

JavaParser should not see a private method?

key.core/src/main/java/de/uka/ilkd/key/logic/MetaSpace.java
```diff
@@ -0,0 +1,89 @@
+/* This file is part of KeY - https://key-project.org
```
unp1: This seems not to contain JavaDL specific dependencies, can we move it to ncore?

key.core/src/main/java/de/uka/ilkd/key/logic/MetaSpace.java
```diff
@@ -0,0 +1,89 @@
...
+    public MetaSpace(Map<String, Object> documentation) {
+        this.metadata.putAll(documentation);
+    }
+
+    public MetaSpace(MetaSpace parent) {
+        this.parent = parent;
+    }
+
+    private @Nullable Object findMetadata(String key) {
+        return metadata.get(key);
```
unp1: the parent is not looked up if nothing has been found. When is a parent actually used?

OK, may be used for Goal local namespace, but querying them might then return falsely null

key.core/src/main/java/de/uka/ilkd/key/logic/MetaSpace.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.Map;
+import java.util.TreeMap;
+
+import org.key_project.logic.HasMeta;
+
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
+
+/// MetaSpace is a namespace for storing additional information
+///
```
unp1: Could you add 
  `MetaSpace is read-only after load`
to the class documentation. As otherwise the unsynchronized treemap can produce problems later on

key.core/src/main/java/de/uka/ilkd/key/proof/io/IntermediateProofReplayer.java
```diff
@@ -456,9 +456,25 @@ private TacletApp constructTacletApp(TacletAppIntermediate currInterm, Goal curr
...
+                        .collect(Collectors.joining("\n---\n"));
+                throw new TacletAppConstructionException(
+                    "There are more than one possible taclet available with name \"" + tacletName
+                        + "\". " +
+                        "Most three similar names are: " + taclets);
+            }
```
unp1:
```suggestion
               var details = possibleApps.stream()
        .map(app -> {
            String origin = proof.getServices().getNamespaces().docs().findOrigin(app.taclet());
            return origin == null ? app.toString() : app.taclet().name() + " @ " + origin;
        })
        .collect(Collectors.joining("\n---\n"));
throw new TacletAppConstructionException(
    "Several taclets named \"" + tacletName + "\" are available:\n" + details);
    }
```

unp1 (changes requested):

unp1:
Forgotten to write it in the review. Thanks for the work!

unp1 (changes requested):
Thanks!

Doppelt hält besser ;-) (sorry war dieses Mal der falsche Tab)

key.core/src/main/java/de/uka/ilkd/key/logic/NamespaceSet.java
wadoon: I generalized the class.

key.ncore/src/main/java/org/key_project/logic/HasMetaSpaceKey.java
wadoon: change to HasMetaSpaceKey

key.core/src/main/java/de/uka/ilkd/key/proof/TacletIndex.java
wadoon: LinkedHashMap does not prevent indeterminsim here. It will just preserve earlier indeterminsim, i.e. the order of Taclets in files. 

If you really want to prevent indeterminism you take a sorted set.

unp1 (approved):
Thanks

PR #3788, opened 2026-03-23
Options Panel in Load Dialog. Options in Recent Files

### LINKED ISSUE #3739: Make WD semantics selectable during loading

WD profile can be selected in the "Load" dialog. More utilization can be obtained, by also making WD semantics selectable. 

## Underlying problem

Changing requires changing the Taclet options. This requires re-loading of proofs, because the Taclet option are only known, after a `KeYUserProblemFile` has been read. 

For WD profile, the selection of the WD semantics is a crucial aspect directly from the start. The WD semantics are defined in [`optionsDeclarations`](https://wadoon.github.io/key-tacletdoc/optionsDeclarations.html#CATEGORY-wdOperator). 

## A general approach

1. `key.core.*` should be free of UI stuff. Therefore, we need a dynamic way that we can determine an additional supplier of UI elements.
2. The UI elements, e.g. in this case provide the combobox for the selection of WD semantics. 
3. Somehow in a dynamical fashion these information are delivered to a method in the profile, which configures the `InitConfig`. This may require re-checking the reusability of an existing profile.

## Alternatives 

Reloading of proofs. 

## Estimated effort

3h

### LINKED ISSUE #3742: Who wins: Selected Profile vs. the Profile given by KeY files.
From @WolframPfeifer in #3737 
> One drawback seems to be that you cannot change the profile in the example dialog, so you have to load example files manually with the desired profile selected. Also, I am not sure how it interacts with specifying the profile in the .key file.

## Task 

* [ ] Check the behavior for KeY, (I guess that the combobox wins). 

## Solution candidate

Add an entry to the profile selection combobox in the "Load" dialog, that states "Profile in file or Default".  Selecting this option, uses either the profile given in the KeY file, or the default profile. Hence, this would enforce the old loading behavior.

wadoon:
## Intended Change

This pull request resolves https://github.com/KeYProject/key/issues/3739

### Critical things

* Renovation of `InitConfig`. Choice selection is now stored as `Choice` in a `Map<String,Choice>`. This avoids complicated collision handling. Also default choices, and program/user/file-activated choices are kept separately. 

### UI

image

I was not able to screenshot the tool-tip. 

This PR adds the feature of having Profile-dependent options in the load proof options. 

### *Recent Files* are stored along with their options and load settings. 

I also decided to use a new file `recentFiles_v2.json` to avoid loading problems with older versions.

```json
[
    {
        "options" : null,
        "path" : "/home/weigl/work/key/key.ui/examples/standard_key/prop_log/contraposition.key",
        "profile" : null,
        "singleJava" : false
    }, {
        "options" : {
            "wdOperator" : "wdOperator:Y"
        },
        "path" : "/home/weigl/work/key/key.ui/examples/heap/simple/src/MySubclass.java",
        "profile" : "java-wd",
        "singleJava" : false
    }
]
```
## Related Issue 

This pull request resolves #3742, #3739.

## Plan

* [x] ConfigurationWriter / Configuration patches are missing 

## Type of pull request

## Ensuring quality

    
  `.github/workflows/tests.yml`

### DIFF
--- a/key.core.symbolic_execution/src/main/java/de/uka/ilkd/key/symbolic_execution/profile/SymbolicExecutionJavaProfile.java
+++ b/key.core.symbolic_execution/src/main/java/de/uka/ilkd/key/symbolic_execution/profile/SymbolicExecutionJavaProfile.java
@@ -325,4 +325,9 @@ public class SymbolicExecutionJavaProfile extends JavaProfile {
             return false;
         }
     }
+
+    @Override
+    public String displayName() {
+        return NAME;
+    }
 }
--- a/key.core.symbolic_execution/src/main/java/de/uka/ilkd/key/symbolic_execution/util/SymbolicExecutionSideProofUtil.java
+++ b/key.core.symbolic_execution/src/main/java/de/uka/ilkd/key/symbolic_execution/util/SymbolicExecutionSideProofUtil.java
@@ -27,7 +27,6 @@ import de.uka.ilkd.key.proof.mgt.RuleJustificationInfo;
 import de.uka.ilkd.key.rule.BuiltInRule;
 import de.uka.ilkd.key.rule.OneStepSimplifier;
 import de.uka.ilkd.key.rule.Taclet;
-import de.uka.ilkd.key.rule.tacletbuilder.TacletBuilder;
 import de.uka.ilkd.key.settings.ProofSettings;
 import de.uka.ilkd.key.strategy.StrategyProperties;
 import de.uka.ilkd.key.symbolic_execution.profile.SimplifyTermProfile;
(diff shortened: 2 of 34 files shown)


unp1 (changes requested):
When loading 
    key.ui/examples/firstTouch/06-BinarySearch/project.key
and selecting 
   Java Profile + Well....
I get a strange JAR file error (not running from jar but via gradlew :key.ui:run):

Caused by: java.lang.RuntimeException: java.io.FileNotFoundException: JAR entry not found in jar file

wadoon:
> When loading key.ui/examples/firstTouch/06-BinarySearch/project.key and selecting Java Profile + Well.... I get a strange JAR file error (not running from jar but via gradlew :key.ui:run):
> 
> Caused by: java.lang.RuntimeException: java.io.FileNotFoundException: JAR entry not found in jar file

Also on `main`: 

image

The error was in the construction of the `FileSystem`.


fixed on #3810

unp1:
Loading a proof with a different file is ignored:
1) Load  "examples/heap/removeDups/project.key  with "Java Profile + WD"
2) Do proof (WD PO)
3) Save proof
4) Load proof with "Java Profile (Default)"  which succeeds and applies WD

I would have expected it to fail.

unp1 (approved):
Thanks. Everything seems to work as intended.

wadoon:
History rewrite required before merge.

PR #3812, opened 2026-04-28
Fix bug in ProofManagementDialog

unp1:
## Intended Change

Fixes a bug in the proof obligation browser.

When a proof is closed, the PO browser displays the corresponding closed icon inside the "Goto Proof" button. It happened that when selecting a PO, where the goal was closed by caching, the corresponding icon was not found causing an assertion failure. This renderes the browser unusable as it did not go to the corresponding proof and one could not open the dialog again.

This PR fixes this bug.

## Type of pull request

## Ensuring quality

- I have tested the feature as follows: Closed a proof by caching, open the dialog, selected the closed PO: no exception thrown, Goto Proof worked and PO browser could be used again afterwards

Maybe we should also replace the assertion with simple logging? In the worst case the button would display the wrong close icon.

### DIFF
--- a/key.ui/src/main/java/de/uka/ilkd/key/gui/ProofManagementDialog.java
+++ b/key.ui/src/main/java/de/uka/ilkd/key/gui/ProofManagementDialog.java
@@ -132,7 +132,10 @@ public final class ProofManagementDialog extends JDialog {
                     } else if (ps.getProofClosedByCache()) {
                         label.setIcon(KEY_CACHED_CLOSED);
                     } else {
-                        assert ps.getProofOpen();
+                        if (!ps.getProofOpen()) {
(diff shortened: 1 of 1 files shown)


unp1:
No idea why the formatting check failed. I did not change code in any of the modules mentioned besides key.ui and here locally a spotlessApply does not reformat anything.

WolframPfeifer (approved):
I re-run the formatter task and it succeeds, I think it just crashed somehow (also, the log didn't show any formatting violation) ...

Regarding the assertion, I agree that we should remove it, since it crashes the GUI. I think this is a side effect from enabling the assertions by default. Maybe changing the if-else to enhanced switch could solve the problem for the future, since it has to cover all cases of the enum then.

PR #3819, opened 2026-06-06
Deployment to Maven Central SNAPSHOT

wadoon:
## Related Issue

This pull request allows the nightly_deployer to deploy the Maven CENTRAL Snapshot. 

Currently, my tokens are added as secrets. Job step was tested here: https://github.com/KeYProject/key/actions/runs/27469080733

In the current version, we deploy to [https://git.key-project.org] only. After the PR, we deploy additionally to Maven Central. 

As Maven Central/sonatype decided to diverge from the Maven standard and use a special API, special plugins are required. 

## Intended Change

Weekly deployment of SNAPSHOT to Maven Central. 

## Type of pull request


unp1 (approved):
Thanks!

PR #3832, opened 2026-06-15
Make semicolon after rule optional

Drodt:
## Intended Change

This PR makes the semicolon after rules optional in the grammar. It is one of most common and annoying syntax choices KeY beginners run into and I see no reason to require the semicolon.

## Type of pull request

- Other: Grammar change

## Ensuring quality

    
- I have tested the feature as follows: Removed some semicolons from they tesruley.key file


unp1 (approved):
Thanks. No longer missing semicolons 👍

PR #3837, opened 2026-06-17
Performance (4/5): reuse rule-app cost across re-expansion (+ age as a cost term)

unp1:
_This PR is **4/5** of a performance series that prepares a clearer match → evaluate → apply pipeline before larger optimisations; it builds on the compiled matcher (1/5, #3831)._

📖 Developer docs: [Performance Optimizations (3.1)](https://keyproject.github.io/key-docs/devel/PerformanceOptimizations/#strategy-evaluation-cost-reuse-and-age)

## Intended Change

The strategy re-costs rule-app containers far more often than necessary. On every peek round, `TacletAppContainer.createFurtherApps` calls the full `Strategy.computeCost` again for a base app that has not changed — and `computeCost` (the feature-tree evaluation) is the dominant CPU cost of automode. This PR carries a container's cost forward across re-expansion instead of recomputing it, for the taclets where that is provably sound. It contains two coupled pieces:

**1. Age as a first-class container-level cost term.** Previously the goal-age contribution was baked into the stored cost (via `AgeFeature`), which meant the stored cost could not be reused across rounds (age changes every round). Age is lifted out: a container stores its **age-free cost**, and `RuleAppContainer#withAge` re-adds the current goal age when the container is (re-)inserted. This is a clean refactor (the `AgeFeature` leaf is removed) and is the enabler for reuse.

**2. Cost reuse across re-expansion.** When a taclet's cost is a pure function of the app + its find-subterm (plus the always-refreshed age term and the `NonDuplicateApp` vetoes), the stored age-free cost is carried forward verbatim instead of recomputed. Eligibility is decided by a **sound-by-construction, annotation-driven classification** (`@CostLocal` / `@CostNonLocal`): a taclet is eligible only if every feature reachable in its cost bindings is explicitly marked local; everything unannotated is treated as non-local (the safe default — forgetting an annotation costs performance, never soundness). The classification is generator-aware: a composite summing over a sequent-scanning generator stays non-local. The `NonDuplicateApp`-family vetoes that can still fire are re-checked on every reuse, so an app that became a duplicate is still dropped.

A small related determinism fix is included: `introductionTime` no longer caches the "not introduced yet" sentinel (`-1`), which could otherwise freeze a value depending on query order and make term ordering subtly non-deterministic.

```mermaid
flowchart LR
  A["re-expand base"] --> B{"cost-local?"}
  B -- no --> C["full computeCost"]
  B -- yes --> D["reuse cost, re-add age"]
```

## Type of pull request
- Refactoring (age-as-cost-term) + performance feature (cost reuse)

## Performance

Measured on the 6-problem `perfTest` set (median of 3 runs). Cost reuse is byte-identical on the perfTest/perfValidation corpus, so node counts are unchanged; the win is reduced `computeCost` time. A `-Dkey.strategy.costReuse.verify` development flag (default off) recomputes the cost and warns on any mismatch.

| Problem | main time (ms) | this PR (ms) | speedup |
|---|--:|--:|--:|
| symmArray | 21396 | 19301 | 1.11× |
| gemplusDecimal/add | 11474 | 11759 | 0.98× |
| ArrayList.remove.1 | 3629 | 3460 | 1.05× |
| SimplifiedLinkedList.remove | 28903 | 26330 | 1.10× |
| Saddleback_search | 23510 | 21964 | 1.07× |
| coincidence_count/project | 4935 | 4678 | 1.05× |
| **Total** | **93847** | **87492** | **1.07×** |

Node counts identical to `main` (byte-identical proofs); the single sub-1.0 entry (gemplusDecimal/add, 0.98×) is run-to-run noise. The win is reduced `computeCost` time on cost-bound proofs.

## Ensuring quality
- Sound-by-construction (default non-local); `costReuse.verify` cross-checks against full recompute.
- Byte-identical on perfTest/perfValidation; full RAP closes the same goals.
- Active by default (the previous `-Dkey.strategy.costReuse` gate is removed).

---
📖 Conceptual overview in the developer docs: [Performance Optimizations (3.1)](https://keyproject.github.io/key-docs/devel/PerformanceOptimizations/#strategy-evaluation-cost-reuse-and-age)


key.core/src/main/java/de/uka/ilkd/key/strategy/feature/NoSelfApplicationFeature.java
```diff
@@ -9,13 +9,15 @@
...
 import org.key_project.prover.strategy.costbased.MutableState;
+import org.key_project.prover.strategy.costbased.feature.CostLocal;
 import org.key_project.prover.strategy.costbased.feature.Feature;
 import org.key_project.util.collection.ImmutableList;
 
 /**
  * This feature checks that the position of application is not contained in the if-formulas. If the
  * rule application is admissible, zero is returned.
  */
+@CostLocal
```
Drodt, line 20: It sounds to me that this is _not_ local: It considers not just the find term of the rule app.

unp1: Classifying it as cost-local is correct also the name is a bit misleading

- Actually a feature is reuse safe if it does not cause a different value of the costs of a taclet app, if the same app (with the same not changed instantiations) is evaluated at a later point again. 

  This is always true for features whose value is purely determined by the formula against which find is matched, hence, the name CostLocal. 
  
  But there are also other cases where this is the case and NoSelfApplication is such a feature.
  Assume replace_known_right
  \assumes(==> b)
  \find (b)
  ...
  
  when it is matched first an incomplete rule app container is created
  rAppCont_1 (ruleApp, cost:c + 0) (0 being the cost of NoSelfApp), but the assumes sequent has not yet been matched.
  When it is picked as next rule to be evaluated the following steps are performed:
    1. revaluate base cost of rAppCont_1  (that are only the costs w.r.t. find) and put the incomplete container back into the queue (for later expansion if new formulas arrive on the sequent)
    2. the ruleApp is matched against every newly added sequent formula (since the last time) 
       for each successfull match a new container is created and the _full_ costs are recomputed (no cost reuse between different containers)

  Also NoSelfApp does depend on other sequent formulas, it does not change the cost for an existing rule app container when evaluating its costs again later:
  1. incomplete (assumes not matched) -> cost is constant 0
  2. complete -> the container does not change and as NoSelfApp only checks identity no changes to formulas can make its value different

Drodt (approved):
Thanks!

Issue #3847, opened 2026-06-19
Proof Slicing Broken


WolframPfeifer:
## Description

Proof Slicing is broken on the current main branch

## Reproducible
  
always

### Steps to reproduce 

1. Load the SumAndMax example.
2. Press the green button to run the auto mode and wait until the proof is closed.
3. Run the slicing: "Proof Slicing" tab --> "Slice proof"

KeY throws an exception:
```
java.lang.IllegalStateException: failed to find new formula
	at de.uka.ilkd.key.proof.replay.AbstractProofReplayer.constructBuiltinApp(AbstractProofReplayer.java:153)
	at de.uka.ilkd.key.proof.replay.AbstractProofReplayer.reApplyRuleApp(AbstractProofReplayer.java:84)
	at org.key_project.slicing.SlicingProofReplayer.slice(SlicingProofReplayer.java:201)
	at org.key_project.slicing.ui.SlicingLeftPanel.lambda$sliceProof$8(SlicingLeftPanel.java:447)
	at org.key_project.slicing.util.GenericWorker.doInBackground(GenericWorker.java:48)
	at org.key_project.slicing.util.GenericWorker.doInBackground(GenericWorker.java:16)
	at java.desktop/javax.swing.SwingWorker$1.call(SwingWorker.java:305)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
	at java.desktop/javax.swing.SwingWorker.run(SwingWorker.java:342)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at java.base/java.lang.Thread.run(Thread.java:1583)
```

### Additional information

For simpler examples (tested with AuntAgatha), the slicing works.

WolframPfeifer:
@FliegendeWurst You already looked into that. Do you have any insights why this happens?

WolframPfeifer:
This is indeed a duplicate of #3834.

PR #3848, opened 2026-06-19
Make new name creation only dependent on goal namespaces not on global counter

### LINKED ISSUE #3834: Loop rule creates different variable name in replay (breaking slicing)
## Description

The Loop Scope rule saves modified local variables in an extra variable with the suffix `_before#X`, where X is a value taken from a proof-global counter.
This means that pruning and re-applying the rule will result in a different number.
It also means that, due to some performance optimisations in the replay, the number is different when reloading the proof.

Since the slicing code relies on the reloading mechanism creating an equivalent formula, it breaks if there is a loop in the program (and this mechanism is triggered).

## Reproducible
  
always

### Steps to reproduce 

1. Load SumAndMax
2. Run automode
3. Observe the first proof step in "Invariant Preserved and Used": it contains `k_before#1`
4. Save and reload the proof
5. Observe the same proof step: it contains `k_before#0`

exected: variable names stay the same
actual: they differ

### Additional information

The problem is in the variable condition that implements the transformation

https://github.com/KeYProject/key/blob/0ded37352caf2fe8c6abf2495252f6125373ac71/key.core/src/main/java/de/uka/ilkd/key/rule/conditions/NewLocalVarsCondition.java#L86-L87

This code is run when updating the taclet index. When doing the proof for the first time, it is run twice (for an OSS step and the Loop Scope step). When replaying the proof it seems the taclet index update is optimised in some way to avoid running it twice.

---

* Commit: any since #3525

unp1:
## Related Issue

This pull request resolves #3834, #3847 hopefully
.

## Intended Change

New names are allocated only based on the goals namespace and the already reserbed names
(the varcond creates more than one name, so we have to consider them to not cause a clash).

This makes it independent of double checking (which caused before this PR a global counter variable to be modifed, and hence the breaking side-effect)
## Type of pull request

## Ensuring quality

    
### DIFF
--- a/.github/temp_pr_body.md
+++ b/.github/temp_pr_body.md
@@ -1,0 +1,44 @@
+## Description: Make Packages Sealable
+
+### Overview
+This pull request refactors the KeY project's package structure to enable Java package sealing—a security feature that prevents unauthorized classes from being added to critical packages. The changes involve extensive package reorganization and restructuring across 470 files.
+
+### What are Sealable Packages and Why Are They Useful?
+
+**Sealable packages** (also called sealed packages) are a Java security mechanism where you can seal a package within a JAR file to prevent other code from adding classes to that package. This is valuable because:
+
+1. **Security Hardening**: Prevents attackers from injecting malicious classes into core packages by adding them to the classpath or loading them dynamically
+2. **API Integrity**: Protects internal package structures—external code cannot define new classes in sealed packages, ensuring the API remains stable and predictable
+3. **Modularity**: Clearly demarcates which packages are public/stable versus internal, reducing accidental coupling to implementation details
+4. **Version Safety**: When deploying updates, sealed packages guarantee that no legacy or conflicting versions of package-internal classes will be loaded from other JARs
+
+### Main Changes in This PR
+
+The PR makes packages sealable by:
+
+1. **Package Reorganization**: Moving files from overly broad packages into more specific, focused packages:
+   - GUI-related classes moved to `de.uka.ilkd.key.ui.gui.*`
+   - Core UI infrastructure moved to `de.uka.ilkd.key.ui.core.*`
+   - Symbolic execution utilities organized under `de.uka.ilkd.key.symbolic_execution.*`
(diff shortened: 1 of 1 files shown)


FliegendeWurst (approved):
The fix looks good to me.

Though I could not come up with an example where the freeName counter logic is applied. I tried naming a variable `k_before` and nesting two loops which assign the same variable. In both cases the `baseName` is used as is.

key.core/src/main/java/de/uka/ilkd/key/rule/conditions/NewLocalVarsCondition.java
```diff
@@ -106,6 +119,28 @@ public MatchResultInfo check(SchemaVariable var, SyntaxElement instCandidate,
...
+     * {@link de.uka.ilkd.key.logic.VariableNamer#getTemporaryNameProposal(String)}, this does not
+     * depend on a mutable proof-global counter, so it stays stable across proof reloads and
+     * prune/reapply cycles (issue #3834).
+     *
+     * @param basename the desired base name, e.g. {@code "k_before"}
+     * @param services the services whose namespaces are consulted for collisions
+     * @param reserved names already used within the current application; the chosen name is added
+     * @return a fresh, collision-free name
+     */
+    private static ProgramElementName uniqueBeforeName(String basename, Services services,
```
WolframPfeifer, line 136: I have the feeling that this method (or a similar one) should be the default over `VariableNamer.getTemporaryNameProposal(String)`. We had troubles with these proof-global counters every once in a while as far as I remember, and deterministic names only depending on the current proof state (and not, for instance, the history of rule applications and pruning operations) really sound like a good idea ...

WolframPfeifer (approved):
I tested it a bit in the GUI, and it really looks like it fixes the bug. Also, the test case is a good idea. Thank you!

key.core/src/main/java/de/uka/ilkd/key/rule/conditions/NewLocalVarsCondition.java
unp1: I agree. To not forget about it, I turned your suggestion to an issue #3851

PR #3850, opened 2026-06-19
Fix "Rule of same name exists already" error after pruning (rare but occuring cases after mass pruning with try-closeable-goals)

unp1:
After using the Full-Auto-Macro where some proof goals remained open,  I got an error "Rule of same name exists already ", when starting the proof search again. 

The problem is a slight oversight in the proof pruner which misses to unregistered locally introduced rules at leaves. This fixes this rare but happening error. It also fixes a separate error that if at the cutting point a new taclet was registered, then this got lost.
 
## Intended Change

Fix spurious double registration bug for taclets in \addrules

## Type of pull request

## Ensuring quality

    
- I added new test case(s) for new functionality. The added test fails on main and suceeds on this PR
 

Drodt (approved):
Thanks!

PR #3857, opened 2026-06-22
automode to choose from a dropdown list

mattulbrich:
## Related Issue

n/a

## Intended Change

Currently, only the full automation is on a button in the toolbar. But the autopilots have reached a similar degree of relevance and will even more with the  JML proof scripts.

## Plan

* [x] Implement a dropdown and adapt calling automation
* [x] Documented

## Type of pull request

*New feature (non-breaking change which adds functionality)*

A small UI feature that we had discussed already a few times.

## Ensuring quality
  
Manually

Co-Authored by: kit.qwen3.5-397b-A17b @ KIT, Local AI
     

mattulbrich:
Drafting for now, since it was unstable so far and can have one more round of UI testing.

mattulbrich:
As soon as the proof scripts are on the main branch, another button for the script-aware autopilot should be added.

Drodt:
Can we maybe workshop a better name for the regular automation? "Full Automation" vs. "Full Auto Pilot" is not exactly user friendly.

Perhaps "Apply Strategy"?

key.ui/src/main/java/de/uka/ilkd/key/gui/fonticons/IconFactory.java
```diff
@@ -366,6 +367,102 @@ public static Icon autoModeStopLogo(int size) {
...
+     * <p>
+     * This is the legacy default automation mode that starts/stops automated proof search.
+     * The icon shows a green play button with an "R" overlay.
+     * </p>
+     *
+     * @param size the size of the icon in pixels
+     * @return the automation run icon
+     */
+    public static Icon automationRunLogo(int size) {
+        return automationWithOverlay(size, "R");
```
Drodt: The version with "R" is not shown when I test it

unp1:
Great idea!

Concerning:
> Can we maybe workshop a better name for the regular automation? "Full Automation" vs. "Full Auto Pilot" is not exactly user friendly.
> 
> Perhaps "Apply Strategy"?
Just a suggestion:

Maybe sth along the lines

Full Automation (Efficient)                  <-- former Apply Strategy
Full Automation (Structured)              <-- former Full Auto Pilot

One could then have small tooltip explaining 
- Narrow branching, less memory usage, difficult too understand
- Produces a human-centric proof structure, open goals can be understood easily

key.ui/src/main/java/de/uka/ilkd/key/gui/fonticons/IconFactory.java
mattulbrich: Left over from the initial version; not meant to be shown.

mattulbrich: removed the method.

mattulbrich:
I like your UI-concerns. Thanks for rasing them.

To unify the concepts: "Apply strategy" should be part of the macro context menu.
The user should not be bothered with the details.

How about the following?
- Full Automation
- Structured Automation
- Structured Automation (Prep.)
- Script-aware Automation

mattulbrich:
I have updated the UX according to our plan.

key.ui/src/main/java/de/uka/ilkd/key/gui/fonticons/IconFactory.java
```diff
@@ -366,6 +367,88 @@ public static Icon autoModeStopLogo(int size) {
...
+     * invariant definitions, and try to close all proof obligations.
+     * The icon shows a green play button with an "F" overlay.
+     * </p>
+     *
+     * @param size the size of the icon in pixels
+     * @return the full auto pilot icon
+     * @see de.uka.ilkd.key.macros.FullAutoPilotProofMacro
+     */
+    public static Icon automationFullPilotLogo(int size) {
+        return automationWithOverlay(size, "F");
```
Drodt: I hate to be nit-picky but  "S" makes more sense for the structural auto pilot, imo
```suggestion
        return automationWithOverlay(size, "S");
```

Drodt (changes requested):

key.ui/src/main/java/de/uka/ilkd/key/gui/fonticons/IconFactory.java
mattulbrich: Thanks. I understand.

However, ... I wanted to keep "S" for scripted once this is merged soon.

Perhaps another lettre? 
Should there be a letter for the "normal" auto, too? "A" ?
"P" for prep seems to work.

Drodt: Fair enough

Drodt (approved):

key.ui/src/main/java/de/uka/ilkd/key/gui/fonticons/IconFactory.java
mattulbrich: How about ... :question: 

A -> Auto
S -> Structured
P -> Prep
J -> JML Scripts

Drodt: Sounds good!

PR #3858, opened 2026-06-22
Fix pruning of cached branches

### LINKED ISSUE #3846: Fix Interaction of Proof Caching and Pruning
## Description

Pruning goals that are closed by reference (proof caching) leaves the goals in a broken state. Probably the ProofPruner needs to be adapted to handle cached goals and correctly reopen them.

## Reproducible
  
always

### Steps to reproduce 

1. Load the SumAndMax Example, close the proof by Full Auto Pilot macro.
2. Do the exact same again, there are many cached goals referring to nodes in the first proof.
3. Use "Prune Proof" on one of the cached goals.

The goal is in a broken state afterwards, displaying as `UNKNOWN GOAL KIND (Probably a bug)`.

FliegendeWurst:
## Related Issue

This pull request resolves #3846.

## Intended Change

Two bugfixes

1. Fix proof caching info window to be centered (tested on Linux). Is now centered with the same method as other KeY windows.
2. Fix pruning of cached branches.

The second part is done almost exactly like in #3440. I tested both the minimal prune (prune directly at the cached node), and a larger (multi-branch) prune.
I added a `refreshGoal` call, in case it helps.

## Type of pull request

## Ensuring quality

- I have tested the feature as follows: manually

### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/proof/ProofPruner.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/proof/ProofPruner.java
@@ -7,6 +7,7 @@ import java.util.*;
 import javax.swing.*;
 
 import de.uka.ilkd.key.proof.init.InitConfig;
+import de.uka.ilkd.key.proof.reference.ClosedBy;
 import de.uka.ilkd.key.rule.NoPosTacletApp;
 import de.uka.ilkd.key.rule.merge.MergePartner;
 import de.uka.ilkd.key.rule.merge.MergeRuleBuiltInRuleApp;
@@ -35,6 +36,16 @@ class ProofPruner {
      * @return the subtrees whose common root was the given {@code cuttingPoint}
      */
     public ImmutableList<Node> prune(final Node cuttingPoint) {
+        // special case: prune cached goal = only need to re-open the final node
+        ClosedBy cby = cuttingPoint.lookup(ClosedBy.class);
+        Goal goal = proof.getClosedGoal(cuttingPoint);
+        if (cby != null && goal != null) {
+            cuttingPoint.deregister(cby, ClosedBy.class);
+            proof.reOpenGoal(goal);
+            refreshGoal(goal, cuttingPoint);
+            return ImmutableList.of(cuttingPoint);
+        }
+
         // there is only one leaf containing an open goal that is interesting for pruning the
         // subtree of <code>node</code>, namely the first leave that is found by a breadth
         // first search.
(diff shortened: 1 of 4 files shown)


WolframPfeifer (approved):
The call to refreshGoal() indeed seems to have the desired effect, and in my manual tests the pruning seems to work as expected. Thanks!

One thing that I noticed: If you prune a goal closed via cache directly, and then open and close the "Proof Caching" dialog from the status bar, it is closed directly again. Apparently closing the dialog triggers a cache check. I don't think this behavior is a problem, just a little bit unexpected maybe, and would keep it like this for now.

WolframPfeifer:
Weird: In the test cases, the list of open goals again has a `null` entry.

WolframPfeifer:
@FliegendeWurst The newly implemented functionality for pruning is inconsistent with the test in line 79-81 of TestReferenceSearcher. Could you have a look?

key.core/src/main/java/de/uka/ilkd/key/proof/ProofPruner.java
```diff
@@ -35,6 +36,16 @@ class ProofPruner {
...
     public ImmutableList<Node> prune(final Node cuttingPoint) {
+        // special case: prune cached goal = only need to re-open the final node
+        ClosedBy cby = cuttingPoint.lookup(ClosedBy.class);
+        if (cby != null) {
+            cuttingPoint.deregister(cby, ClosedBy.class);
+            Goal goal = proof.getClosedGoal(cuttingPoint);
+            proof.reOpenGoal(goal);
+            refreshGoal(goal, cuttingPoint);
+            return ImmutableList.of(cuttingPoint);
+        }
```
unp1, line 52: ClosedBy cby = cuttingPoint.lookup(ClosedBy.class);
Goal cachedGoal = cby != null ? proof.getClosedGoal(cuttingPoint) : null;
if (cachedGoal != null) {
    cuttingPoint.deregister(cby, ClosedBy.class);
    proof.reOpenGoal(cachedGoal);
    refreshGoal(cachedGoal, cuttingPoint);
    return ImmutableList.of(cuttingPoint);
}

makes it more robust and fix the test cases.

FliegendeWurst: Thanks. It works.

WolframPfeifer: This fixes the test and I thought about it as well. However, I was wondering: Why would a `ClosedBy` be registered at a node, when the node does not correspond to a goal? Isn't that a situation that should not occur (I know it does in the test at the moment)?

FliegendeWurst: I think we could swap these two lines in the test:
```
n55.register(n55Close, ClosedBy.class);
p.pruneProof(n55);
```
> Why would a ClosedBy be registered at a node, when the node does not correspond to a goal? Isn't that a situation that should not occur

I think so too. In normal operation of the extension, registering a ClosedBy is always combined with closing the goal.

WolframPfeifer: Ok, I fixed the test case (there are two places where two lines have to be swapped), and also made the ProofPruner throw an IllegalStateException to indicate the situation with a registered ClosedBy, but no corresponding goal occurs. If the tests run through, I am happy to merge.

PR #3860, opened 2026-06-24
Make cross-platform path/URI/charset handling in load/save more robust

unp1:
## Intended Change

While auditing KeY for cross-platform (Linux/macOS/Windows) robustness of
loading and saving, several latent path/URI/charset issues surfaced. This PR
fixes them. Each issue is a separate commit.

1. **`IOUtil.toURI` corrupted `+` in file paths (all platforms).**
   The central `URL → File` converter decoded the URL path with
   `URLDecoder`, whose `application/x-www-form-urlencoded` semantics turn a
   literal `+` into a space. URL paths never encode a space as `+` (they use
   `%20`), so any file or directory whose name contains `+` (e.g. `C++Model/`,
   `a+b.key`) was silently mangled. The `+` is now protected before decoding.

2. **Loader did not normalize backslashes — legacy Windows proofs failed on POSIX.**
   `KeYFile.readBootClassPath`/`readClassPath`/`readJavaPath` parsed the stored
   path strings directly with `Paths.get`. Newer proofs are written with forward
   slashes, but proofs saved by older KeY versions on Windows contain
   backslashes; on POSIX a backslash is a valid file-name character, so the path
   was not split into segments and could not be resolved. The three readers now
   normalize separators, matching what `IncludeFinder.addInclude` already does
   for `\include`.

3. **`UrlRuleSource` reported the wrong file size on Windows / for spaced paths.**
   The constructor used `new File(url.getFile()).length()`; `url.getFile()` is
   not percent-decoded and keeps the leading `/` of `file:/C:/…` on Windows, so
   for paths with spaces (`%20`) or on Windows it pointed at a non-existent file
   and returned `0`. It now resolves via `Paths.get(url.toURI())`, matching the
   class's own `file()` method, with a stream-counting fallback.

4. **`Filenames.makeFilenameRelative` was not cross-platform safe.**
   `Path.relativize` throws `IllegalArgumentException` when the two paths have
   different roots (e.g. different drive letters on Windows), and the result was
   returned with native separators. It now guards the differing-root case and
   normalizes output to forward slashes, mirroring `IOUtil.safePathRelativeTo`.

5. **Proof-independent settings JSON was written with the platform-default charset.**
   The settings file is read as UTF-8 (`CharStreams.fromPath`) but was written
   with a plain `FileWriter`. On a JDK whose default charset is not UTF-8 (older
   JDKs on Windows), non-ASCII content (umlauts in bookmarked paths, user names)
   was written in one encoding and parsed back in another. The writer is now
   explicitly UTF-8.

## Type of pull request

## Ensuring quality

- I have tested the feature as follows: The full test suites are left to CI as I do not have access to a local windows machine

This PR has been done with AI tooling.

### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/proof/io/KeYFile.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/proof/io/KeYFile.java
@@ -228,6 +228,22 @@ public class KeYFile implements EnvInput {
     }
 
 
+    /**
+     * Normalizes a path string read from a KeY/proof file so that it can be parsed by
+     * {@link Paths#get(String, String...)} on the current platform. KeY now always writes path
+     * separators as forward slashes, but proofs saved by older versions on Windows may contain
+     * backslashes. On POSIX a backslash is a valid file-name character, so such a path would not be
+     * split into segments and could not be resolved. Converting backslashes to forward slashes is
+     * safe on all platforms (Windows accepts '/' as a separator) and keeps these legacy proofs
+     * loadable.
+     *
+     * @param path the raw path string as stored in the file
+     * @return the path string with backslashes replaced by forward slashes
+     */
+    private static String normalizeStoredPath(String path) {
+        return path.replace('\\', '/');
+    }
+
     @Override
     public Path readBootClassPath() {
         ProblemInformation pi = getProblemInformation();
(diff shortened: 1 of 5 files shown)


Drodt (approved):
Thanks!

PR #3861, opened 2026-06-24
LogicPrinter: reduce per-subterm overhead and harden against unbalanced blocks

unp1:
## Intended Change

A focused review of `LogicPrinter` for two aspects — printing performance and
the intermittent "unbalanced blocks" failures — surfaced three issues. Each is
addressed in its own commit.

1. **Performance: redundant lookups in the recursive hot path.**
   `LogicPrinter.printTerm` is invoked once per subterm of a sequent. For a
   labeled term it recomputed the same values up to three times:
   `notationInfo.getNotation(t.op())` (a map lookup with a fallback cascade) and
   `getVisibleTermLabels(t)`. The latter is overridden in
   `SequentViewLogicPrinter` to allocate a `LinkedList`, scan every label for
   visibility and build a fresh `ImmutableArray` on each call, so the repeated
   evaluation produced identical results at real allocation cost. The notation
   and the parenthesization decision are now computed once; the visible labels
   are still only consulted when the term has labels (preserving the previous
   short-circuit). Behaviour is unchanged.

2. **Balanced blocks: reset the layouter on a printing failure.**
   The top-level sequent printing methods caught only
   `UnbalancedBlocksException` and rewrapped it. A primary failure deep in the
   recursive term printing (e.g. a `NullPointerException`) instead unwinds past
   the pending `end()` calls and leaves the `Layouter`'s block stack non-empty.
   Because a `LogicPrinter` (and its layouter) can be reused, the next print
   then fails with a confusing `UnbalancedBlocksException` that masks the real
   cause, the intermittent problem. `printFilteredSequent` and `printSequent`
   now `reset()` the layouter (fresh clone) on any `RuntimeException`/`Error`
   before propagating. The nested helper `printSequentInExistingBlock` is left
   untouched: it runs inside an outer block, so resetting there would break the
   caller's matching `end()`.

3. **Balanced blocks: read the printed sequent only on success in `GoalList`.**
   `GoalList.seqToString` read `sp.result()` in a `finally` block, i.e. also
   when `printSequent` had thrown. `result()` returns the raw backend buffer
   without flushing pending layout tokens, so on a failed print this cached a
   silently truncated/empty sequent. The result extraction is moved onto the
   success path, with an explicit placeholder cached on failure.

## Type of pull request

## Ensuring quality

- I have tested the feature as follows: `:key.core` and `:key.ui` compile;
  `spotlessCheck` passes on both modules; the pretty-printer unit tests
  (`LogicPrinterTest`, `PrettyPrinterTest`, `PrettyPrinterRoundtripTest`,
  `FinalPrinterTest`) all pass (42 tests), confirming the `printTerm` refactor
  is behaviour-preserving. The full test suites are left to CI.
- I have checked that runtime performance has not deteriorated (the printTerm
  change removes redundant per-subterm lookups/allocations).

PR created with AI tooling support.

### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/pp/LogicPrinter.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/pp/LogicPrinter.java
@@ -788,13 +788,19 @@ public class LogicPrinter {
             layouter.startTerm(0);
             layouter.print(notationInfo.getAbbrevMap().getAbbrev(t));
         } else {
-            if (t.hasLabels() && !getVisibleTermLabels(t).isEmpty() && notationInfo
-                    .getNotation(t.op()).getPriority() < NotationInfo.PRIORITY_ATOM) {
+            // printTerm is the central recursive method (called once per subterm), so the
+            // notation and the visible-label set are looked up once here instead of up to three
+            // times. getVisibleTermLabels is only consulted when the term actually has labels,
+            // preserving the previous short-circuit (its SequentViewLogicPrinter override
+            // allocates and filters on every call).
+            final Notation notation = notationInfo.getNotation(t.op());
+            final boolean parens = t.hasLabels() && !getVisibleTermLabels(t).isEmpty()
+                    && notation.getPriority() < NotationInfo.PRIORITY_ATOM;
+            if (parens) {
                 layouter.print("(");
             }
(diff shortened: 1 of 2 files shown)


Drodt (approved):
Thanks!

PR #3863, opened 2026-06-26
Make proof-tree painting cheap on large proofs 

unp1:
## Intended Change

`ProofTreeView.ProofRenderer.renderBranch` decides the rare "linked-folder" icon by running a
breadth-first search over the branch's **entire subtree on every repaint**, calling the
linear-scan `Proof.getOpenGoal` on each visited node.  That is O(subtree × open goals) per open branch,
per paint no matter **whether** the merge rule is used or not. On large proofs (tens of thousands of nodes, many open goals) this turns tree painting
into billions of operations and freezes the EDT. A thread dump on a ~300k-node proof showed the
EDT RUNNABLE with 70+s of CPU in `renderBranch`.

The fix iterates the proof's **open goals** instead of the whole subtree (far fewer of them),
short-circuited by the cheap `isLinked()` test and walking ancestors only for linked goals. A
branch is still marked linked iff it contains a linked open goal — same result, a tiny fraction of
the cost. `renderLeaf` additionally skips the `getOpenGoal` scan for closed leaves (a closed leaf
has no open goal). Behaviour is unchanged; only the cost of computing the icon changes, and large
proof trees now paint smoothly.

## Type of pull request

## Ensuring quality

- I have tested the feature as follows: the change is behaviour-preserving (the linked-folder icon
  still appears for branches that contain a linked goal); on a >300k-node proof the tree now paints
  smoothly instead of freezing the GUI.
- I have checked that runtime performance has not deteriorated (it is a performance improvement).

Found via a thread dump while working on the multi-core prover branch (#3842), which makes such
large proofs easy to reach, but the fix is stock-KeY and independent of it — factored out here as a
standalone change.

PR created with AI tooling

### DIFF
--- a/key.ui/src/main/java/de/uka/ilkd/key/gui/prooftree/ProofTreeView.java
+++ b/key.ui/src/main/java/de/uka/ilkd/key/gui/prooftree/ProofTreeView.java
@@ -1238,38 +1238,52 @@ public class ProofTreeView extends JPanel implements TabPanel {
             if (node.isClosed()) {
                 // all goals below this node are closed
                 style.icon = IconFactory.provedFolderIcon(iconHeight);
-            } else {
-                // Find leaf goal for node and check whether this is a linked goal.
+            } else if (hasLinkedGoalBelow(node.getNode())) {
+                // Some open goal below this branch is linked -> linked-folder icon.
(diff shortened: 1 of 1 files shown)


Drodt (approved):
Thanks!

Issue #3570, opened 2025-02-23
Real literals can now longer be parsed


mattulbrich:
## Description

Real number literals like `1.0r` cannot be used in KeY any more.

## Reproducible

always

### Steps to reproduce 

Load a file containing `\problem { 1.0r = 1.0r }`

It yields the following error:
*The given float literal does not have a suffix. This is essential to determine its exact meaning. You probably want to add 'r' as a suffix. at /tmp/a.key:1:12*

### Additional information

The code responsible reads:
```
if(lastChar == 'R' || lastChar == 'r') {
            semanticError(ctx, "The given float literal does not have a suffix. This is essential to determine its exact meaning. You probably want to add 'r' as a suffix.");
}
```
introduced in ae9fe0167a3042977660760ce324b29843eaddd2

---

* Commit: dfc5b506e3014877cb6d1e4dd6df85696b25ddb8

mattulbrich:
Well, ... this is not really important: Without the suffix "r", a `throw new Error("not yet implemented");` is hit anyways.

PR #3865, opened 2026-06-27
Implement \real literals (#3570); fix LogicPrinter.quickPrintSemisequent (#243)

### LINKED ISSUE #243: No output of last term when using LogicPrinter.printSemisequent
This issue was created at [git.key-project.org](https://git.key-project.org/key/key/-/issues/243) where the discussions are preserved.
    
----
    
* Mantis: [MT-1584](http://i12www.ira.uka.de/~klebanov/mantis/view.php?id=1584)
* Submitted on: 2016-04-01 by (at)weigl
* Updated: 2016-04-01

### Description

> If you print a Semisequent through LogicPrinter.{printSemisequent,quickPrintSemisequent} the last term is not written. 
> Constructing a Sequent with the same Semisequent and calling LogicPrinter.quickPrintSequent results in the correct output. 
> 
> The attached program produces the following output:
> 
> ---8<----------------------
> [false,p]
> false,
> 
> false, p ==> false, p
> --------------------->8----
> 
> You should see, that p is missing. 

### Steps to reproduce

```
import de.uka.ilkd.key.java.Services;
import de.uka.ilkd.key.logic.*;
import de.uka.ilkd.key.logic.op.Function;
import de.uka.ilkd.key.logic.op.Junctor;
import de.uka.ilkd.key.logic.sort.Sort;
import de.uka.ilkd.key.pp.LogicPrinter;
import de.uka.ilkd.key.pp.NotationInfo;
import de.uka.ilkd.key.pp.ProgramPrinter;

import java.io.IOException;
import java.util.HashMap;

public class TestCaseBug1 {
    public static void main(String argv[]) {
        TermFactory termFactory = new TermFactory(new HashMap<>());
        Term p = termFactory.createTerm(new Function(new Name("p"), Sort.FORMULA));
        Term flse = termFactory.createTerm(Junctor.FALSE);

        SequentFormula b = new SequentFormula(p);
        SequentFormula c = new SequentFormula(flse);

        Semisequent a = new Semisequent(b);
        Semisequent d = a.insertFirst(c).semisequent();

        System.out.println(d);
        System.out.println(LogicPrinter.quickPrintSemisequent(d, null));

        Sequent s = new Sequent(d, d);
        System.out.println(LogicPrinter.quickPrintSequent(s, null));


    }
}

```

### Additional Information

```
I tracked down the problem to Layouter.print. 
The term "p" (in the above example) is enqueued into the stream, but never dequeued. 
 
```

## History

* (at)weigl -- (`NEW_BUG`) _2016-04-01_

## Attributes

* Category: CLI
* Status: NEW
* Severity: MINOR
* OS: 
* Target Version: 
* Resolution: OPEN
* Priority: NORMAL
* Reproducibility: ALWAYS
* Platform: Linux
* Commit: 5d09dae7fd1162f3343d1b91bdd5b2033cd1d182
* Build: 
* Tags []
* Labels:  ~Command Line Interface  ~Bug  ~NORMAL  
* Version: _2.6_ 

---

[View in Mantis](http://i12www.ira.uka.de/~klebanov/mantis/view.php?id=1584)
    
----
    
Information: 
* created_at: 2017-05-29T02:32:55.942Z
* updated_at: 2017-05-29T02:32:56.184Z
* milestone: 
* user_notes_count: 0

unp1:
## Related Issue

This pull request resolves #243 and #3570.

### #243 — `LogicPrinter.quickPrintSemisequent` dropped the last formula

The method printed the semisequent without an enclosing layouter block, so the
pretty-printer never flushed its final pending break and the trailing formula was lost
(e.g. `true,` instead of `true, false`). The same formulas printed via a full sequent
rendered correctly because `printSequent` wraps its output in `beginC/end`. Fixed by
wrapping the semisequent in an explicit block, mirroring `quickPrintTerm`.

### #3570 — real literals are now supported

`\real` literals such as `1.0r` used to be rejected by the term parser. This PR
implements them end to end.

* **Exact, unbounded representation.** A real is a mathematical value, so `RealLiteral`
  now stores it as an arbitrary-precision pair `(unscaledValue, scale)` with value
  `unscaledValue * 10^(-scale)` — the `BigDecimal` scheme, but with an unbounded
  `BigInteger` scale rather than an `int`, and kept in canonical form (trailing zeros
  stripped) so equal values have equal fields. `BigDecimal` is used only to parse bounded
  source strings, never to hold the value.
* **Term encoding.** The literal is encoded as `__R(unscaledValue, scale)` over `numbers`
  (`realHeader.key` now declares `real __R(numbers, numbers);` and the `real` sort, moved
  out of `integerHeader.key`). `TermBuilder.rTerm` / `RealLDT.translateLiteral` /
  `translateTerm` build and decode it; because both operands are unbounded `numbers`, the
  value never passes through an `int` (a scale beyond `Integer.MAX_VALUE` round-trips).
* **Parsing.** `ExpressionBuilder.visitRealLiteral` builds the term; a negative literal
  folds the sign into the literal (the real LDT has no `neg`), mirroring the existing
  integer-literal handling and using a bootstrap-safe namespace lookup.
* **Pretty-printing.** A new `Notation.RealLiteral` prints `__R(u, s)` back as its
  canonical decimal with an `r` suffix (`1.25r`, `-0.5r`, `100r`), the inverse of the
  parser; registered next to `Z`/`FP`/`DFP`.
* **Example + regression test.** `standard_key/reals/realLiterals.key` showcases real
  literals, real-sorted quantification and a real-valued assignment (via an update); it is
  wired into `runAllProofs` (it proves *and* reloads, so the printer→parser round-trip is
  checked on every CI run).

#### Decisions worth highlighting

* **Reserved `__` name.** The wrapper is named `__R` rather than a bare identifier so it
  does not steal a name a user might declare in a problem file. This is the first user of a
  reserved-prefix convention; migrating the other built-in literal symbols
  (`Z`/`C`/`FP`/`DFP`) out of the user namespace is left as separate follow-up work.
* **Scope.** The `RealLDT` is intentionally still a stub: there are **no arithmetic axioms**
  yet (no `+`, `*`, `<` on reals), so only equality/structural reasoning is available. The
  `real` type is also not yet usable inside Java program blocks (`\<{ ... }\>`), only in the
  logic and in JML; the example therefore models an assignment with an update. Axiomatising
  real arithmetic is a separate, larger piece of work.

## Type of pull request

## Ensuring quality

- I added new test case(s):
  - `Issue243Test` — red before, green after the semisequent fix.
  - `RealLDTTest` — the `(unscaledValue, scale)` encoding round-trips (incl. negatives,
    leading-zero fractions, trailing-zero integers, and a scale beyond `Integer.MAX_VALUE`).
  - `Issue3570Test` — a real literal parses to sort `real`, and parser and pretty-printer
    are inverse (incl. negatives and trailing-zero canonicalization).
- I have tested the feature as follows: the unit tests above, plus the full regression
  suites `:key.core:testRAP -PrapForks=10` (674 — including the new `reals` proof, which is
  proved and reloaded) and `:key.core:testProveRules` (203) — all green.
- I have checked that runtime performance has not deteriorated (the added parser/printer
  branches only fire for real literals; the extra namespace lookup mirrors the existing
  integer-literal handling on the same code path).

PR prepared with AI tooling support.

### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/nparser/builder/ExpressionBuilder.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/nparser/builder/ExpressionBuilder.java
@@ -13,6 +13,7 @@ import java.util.stream.Collectors;
 
 import de.uka.ilkd.key.java.*;
 import de.uka.ilkd.key.java.ast.abstraction.KeYJavaType;
+import de.uka.ilkd.key.java.ast.expression.literal.RealLiteral;
 import de.uka.ilkd.key.java.ast.expression.literal.StringLiteral;
 import de.uka.ilkd.key.ldt.*;
 import de.uka.ilkd.key.logic.*;
@@ -274,7 +275,8 @@ public class ExpressionBuilder extends DefaultBuilder {
         JTerm result = accept(ctx.sub);
         assert result != null;
         if (ctx.MINUS() != null) {
-            Function Z = functions().lookup("Z");
+            Function Z = functions().lookup(IntegerLDT.NUMBERS_NAME);
+            final Function realNumbers = functions().lookup(RealLDT.REAL_NUMBERS_NAME);
             if (result.op() == Z) {
                 // weigl: rewrite neg(Z(1(#)) to Z(neglit(1(#))
                 // This mimics the old JavaKeYParser behaviour. Unknown if necessary.
@@ -282,6 +284,12 @@ public class ExpressionBuilder extends DefaultBuilder {
                 final JTerm num = result.sub(0);
                 return capsulateTf(ctx,
                     () -> getTermFactory().createTerm(Z, getTermFactory().createTerm(neglit, num)));
+            } else if (realNumbers != null && result.op() == realNumbers) {
+                final Function neglit = functions().lookup("neglit");
+                final JTerm unscaled = result.sub(0);
+                final JTerm scale = result.sub(1);
+                return capsulateTf(ctx, () -> getTermFactory().createTerm(realNumbers,
+                    getTermFactory().createTerm(neglit, unscaled), scale));
             } else if (result.sort() != JavaDLTheory.FORMULA) {
                 Sort sort = result.sort();
                 if (sort == null) {
@@ -1704,13 +1712,16 @@ public class ExpressionBuilder extends DefaultBuilder {
 
     @Override
     public Object visitRealLiteral(RealLiteralContext ctx) {
-        String txt = ctx.getText(); // full text of node incl. unary minus.
-        char lastChar = txt.charAt(txt.length() - 1);
-        if (lastChar == 'R' || lastChar == 'r') {
-            semanticError(ctx,
-                "The given float literal does not have a suffix. This is essential to determine its exact meaning. You probably want to add 'r' as a suffix.");
+        // full text incl. an optional leading '-' and the optional 'r'/'R' suffix
+        String txt = ctx.getText();
+        final char last = txt.charAt(txt.length() - 1);
+        if (last == 'r' || last == 'R') {
+            txt = txt.substring(0, txt.length() - 1);
         }
-        throw new Error("not yet implemented");
+        // RealLiteral parses the decimal exactly into (unscaledValue, scale); rTerm wraps it as
+        // __R(unscaledValue, scale) -- the same encoding RealLDT.translateLiteral produces.
+        final RealLiteral real = new RealLiteral(txt);
+        return getServices().getTermBuilder().rTerm(real.getUnscaledValue(), real.getScale());
     }
 
     @Override
--- a/key.core/src/main/java/de/uka/ilkd/key/java/TypeConverter.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/java/TypeConverter.java
@@ -96,6 +96,10 @@ public final class TypeConverter {
         return (DoubleLDT) getLDT(DoubleLDT.NAME);
     }
 
+    public RealLDT getRealLDT() {
+        return (RealLDT) getLDT(RealLDT.NAME);
+    }
+
     public BooleanLDT getBooleanLDT() {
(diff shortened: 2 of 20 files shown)


key.core/src/main/java/de/uka/ilkd/key/nparser/builder/ExpressionBuilder.java
```diff
@@ -1703,13 +1703,13 @@ public Object visitDoubleLiteral(DoubleLiteralContext ctx) {
...
     @Override
     public Object visitRealLiteral(RealLiteralContext ctx) {
-        String txt = ctx.getText(); // full text of node incl. unary minus.
-        char lastChar = txt.charAt(txt.length() - 1);
-        if (lastChar == 'R' || lastChar == 'r') {
-            semanticError(ctx,
-                "The given float literal does not have a suffix. This is essential to determine its exact meaning. You probably want to add 'r' as a suffix.");
-        }
-        throw new Error("not yet implemented");
+        // Real literals (\real, suffix 'r'/'R') are not yet supported by the term parser: RealLDT
```
wadoon: Then make it parsable. This PR has no progress.

unp1: Will look into that. For the moment I moved it to 3.1.x and moved the PR back to in "Progress" as it might take a bit longer. The work around in reusing only the existing numbers theory for floats/doubles was to use bit encoding. For reals that does not work. I will go with a binary R(integral, decimal) and see how that works out. In any case, if I complete it in time it is fine to go into 3.0.0 but it should not block that version. 

Btw. for the real stuff I use __R (to avoid blocking free names, any suggestions) for the number function. Did not dare to use $ ;-)

unp1: Literals can now be parsed, represented as a term and pretty printed. 
Missing but out of scope of this PR: a theory for reals

wadoon (approved):

PR #3869, opened 2026-06-27
Reject generic sorts at user-input boundaries: \problem, JML, taclet instantiation (#3409)

### LINKED ISSUE #3409: Generic sort in problem definition leads to broken sequent
## Description

If a generic sort is used inside a problem definition (in a .key file), this leads to a broken sequent.

## Reproducible
  always

### Steps to reproduce 

Load a .key file containing the following:
   ```
   \sorts {
       \generic S;
   }

   \functions {
       boolean S::test(int);
   }

   \problem {
       S::test(5) = TRUE
   }
   ``` 
KeY loads the file without error message. However, in the sequent S does not occur anymore. Instead, "null" is printed.
The sequent looks like this:
image

#### Expected behavior
Some context on generic sorts (that was at least not completely clear to me before a discussion with @mattulbrich):
Generic sorts are placeholders that are to be instantiated with existing sorts "at verification time". They are allowed to appear in taclets (to express a family of rules), but not in the problem. The latter would result in a parameterized proof, which could be instantiated for concrete sorts (something that is possible in Isabelle for example, but not in KeY).

To avoid name clashes between the generic sorts and Java classes, the generic sorts are erased from the namespaces before starting the proof. It looks like this erasure happens before starting the proof (the resulting "null" probably stems from a toString() call where the sort object is null ...), but after parsing the problem. Thus, no error is thrown.

### Additional information

The existing issues about generic sorts (#355, #720) seem to be outdated. It may be worth to investigate and close them if possible.

unp1:
## Related Issue

This pull request resolves #3409.

## Intended Change

A generic sort is a schematic placeholder for taclets and must never occur in a concrete sequent.
Previously such input was accepted silently. This PR rejects generic sorts at the **user-input /
parser boundaries**, with a clear, positioned message, e.g.:

> The generic sort 'S' must not occur in a \problem. Generic sorts may only be used in taclets.

### Design — a single place for user-input checks

Rather than inlining the check at each parser boundary, the validations are centralized:

- `GenericSortDetector` — low-level walker for `GenericSort` occurrences (subterm sorts + bound
  variables).
- `UserInputValidator` — a small **registry of checks** applied to terms/sequents that originate
  from user input. New checks are added in one place (`CHECKS`); each boundary just calls
  `validate(...)` and raises its own natural, positioned exception with the returned message.

Because the validator is only invoked at user-input boundaries, **every registered check is
automatically scoped to user input** — automatic term creation never runs them, so there is no
proof-search overhead. This is the intended extension
point for future user-only checks (e.g. a stricter well-typedness check): add one `Check`, touch no
boundary. The check is deliberately **not** wired into term construction or automatic proof search,
where a generic-sort leak would be an implementation error to be found by other means.

Boundaries wired up in this PR:
- **`\problem` sections** — `KeYUserProblemFile.readProblem` (e.g. a constant `S c;` with
  `\generic S` used in the problem term).
- **JML translation** — `JmlIO.interpret` (the single funnel for all JML term translation).
- **Interactive taclet instantiation** — `TacletFindModel.parseTerm` (user-entered instantiation
  text), reported via the usual `SVInstantiationParserException` with position.

Notes:
- The originally reported example (`boolean S::test(int)`) is no longer reproducible as a broken
  sequent — it collapses to a concrete function. The genuine remaining leak is a symbol *of*
  generic sort in the term, which this PR addresses.
- Bound variables of generic sort were already rejected elsewhere ("Cannot have logic variable of
  generic sort!").
- Generic sorts remain legal in taclets; taclet soundness POs are loaded via a different path and
  are unaffected.
- **Not included (deliberately):** a check in the proof replayer (`IntermediateProofReplayer`).
  A replayed instantiation can only reference a generic sort if that sort is in the proof's
  namespace; for ordinary proofs generic sorts are erased (and a tampered file declaring one is
  caught at the `\problem` boundary above), whereas for taclet soundness POs they might be *legitimately* (did not investigate further for this PR) in scope. A replayer check would therefore only fire on the legitimate taclet-PO case, so it was
  dropped.

## Plan

* [x] Detect + reject generic sorts in `\problem` (this PR)
* [x] Apply the same check to JML translation (this PR)
* [x] Apply the same check to interactive taclet instantiation (this PR)

## Type of pull request

## Ensuring quality

- I added new test case(s) for new functionality (`GenericSortInProblemTest`,
  `GenericSortDetectorTest`).
- I have tested the feature as follows: ran the new tests (leak rejected at the problem boundary;
  detector unit-tested; declared-but-unused generic sort still loads), and the regression suites
  `:key.core:testRAP -PrapForks=10` (673 tests, incl. the legitimate generic/polymorphic examples
  `pseq.key`, `dt_polymorphic.proof`) and `:key.core:testProveRules` (203 taclet soundness proofs)
  — all green, after both the problem and the JML hooks.
- I have checked that runtime performance has not deteriorated (no check added to term creation or
  automatic proof search).

PR has been created with AI tooling support

### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/logic/GenericSortDetector.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/logic/GenericSortDetector.java
@@ -3,16 +3,20 @@
  * SPDX-License-Identifier: GPL-2.0-only */
 package de.uka.ilkd.key.logic;
 
-import de.uka.ilkd.key.logic.sort.GenericSort;
+import java.util.ArrayList;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Set;
 
 import org.key_project.logic.op.QuantifiableVariable;
+import org.key_project.logic.sort.Sort;
 import org.key_project.prover.sequent.Sequent;
 import org.key_project.prover.sequent.SequentFormula;
 
 import org.jspecify.annotations.Nullable;
 
 /**
- * Detects occurrences of {@link GenericSort}s within terms and sequents.
+ * Detects occurrences of generic sorts within terms and sequents.
  * <p>
  * Generic sorts are schematic placeholders that may only appear in taclets; they must never occur
  * in a concrete sequent. This detector is meant to be used at <em>user-input/parser boundaries</em>
@@ -20,6 +24,11 @@ import org.jspecify.annotations.Nullable;
  * with a clear message. It is deliberately <em>not</em> wired into term construction or automatic
  * proof search, where a generic-sort leak would be an implementation error to be found by other
  * means.
+ * <p>
+ * A sort is reported when {@link Sort#containsGenericSort()} holds. This catches a generic sort
+ * even
+ * when it is nested inside a parametric sort (e.g. {@code List<[E]>}, or the result sort of
+ * {@code select<[E]>(...)}), not only when a term's own sort is itself a generic sort.
  *
  * @author KeY developers
  */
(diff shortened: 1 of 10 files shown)


key.core/src/main/java/de/uka/ilkd/key/logic/GenericSortDetector.java
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     * @param term a term, may be {@code null}
+     * @return the first {@link GenericSort} occurring in {@code term} (in a subterm's sort or a
+     *         bound variable's sort), or {@code null} if none occurs
+     */
+    public static @Nullable GenericSort findIn(@Nullable JTerm term) {
+        if (term == null) {
+            return null;
+        }
+        if (term.sort() instanceof GenericSort gs) {
```
Drodt: This is not sufficient. It would not find, e.g., `select<[E]>(...)` where `E` is a generic sort.

A `containsGenericSort` method was added to `Sort` some time ago that should be used here.

key.core/src/main/java/de/uka/ilkd/key/logic/UserInputValidator.java
```diff
@@ -0,0 +1,108 @@
...
+
+    /**
+     * Runs all registered checks on {@code term}.
+     *
+     * @param term a term produced from user input, may be {@code null}
+     * @param context a short description of the origin for the error message (e.g. {@code "a
+     *        \\problem"}, {@code "a JML expression"}, {@code "an instantiation"})
+     * @return the first validation error, or empty if {@code term} is acceptable
+     */
+    public static Optional<String> validate(@Nullable JTerm term, String context) {
```
Drodt: One of the error reporting improvements in your recent PRs was to report multiple errors. Should we return a list here rather than just the first one?

Drodt (changes requested):

key.core/src/main/java/de/uka/ilkd/key/logic/UserInputValidator.java
unp1: thanks! Is addressed in the latest commit

key.core/src/main/java/de/uka/ilkd/key/logic/GenericSortDetector.java
unp1: Thanks. containsGenericSort is now used

Drodt (approved):
Thanks!

PR #3870, opened 2026-06-27
Report missing \endmodality at the modality opening; move source-location classes to ncore (#3867)

### LINKED ISSUE #3867: Parser error position is wrong for a missing \endmodality (lexer mode runs on to the next modality)
## Description

When a `\modality{…}{… …}\endmodality` block is missing (or has a mistyped) closing
`\endmodality`, the reported error position is far from the actual mistake. The modality is
lexed by a dedicated lexer mode that runs until it sees the next `\endmodality` token. If the
intended closing token is missing, the lexer keeps consuming — across the rest of the rule and
into following rules — until it reaches the *next* `\endmodality`. Everything in between is
swallowed into the modality, so the parser only reports a problem at a token well after the real
error (typically a `;` belonging to a later rule).

The source/position plumbing itself is fine (this is not a regression of #3553/#3560): the
location correctly points at the offending token ANTLR reports. The token is simply the wrong
one, because the missing close token caused the lexer mode to mis-pair across rules.

This was noticed while verifying the #3553 fix: with `\endmodality` deliberately broken in
`javaRules.key`, the error dialog now appears and loads the source (good), but it marks the
semicolon after a later rule's `\replacewith`, not the broken token in the `\find`.

## Reproducible

always

### Steps to reproduce

Parse the following (e.g. as a `.key` rules snippet); the first rule's `\endmodality` is
replaced by `endm`:

```
\rules {
  brokenRule {
    \find(==>
       \modality{#allmodal}{.. ...}endm post)
    "x":
       \replacewith(==> \modality{#allmodal}{.. ...}\endmodality post);
  };
}
```

Observed parser error:

```
line 6:70 one of '\varcond', '\sameUpdateLevel', '\inSequentState', '\antecedentPolarity',
'\succedentPolarity', '\closegoal', ... expected, but found ';'
```

i.e. the error is reported at the `;` on line 6 (after the *second* rule's `\replacewith`),
although the actual mistake (`endm`) is on line 4.

**Expected:** the error should point at (or near) the offending `endm` / the unterminated
modality on line 4 — e.g. report the modality that was opened but never closed.

**Actual:** the error points at a `;` in a later rule, because the modality lexer mode consumed
everything up to the next `\endmodality`.

### Additional information

- The reported `Location` is correct for the offending token; the root cause is the modality
  island-lexer's error recovery, not `ExceptionTools`/`getLocation`.
- A possible direction: when the modality lexer mode does not find its `\endmodality` before an
  unexpected boundary (or reaches EOF), emit an error anchored at the modality's opening token,
  rather than silently running on to the next `\endmodality`.

---

* Commit: cb5770e49b35948f7942a63519dfdda791a883b2

unp1:
## Related Issue

This pull request resolves #3867.

> **Reviewers:** this PR is **3 commits**:
> 1. *Report missing `\endmodality` at the modality opening* — the bug fix (below).
> 2. *Move `Position`, `Location`, `HasLocation` to `key.ncore`* — addresses the review request
>    that `UnterminatedModalityException` implement `HasLocation` instead of being special-cased in
>    `ExceptionTools`. Since the generated lexer (in `key.ncore`) throws it, and `HasLocation`
>    returns a `key.core` `Location`, the fix was to move the *language-independent* source-location
>    classes down into `key.ncore` (neutral package `org.key_project.util.parsing`). `key.ncore`
>    stays free of any JavaParser/JavaDL dependency: the JavaParser-coupled factories
>    (`Position.fromJPPosition`, `Location.fromNode`) move into the new `key.core` helper
>    `JavaSourceLocations`, and `getURIFromTokenSource` moves into the new `key.ncore` helper
>    `SourceNames`. The special-case branch is then **deleted** from `ExceptionTools`. The rest is
>    mechanical import updates.
> 3. *Move `RecognitionException` to ncore; drop `getURIFromTokenSource` delegate* — small follow-on
>    cleanup enabled by (2): `RecognitionException` becomes language-independent and moves to
>    `key.ncore`; the now-redundant `ExceptionTools.getLocation(RecognitionException)` (covered by
>    the generic `instanceof HasLocation` branch) and its helper are removed.
>
> Each commit was validated with `:key.core:testRAP -PrapForks=10` (673) and
> `:key.core:testProveRules` (203) — all green.

## Intended Change

The KeY lexer scans a modality body greedily until it literally finds the closing keyword
(`\endmodality` / `\>` / `\]`); the (schematic) program text is re-parsed separately. As a
consequence, a missing or mistyped closing keyword made the lexer run on to the *next*, unrelated
closing keyword, and the error surfaced far from the actual mistake — e.g. at a `;` several rules
later (see the issue).

A modality body is a schematic program and can never legitimately contain another
modality-*opening* keyword. This PR detects such a nested opening in the modality lexer modes
(`modGeneric`, `modDiamond`, `modBox`) and fails the token at its start — mirroring the existing
end-of-file behaviour — so the error is reported at the opening of the unterminated modality.

Before:
```
... expected, but found ';'           (reported on the line of the next rule)
```
After:
```
Missing '\endmodality': this modality is not terminated (another modality opening was found
before its closing keyword).        (reported at the unterminated modality's opening)
```

### Comments and string literals are exempt

So that the new check (and the lexer in general) does not misinterpret keyword-looking text inside
a Java comment or literal, the modality lexer modes now sub-mode all four cases:

- **String/char literals** (`"…"`, `'…'`) were already sub-moded.
- **Java line comments** (`// …`) and **block comments** (`/* … */`) are now sub-moded as well.

As a side benefit this also fixes a pre-existing problem: a *closing* keyword inside a comment
(e.g. `/* … \> … */`) used to terminate the modality early; it is now correctly ignored. Ordinary
comments (without modality keywords) tokenise exactly as before.

### Why it is safe

- The opening-keyword detection only matches the **backslash** forms (`\modality`, `\diamond`,
  `\box`, `\diamond_transaction`, `\box_transaction`, `\throughout`, `\throughout_transaction`,
  `\<`, `\[`, `\[[`). Bare `<` / `[` / `/` in a Java program (comparisons, array access/types,
  division) are unaffected.
- The position is carried by a new `UnterminatedModalityException` (placed in `key.ncore` so the
  generated lexer can throw it) and turned into a positioned message by `ExceptionTools`.

## Type of pull request

## Ensuring quality

- I added new test case(s) for new functionality (`Issue3867Test`): the repro is reported at the
  modality opening; modality keywords inside line/block comments are ignored; generic/diamond/box
  modalities still parse.
- I have tested the feature as follows: ran the new test, and the full regression suites
  `:key.core:testRAP -PrapForks=10` (673 tests) and `:key.core:testProveRules` (203 taclet
  soundness proofs) — both green, confirming no false positives across the whole (modality-heavy)
  taclet base and example set.
- I have checked that runtime performance has not deteriorated (the new lexer rules only add
  alternatives to existing modality modes; no work is added on the common path).

The end-of-file case ("no `\endmodality` anywhere") already reported at the modality opening; this
PR brings the "a later `\endmodality` exists" case in line with that behaviour.

### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/util/ExceptionTools.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/util/ExceptionTools.java
@@ -24,6 +24,8 @@ import de.uka.ilkd.key.util.parsing.BuildingIssue;
 import de.uka.ilkd.key.util.parsing.HasLocation;
 import de.uka.ilkd.key.util.parsing.SyntaxErrorReporter;
 
+import org.key_project.util.parsing.UnterminatedModalityException;
+
 import org.antlr.v4.runtime.*;
 import org.antlr.v4.runtime.misc.IntervalSet;
 import org.antlr.v4.runtime.misc.ParseCancellationException;
@@ -302,6 +304,11 @@ public final class ExceptionTools {
         // errors had a message but no location, so the GUI/console could not show the source.
         if (exc instanceof NoViableAltException nvae) {
             return Location.fromToken(nvae.getOffendingToken());
+        }
+
+        if (exc instanceof UnterminatedModalityException ume) {
+            return new Location(MiscTools.getURIFromTokenSource(ume.getSourceName()),
+                Position.fromOneZeroBased(ume.getLine(), ume.getCharPositionInLine()));
         }
 
         if (exc.getCause() != null) {
--- a/key.core/src/main/java/de/uka/ilkd/key/util/ExceptionTools.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/util/ExceptionTools.java
@@ -5,8 +5,6 @@ package de.uka.ilkd.key.util;
 
 import java.io.File;
 import java.net.MalformedURLException;
-import java.net.URI;
-import java.net.URISyntaxException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
@@ -24,6 +22,7 @@ import de.uka.ilkd.key.util.parsing.SyntaxErrorReporter;
 import org.key_project.util.parsing.HasLocation;
 import org.key_project.util.parsing.Location;
 import org.key_project.util.parsing.Position;
+import org.key_project.util.parsing.SourceNames;
 
 import org.antlr.v4.runtime.*;
 import org.antlr.v4.runtime.misc.IntervalSet;
@@ -341,7 +340,7 @@ public final class ExceptionTools {
         if (ip == null) {
             return Location.fromToken(offending);
         }
-        return new Location(MiscTools.getURIFromTokenSource(offending.getTokenSource()), ip);
+        return new Location(SourceNames.getURIFromTokenSource(offending.getTokenSource()), ip);
     }
 
     /**
@@ -412,13 +411,4 @@ public final class ExceptionTools {
         }
         return false;
     }
-
-    private static URI parseFileName(String filename) throws MalformedURLException {
-        try {
-            return filename == null ? null : MiscTools.parseURL(filename).toURI();
-        } catch (URISyntaxException e) {
-            throw new RuntimeException(e);
-        }
-    }
-
 }
--- a/key.core/src/main/java/de/uka/ilkd/key/util/ExceptionTools.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/util/ExceptionTools.java
@@ -12,19 +12,18 @@ import java.util.List;
 import java.util.Map;
 import java.util.regex.Pattern;
 
-import de.uka.ilkd.key.java.Position;
 import de.uka.ilkd.key.java.loader.JavaBuildingExceptions;
 import de.uka.ilkd.key.java.loader.JavaBuildingIssue;
-import de.uka.ilkd.key.parser.Location;
 import de.uka.ilkd.key.proof.io.ProblemLoaderException;
 import de.uka.ilkd.key.speclang.PositionedString;
 import de.uka.ilkd.key.util.parsing.BuildingException;
 import de.uka.ilkd.key.util.parsing.BuildingExceptions;
 import de.uka.ilkd.key.util.parsing.BuildingIssue;
-import de.uka.ilkd.key.util.parsing.HasLocation;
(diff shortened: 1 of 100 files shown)


key.ncore/src/main/java/org/key_project/util/parsing/UnterminatedModalityException.java
```diff
@@ -0,0 +1,49 @@
...
+ * ({@code \endmodality} / {@code \>} / {@code \]}) is missing. It carries the position of the
+ * <em>opening</em> of the unterminated modality so the error can be reported there (rather than at
+ * the next, unrelated closing keyword the lexer would otherwise run on to). See issue #3867.
+ *
+ * <p>
+ * This lives in {@code key.ncore} so the generated lexer can throw it; the position is turned into
+ * a
+ * proper {@code Location} by {@code ExceptionTools} in {@code key.core}.
+ */
+public class UnterminatedModalityException extends RuntimeException {
```
wadoon: This class should implement the interface `HasLocation`.

key.core/src/main/java/de/uka/ilkd/key/util/ExceptionTools.java
```diff
@@ -311,6 +313,11 @@ private static String friendlyTokenName(Vocabulary vocabulary, int tokenType) {
             return Location.fromToken(nvae.getOffendingToken());
         }
 
+        if (exc instanceof UnterminatedModalityException ume) {
```
wadoon: This special case can be avoided. Let `UMexception` implement `HasLocation`.

wadoon (changes requested):

key.core/src/main/java/de/uka/ilkd/key/util/ExceptionTools.java
unp1: done

key.ncore/src/main/java/org/key_project/util/parsing/UnterminatedModalityException.java
unp1: The problem is that HasLocation has implicit dependencies to JavaParser and other core classes and cannot move to ncore where the lexer logic lives. That is why the submittedt changes are now a bit larger as they need to refactor the module structure.

wadoon (approved):

PR #3874, opened 2026-06-30
Performance Series 2 (1/5): Faster cost computation by skipping an unnecessary proof-history walk

unp1:
## Intended Change

The strategy's cost computation includes an `IntroductionTime` feature that walks the proof history to find when an operator was first introduced. For operators that are not Skolem symbols this walk can never produce a useful result, so it is now skipped. This affects only the cost heuristic that guides proof search — it never changes soundness or which proofs can be closed.

## Benchmark

Automode time, median of 3 runs, serial forks (`-PrapForks=1`), isolated `user.home` per run; baseline = current main (8 representative proofs: 4 quantifier-heavy + 4 mixed). _ArrayList.remove.1 is high-variance run-to-run — treat as noise._

|---|--:|--:|--:|
| PUZ031p1 | 13712 | 12557 | -8% |
| SimplifiedLinkedList.remove | 17469 | 16492 | -6% |
| Saddleback | 13026 | 11964 | -8% |
| SYN550p1 | 821 | 762 | -7% |
| symmArray | 12737 | 11276 | -11% |
| gemplusDecimal | 8378 | 7154 | -15% |
| coincidence | 2181 | 2032 | -7% |
| ArrayList.remove.1 | 3339 | 2497 | -25% |
| **TOTAL** | **71663** | **64734** | **-10%** |

## Plan

* [x] Add a focused per-PR benchmark to this description

## Type of pull request

## Ensuring quality

### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/strategy/feature/AbstractMonomialSmallerThanFeature.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/strategy/feature/AbstractMonomialSmallerThanFeature.java
@@ -34,6 +34,18 @@ public abstract class AbstractMonomialSmallerThanFeature extends SmallerThanFeat
             return -1;
         }
 
+        // A newSmallSym taclet introduces its symbol as a SkolemTermSV instantiation, which is
+        // always a skolem-constant function (TacletApp.createSkolemConstant). So an op that is not
+        // a skolem-constant function can never have been introduced by one: its time is -1, with no
+        // need to scan the applied-rule history. This is what made the scan a hotspot -- the common
+        // monomial atoms (program variables, ordinary functions) are not skolem constants, yet
+        // walked the full O(history) on every compare and, never being "introduced", were never
+        // cached. (A skolem constant from some OTHER rule still walks and returns -1; only the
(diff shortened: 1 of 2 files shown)


key.core/src/main/java/de/uka/ilkd/key/strategy/feature/AbstractMonomialSmallerThanFeature.java
```diff
@@ -34,6 +34,18 @@ protected int introductionTime(Operator op, Goal goal) {
             return -1;
         }
 
+        // A newSmallSym taclet introduces its symbol as a SkolemTermSV instantiation, which is
```
Drodt: Maybe mention that this is about the `polySimp_newSmallSym` rule set. We should also document this requirement (that the introduction time is only defined for skolem terms) somewhere.

unp1: Thanks. I added some additional comments and mention the rule set  explicitly

Drodt (approved):
Thanks!

PR #3875, opened 2026-06-30
Performance Series 2 (2/5): Faster term equality modulo renaming via a hash pre-filter

unp1:
## Intended Change

`RenamingTermProperty.equalsModThisProperty` decides whether two terms are equal up to renaming of bound variables (alpha-equivalence), and is evaluated very frequently during proof search. This adds a renaming-invariant `hashCode` pre-filter: two terms with different renaming-invariant hash codes cannot be alpha-equivalent, so they are rejected in constant time before the more expensive structural comparison. Behaviour-preserving — the hash is renaming-invariant, so the pre-filter only ever rejects pairs that are genuinely unequal.

## Benchmark

Automode time, median of 3 runs, serial forks (`-PrapForks=1`), isolated `user.home` per run; baseline = current main (8 representative proofs: 4 quantifier-heavy + 4 mixed). _ArrayList.remove.1 is high-variance run-to-run — treat as noise._

|---|--:|--:|--:|
| PUZ031p1 | 13712 | 12644 | -8% |
| SimplifiedLinkedList.remove | 17469 | 15789 | -10% |
| Saddleback | 13026 | 12241 | -6% |
| SYN550p1 | 821 | 754 | -8% |
| symmArray | 12737 | 10768 | -15% |
| gemplusDecimal | 8378 | 8536 | +2% |
| coincidence | 2181 | 2171 | -0% |
| ArrayList.remove.1 | 3339 | 2689 | -19% |
| **TOTAL** | **71663** | **65592** | **-8%** |

## Plan

* [x] Add a focused per-PR benchmark to this description

## Type of pull request

## Ensuring quality

### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/logic/JTerm.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/logic/JTerm.java
@@ -119,6 +119,13 @@ public interface JTerm
     boolean containsJavaBlockRecursive();
 
     /**
+     * Hash code modulo bound renaming, computed lazily and cached. Two terms that are equal modulo
+     * renaming share this value, so a mismatch is a cheap proof of inequality (used by
+     * {@code RenamingTermProperty}).
+     */
+    int hashCodeModRenaming();
+
+    /**
      * Checks if this {@link JTerm} or one of its direct or indirect children has a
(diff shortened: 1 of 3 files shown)


Drodt (approved):
Thanks!

PR #3877, opened 2026-06-30
Performance Series 2 (4/5): Lower memory for the per-goal rule-application queue

unp1:
## Intended Change

A set of memory reductions for the per-goal data structures that dominate heap usage on large proofs (identified with the Eclipse Memory Analyzer): the per-goal rule-application queue's parking buckets are slimmed and large buckets promoted to a set (keeping a key operation O(1)); the taclet index shares its maps copy-on-write across goals; and the `LexPathOrdering` comparison cache is bounded with an LRU. Behaviour-preserving. The goal is to reduce the peak heap held *during* a proof.

## Memory benchmark (the headline metric)

Peak heap during automode, via `-Xlog:gc`; baseline = main. The per-goal rule-application queue savings scale with the **number of goals**, so the effect appears on goal-heavy (wide-branching) proofs:

| proof | peak heap (working set) | live set (after GC) |
|---|--:|--:|
| bike — 150k-node wide-branching stress case | 2268M → **2051M (−10%)** | 1921M → **1718M (−11%)** |
| SimplifiedLinkedList.remove — regular proof (median of 3) | 886M → 912M (+3%) | 470M → 464M (-1 %) |

On the goal-heavy case the slimmer per-goal queues cut peak heap ~10%. On a regular proof (few goals) the effect is within run-to-run noise — the change is **targeted, not a universal cost**.

## Benchmark (time — no regression)

This is a memory optimization; the table below confirms it does not regress runtime (total within run-to-run noise). _Automode time, median of 3 runs, serial forks, isolated home; baseline = current main. ArrayList.remove.1 is high-variance — noise._

|---|--:|--:|--:|
| PUZ031p1 | 13712 | 12877 | -6% |
| SimplifiedLinkedList.remove | 17469 | 17687 | +1% |
| Saddleback | 13026 | 13203 | +1% |
| SYN550p1 | 821 | 798 | -3% |
| symmArray | 12737 | 12405 | -3% |
| gemplusDecimal | 8378 | 8511 | +2% |
| coincidence | 2181 | 2247 | +3% |
| ArrayList.remove.1 | 3339 | 2596 | -22% |
| **TOTAL** | **71663** | **70324** | **-2%** |

## Plan

* [x] Confirm no runtime regression
* [x] Add the peak in-flight heap measurement (headline metric)

## Type of pull request

## Ensuring quality

### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/rule/SyntacticalReplaceVisitor.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/rule/SyntacticalReplaceVisitor.java
@@ -5,7 +5,6 @@ package de.uka.ilkd.key.rule;
 
 import java.util.ArrayDeque;
 import java.util.Deque;
-import java.util.Stack;
 
 import de.uka.ilkd.key.java.Services;
 import de.uka.ilkd.key.java.ast.*;
@@ -60,7 +59,7 @@ public class SyntacticalReplaceVisitor implements DefaultVisitor {
      * indicate that a term using these subterms should build a new term instead of using the old
      * one, because one of its subterms has been built, too.
      */
-    private final Stack<Object> subStack; // of Term (and Boolean)
+    private final Deque<Object> subStack; // of Term (and Boolean)
     private final Boolean newMarker = Boolean.TRUE;
     private final Deque<JTerm> tacletTermStack = new ArrayDeque<>();
 
@@ -91,7 +90,7 @@ public class SyntacticalReplaceVisitor implements DefaultVisitor {
         this.ruleApp = ruleApp;
         this.labelHint = labelHint;
         this.goal = goal;
-        subStack = new Stack<>(); // of Term
+        subStack = new ArrayDeque<>(); // of Term
         if (labelHint != null) {
             labelHint.setTacletTermStack(tacletTermStack);
         }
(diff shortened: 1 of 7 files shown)


key.core/src/main/java/de/uka/ilkd/key/rule/SyntacticalReplaceVisitor.java
```diff
@@ -60,7 +59,7 @@ public class SyntacticalReplaceVisitor implements DefaultVisitor {
      * indicate that a term using these subterms should build a new term instead of using the old
      * one, because one of its subterms has been built, too.
      */
-    private final Stack<Object> subStack; // of Term (and Boolean)
+    private final Deque<Object> subStack; // of Term (and Boolean)
```
Drodt, line 62: Why the change to `Deque`? I see no discussion of this.

unp1: That is just a minor add on. The SyntacticalReplaceVisitor is used only locally not shared among threads so the synchronization Vector uses is not needed overhead. And the cass uses a Deque already for 'tacletTermStack' so no new dependencies.

Drodt (approved):
Thanks!

PR #3884, opened 2026-07-03
Term label cleanup: simpler framework, label-agnostic term equality

unp1:
**In one paragraph:** the term-label subsystem carried a lot of generality that was designed up-front but never actually used, and its central design choice, term `equals`/`hashCode` being label-*sensitive*, forced a comparison-modulo-labels idiom to spread across the code base. This PR removes the unused machinery, flips term equality to ignore labels by default (the common and now-uniform case), restores the few places that genuinely need label-sensitive comparison via a small explicit apparatus, and simplifies `TermLabelManager` and the hook API. Proofs are unaffected: replaying a 17-proof corpus yields byte-identical sequents *including every term label*, and the information-flow and well-definedness proof suites match stock `master`.

## Main changes and intentions

1. **Remove speculative-but-unused features.** Several extension points were anticipated but never materialized in any profile:
   - the `ChildTermLabelPolicy` hook family (direct + child/grandchild) — registered by nobody;
   - the `APPLICATION_DIRECT_CHILDREN` refactoring scope — returned by no refactoring;
   - the `LabeledTermImpl` subclass — folded into `TermImpl`.

2. **Change the semantics of term equality: `equals`/`hashCode` now ignore all term labels.** Term labels are, by contract, not soundness relevant, and sequent redundancy elimination already ignored them — so the *default* now agrees with the calculus' own notion of "same formula". The two label-sensitive needs that remain are served explicitly: `IRRELEVANT_TERM_LABELS_PROPERTY` (ignore only cosmetic labels, keep proof-relevant ones) and `equalsIncludingLabels`/`labeledHashCode` (respect all labels, used only by term interning / index caches). Rationale and the full pro/contra discussion are in the table below.

3. **Simplify `TermLabelManager` and the hook API.** `TermLabelManager` shrinks from 2211 to 1629 lines. The hooks (`TermLabelPolicy`/`Update`/`Refactoring`) no longer thread an 8-parameter rule-application tuple through every signature and every internal delegation; a single immutable `TermLabelContext` record carries it.

Net: **40 files, +1015 / −1691** (−676 lines), Java only, no taclet changes.

## Related Issue

Addresses the two long-standing complaints about the term-label implementation: the
framework being too general/partly inefficient, and the proliferation of
`equalsModIrrelevantTermLabels` comparisons.

### (A) Prune unused generality

- **`ChildTermLabelPolicy`** (and its direct / child-and-grandchild variants) is deleted together with its dispatch in `TermLabelManager`. No `TermLabelConfiguration` in any profile ever registered one.
- **`RefactoringScope.APPLICATION_DIRECT_CHILDREN`** and its `RefactoringsContainer` bucket / dispatch are deleted; no live refactoring returns that scope.
- **`LabeledTermImpl` is merged into `TermImpl`.** With labels no longer part of term identity, the subclass existed only to keep a `labels` field off unlabeled terms; a single nullable field (empty normalized to a shared array) replaces it, removing the two-class hierarchy and the "same value, different class, still equal" hazard.

### (B) Term equality ignores labels

`TermImpl.equals`/`hashCode` now ignore all term labels. Three comparison modes now coexist, cleanly separated by *who owns the knowledge*:

| Mode | API | Knows about label kinds? | Used for |
|---|---|---|---|
| ignore **all** labels | `t1.equals(t2)` / `hashCode` | no — labels are opaque | default: logical identity ("same formula") |
| ignore only **cosmetic** labels | `equalsModProperty(t, IRRELEVANT_TERM_LABELS_PROPERTY)` | yes — consults `isProofRelevant()` | the ~51 sites that must keep, e.g., an `anonHeapFunction`-labeled term distinct while ignoring origin labels |
| respect **all** labels | `equalsIncludingLabels` + `labeledHashCode` | no — labels are opaque | term-factory interning and taclet-index caches (`StrictTermKey`) |

Supporting apparatus (all label-opaque, hence on `TermImpl`, not leaking label kinds): - `equalsIncludingLabels(Object)` and cached `labeledHashCode()` — the label-sensitive counterparts of `equals`/`hashCode`. `labeledHashCode` is cached on the term because `hashCodeModProperty` is *not* and the term factory cache queries it per construction. - `StrictTermKey` — a label-sensitive map key. The `TermFactory` intern cache and the taclet-app index `CacheKey` use it, so labeled terms are still interned exactly as before (memory sharing and `==` fast paths preserved), even though `equals` ignores labels. Taclet matching can depend on labels, so the index must stay label-sensitive. - A cached `containsLabelsRecursive()` guard so label-free subtrees are skipped in O(1).

The four `TERM_LABELS_PROPERTY` comparison sites (which meant "ignore all labels") are now plain `equals`; that property is no longer used in application code.

### (C) API simplification

- **`TermLabelContext`**: an immutable record bundling the rule application (`state, services, applicationPosInOccurrence, applicationTerm, modalityTerm, rule, ruleApp, goal, hint, tacletTerm`), built once per `TermLabelManager` entry point. Hooks now take `(context, payload…)`: `keepLabel` went 10→3 params, `updateLabels` 11→3, `defineRefactoringScope` 8→1, `refactorLabels` 10→3. 26 manager methods that re-threaded the tuple collapse to one-argument delegation. `TermLabelMerger` is unchanged (its parameters are genuine payload). The static entry points keep their signatures, so rule implementations are untouched.

## Why "ignore all" and not "keep the old label-sensitive equals": pro / contra

The decision (which default `equals` should have) was deliberate. Both candidates are label-*opaque* (neither leaks specific label knowledge into `Term`).

| | `equals` **includes** all labels (old) | `equals` **ignores** all labels (this PR) |
|---|---|---|
| conceptual contract | implementation identity ("same content incl. annotations") | logical identity ("same formula") — consistent with *labels must not be soundness relevant* |
| consistency with sequents | inconsistent: sequent membership ignores labels but `equals` does not (the "rejected as duplicate yet not equal to any member" trap) | aligned — sequent redundancy already ignores labels |
| opt-out burden | ~55 sites must *weaken* to ignore labels; proof-relevance leaks widely | 4 infrastructural sites must *strengthen* (the interning caches), all in label-maintenance code |
| new-code failure mode | a naive `equals`/map **breaks once origin labels are enabled** (the historical bug pattern) | a naive term-keyed map **silently merges label variants** → labels dropped; bounded — never unsound, only lost annotations |
| extra machinery | none | `equalsIncludingLabels` + cached `labeledHashCode` + `StrictTermKey` (small, and interning is preserved) |
| perf | baseline | parity on the cached-construction hot path (measured); `equals` marginally faster on labeled terms |

The deciding factor is that the proliferation of comparison-modulo-labels only has an exit under "ignore all": under the old default, plain equality is unusable the moment origin labels are on, so the weakening idiom is *structurally* forced everywhere. Under this PR the weakening residue is the ~51 `IRRELEVANT_TERM_LABELS_PROPERTY` sites that genuinely need the middle mode (a per-site review confirmed each: emptiness/constant classification in the loop/contract rules and WD, instantiation identity in `SVInstantiations`/`RuleCommand`, and a few non-soundness strategy heuristics).

## Plan

Finished at submission time.

* [x] Remove unused `ChildTermLabelPolicy` hook family
* [x] Make term equality label-agnostic; add `equalsIncludingLabels`
* [x] Add term-label equivalence regression harness
* [x] Merge `LabeledTermImpl` into `TermImpl`
* [x] Intern labeled terms via a label-sensitive cache key (`StrictTermKey`)
* [x] Update `TestLinkedHashMapWrapper` for label-agnostic equality
* [x] Short-circuit `labelsEqualRecursive` on reference-identical subterms (perf)
* [x] Pack `TermImpl` lazy tri-state caches into a single int (−12 bytes/term)
* [x] Collapse `TERM_LABELS_PROPERTY` comparisons to plain `equals`
* [x] Remove unused `APPLICATION_DIRECT_CHILDREN` refactoring scope
* [x] Rewrite term-label API documentation
* [x] Introduce `TermLabelContext` parameter object

## Type of pull request

- Breaking change (API: `equals`/`hashCode` semantics of terms, and the `TermLabelPolicy`/`Update`/`Refactoring` hook signatures — relevant only to code implementing those hooks or relying on label-sensitive term equality)

## Ensuring quality

- I made sure that introduced/changed code is well documented (javadoc and inline comments); the equality contract and the fixed hook execution order are documented on `TermLabel`, the hook interfaces, and `TermLabelManager`.
- I made sure that end-user/developer docs are updated: companion PR in https://github.com/KeYProject/key-docs adds a "Term Labels" internals page (concept → design → hook taxonomy) and an "Add a Term Label" how-to.
- I added new test case(s): a term-label equivalence regression harness (`TermLabelEquivalenceDumpTest`) that replays console proofs and dumps every node's sequent including all labels; plus label-interning and equality-contract unit tests.
- I have tested the feature as follows: the equivalence harness (17 proofs, 23 373 nodes) is **byte-identical** including labels across every commit; `:key.core:test`, `:key.core.infflow:testRunAllInfProofs`, and `:key.core.wd:testRunAllWdProofs` are green and match stock `master`.
- I have checked that runtime performance has not deteriorated: a construction/equals/hashCode micro-benchmark against stock `master` shows parity on the cached-construction hot path (the production path); labeled-term `equals` is marginally faster. The flag packing saves ~12 bytes per term object.

## Additional information

@flo2702 thanks a lot for offering to look into this PR

- The equality change is behavior-preserving for proofs but is a semantic change for any code that keyed a `Map`/`Set` on `JTerm` expecting label sensitivity; such code should wrap keys in `StrictTermKey`. The (deprecated) `key.core.symbolic_execution` and `key.core.proof_references` modules were left untouched and pass their proof suites.
- `TermLabelsProperty` (mode "ignore all") is now equivalent to plain `equals` and unused in application code; it is retained only for tests and can be removed in a follow-up.


key.core/src/main/java/de/uka/ilkd/key/logic/equality/IrrelevantTermLabelsProperty.java
```diff
@@ -13,7 +13,17 @@
...
  * A property that can be used in
  * {@link EqualsModProperty#equalsModProperty(Object, Property, Object[])} for terms.
- * All irrelevant term labels are ignored in this equality check.
+ * All irrelevant term labels are ignored in this equality check, while labels that are
+ * {@link TermLabel#isProofRelevant() proof relevant} still count.
+ * <p>
+ * This is the middle one of the three label comparison modes: plain
+ * {@link de.uka.ilkd.key.logic.JTerm#equals(Object)} ignores <em>all</em> term labels, this
+ * property ignores only the non-proof-relevant (cosmetic) ones, and
+ * {@link de.uka.ilkd.key.logic.JTerm#equalsIncludingLabels(Object)} respects all labels. Use this
```
flo2702, line 22: There's also a `TermLabelsProperty` which is only used in test code and seems to be redundant with `JTerm.equalsIncludingLabels`. Should that be removed?

key.core/src/main/java/de/uka/ilkd/key/logic/JTerm.java
```diff
@@ -118,6 +118,36 @@ public interface JTerm
...
+    /**
+     * Checks term equality <em>including</em> {@link TermLabel}s on this term and all subterms
+     * (labels compared as sets). This is the label-sensitive counterpart of
+     * {@link #equals(Object)}.
+     *
+     * @param o the object to compare with
+     * @return {@code true} iff {@code o} is a term syntactically equal to this one with equal
+     *         label sets on all subterms
+     */
+    boolean equalsIncludingLabels(Object o);
```
flo2702, line 139: Probably, the documentation here should reference `IrrelevantTermLabelProperty`, so people don't get confused and use one when they really want the other.

PR #3887, opened 2026-07-03
Soundiness dialog for making assumptions explicit

mattulbrich:
This pull request is a new feature: A dialog to display the sound(i)ness assumptions of KeY explicitly.

## Intended Change

Implements a soundiness info window (reachable via context menu on the proof list or a button in the proof statistics dialog) that analyzes KeY proofs and displays potential sources of unsoundness in a structured HTML report. The feature helps users identify taclet options, and proof characteristics, and dependencies that may affect verification soundness.

Key features:
- [x] General KeY Settings analysis (source consistency, JML activation)
- [x] Taclet Options analysis with plain English impact descriptions
- [x] **section removed from dialog, not yet implemented:** Source Analysis (skeleton for future expansion)
- [x] Proof Tree Analysis (goals, lemmas, assumptions, rule applications, dependency analysis)
- [x] Disclaimer about KeY limitations (reflection, native methods, sequentiality, floating-point)

Future tasks (move to future feature request for next KeY version):
- [ ] Refactor to StringTemplate or similar
- [ ] Implement source analysis (What exactly should this contain?)
- [ ] Address some of the comments of https://github.com/KeYProject/key/pull/3887#issuecomment-4926240674, e.g. refer to supported Java/JML features, explain proof caching

Screenshot:
image

## Type of pull request

## Ensuring quality
    
- [x] I made sure that introduced/changed code is well documented (javadoc and inline comments): More or less, and easy to understand
- [x] I have tested the feature as follows: Opened proof statistics dialog, clicked "Show Soundiness Report" button, verified HTML report displays correctly with all four compartments (General KeY Settings, Taclet Options, Source Analysis, Proof Tree Analysis).
- [x] I have checked that runtime performance has not deteriorated: The information is only collected when opening the dialog, which seems instant on the proofs I tested (a few thousand nodes).

Files added/modified:
- `key.ui/src/main/java/de/uka/ilkd/key/gui/soundiness/SoundinessAnalyzer.java` - Static utility class generating HTML reports
- `key.ui/src/main/java/de/uka/ilkd/key/gui/soundiness/SoundinessDialog.java` - Simple modal dialog displaying HTML
- `key.ui/src/main/java/de/uka/ilkd/key/gui/soundiness/SoundinessExtension.java` - Extension point registration
- `key.ui/src/main/java/de/uka/ilkd/key/gui/soundiness/ShowSoundinessAction.java` - Context menu action
- `key.ui/src/main/java/de/uka/ilkd/key/gui/windows/ShowProofStatistics.java` - Added soundiness button

Documentation updated:
- [x] to delete: `AGENT.soundiness.md` - Design document
- [x] to delete: `IMPLEMENTATION_SUMMARY.md` - Implementation summary

by Qwen3.5 397B A17B, authored and extended by @WolframPfeifer


unp1:
Thanks! Like it :-) Only bikesheding: For stricter then necessary, orange seems more like a warning. Dark green or blue?

key.ui/src/main/java/de/uka/ilkd/key/gui/actions/ShowSoundinessAction.java
```diff
@@ -0,0 +1,39 @@
...
+import de.uka.ilkd.key.gui.MainWindow;
+import de.uka.ilkd.key.gui.fonticons.IconFactory;
+import de.uka.ilkd.key.gui.soundiness.SoundinessDialog;
+import de.uka.ilkd.key.proof.Proof;
+
+/**
+ * Action to show the soundiness report for the currently selected proof.
+ * 
+ * @author opencode
+ * @since 2.13
```
wadoon: Should it not be 3.0?

key.ui/src/main/java/de/uka/ilkd/key/gui/actions/ShowProofStatistics.java
```diff
@@ -317,6 +317,15 @@ private void init(MainWindow mainWindow, String stats) {
             buttonPane2.add(saveBundleButton);
 
             // spotless:off
+            JButton soundinessButton = new JButton("Show Soundiness Report");
+            soundinessButton.setIcon(IconFactory.WARNING_UNSOUND.get());
+            soundinessButton.addActionListener(e -> {
+                de.uka.ilkd.key.gui.soundiness.SoundinessDialog dialog =
```
wadoon: Strange to use fqn here.

key.ui/src/main/java/de/uka/ilkd/key/gui/extension/impl/SoundinessExtension.java
```diff
@@ -0,0 +1,46 @@
...
+    
+    @Override
+    public <T> List<javax.swing.Action> getContextActions(KeYMediator mediator, ContextMenuKind<T> kind, T underlyingObject) {
+        List<javax.swing.Action> actions = new ArrayList<>();
+        
+        // Add to proof list context menu
+        if (kind == ContextMenuKind.PROOF_LIST && underlyingObject instanceof Proof) {
+            MainWindow mainWindow = MainWindow.getInstance();
+            if (mainWindow != null) {
+                actions.add(new ShowSoundinessAction(mainWindow));
```
wadoon, line 40: Would it not be wise to give the proof object to the action?

key.ui/src/main/java/de/uka/ilkd/key/gui/actions/ShowSoundinessAction.java
```diff
@@ -0,0 +1,39 @@
...
+        super(mainWindow);
+        setName("Show Soundiness Report");
+        // TODO Can we have an extension w/o extra menu?
+        putValue(KeyAction.PATH, "Show Soundiness Report");
+        getMediator().enableWhenProofLoaded(this);
+    }
+    
+    @Override
+    public void actionPerformed(ActionEvent e) {
+        Proof proof = getMediator().getSelectedProof();
```
wadoon, line 32: Is the selected proof, also the proof for which the ContextMenu was requested?

AGENT.soundiness.md
```diff
@@ -0,0 +1,321 @@
+# Soundiness Info Window - Design Document
```
wadoon: Will this file be removed?

wadoon (changes requested):

WolframPfeifer:
List of things that came to my mind while playing around a bit with the prototype:
* The taclet options list should also show the key, not only the value (i.e. "runtimeExceptions:ban" instead of "ban" at the moment).
* I think the description "arithmetic wraps around like in Java" for `arithmeticSemanticsIgnoringOF` is wrong
* Should the dialog also explain about proof caching? Due to the pragmatics of the implementation, we are only absolutely sure that a cache hit is applicable (i.e., replayable at the current proof) after doing the replay. There might be "spurious" cache hits, even if I am not sure where they come from.
* The proof statistics dialog looks weirdly wide now due to the new button.
* I would prefer a bullet point list also for "1. General KeY Soundness", to make it easier to read.
* http://soundiness.org/ lists the following things (among what is already shown in the dialog). Not sure what we should mention explicitly:
  * implicit method invocations (finalizers, class initializers, Thread.&lt;init&gt;, etc.)
  * exceptions and flow related to that
  * invokedynamic
  * code generation at runtime, dynamic loading
  * different class loaders
  * key native methods (Object.run, Object.doPrivileged)
* The dialog might also be a good place to list some information from https://www.key-project.org/applications/program-verification/ (generics, library methods, ...) or even https://www.key-project.org/jml-support-in-key/ (maybe semantics of assignable/accessible, invariants?), or link to these pages.
* What about dependencies to other contracts? The information is present (I think in class `ProofCorrectnessMgt`) and shown in the proof list, but maybe it should also be listed here.

key.ui/src/main/java/de/uka/ilkd/key/gui/actions/ShowSoundinessAction.java
WolframPfeifer: Yes, that should be the case. How would the context menu be requested otherwise?

key.ui/src/main/java/de/uka/ilkd/key/gui/extension/impl/SoundinessExtension.java
WolframPfeifer: At the moment we don't need it, so I would keep it as is.

AGENT.soundiness.md
WolframPfeifer: I just removed them

key.ui/src/main/java/de/uka/ilkd/key/gui/actions/ShowProofStatistics.java
WolframPfeifer: I agree, removed it

WolframPfeifer:
> Thanks! Like it :-) Only bikesheding: For stricter then necessary, orange seems more like a warning. Dark green or blue?

I changed the color to blue now.

wadoon (approved):

WolframPfeifer:
Just fixed one more thing: The calculation of goals/steps was wrong (showed steps but labelled it as goals, and did not correctly calculate the automatic/interactive steps). Now, the implementation is fixed by using the existing proof statistics.

PR #3888, opened 2026-07-03
Feature: Version-based Configuration Paths

wadoon:
## Summary
This PR introduces version-based configuration paths for KeY. Starting from KeY 3.0, configuration directories now include the major-minor version as a subdirectory (e.g., `~/.key/v3.0/`).

If the current folder `vX.Z` is not found, the SemVer-newest `vX.Y` folder is taken as the *previous* configuration and `vX.Z` is created. 

Configuration usages are now aware of this circumstance and are able to read from one source and write to a different sink. 

@mattulbrich: Is it on purpose that SMT settings are not respecting the `DISREGARD_SETTINGS_PROPERTY` key for omitting to load settings during test cases?

### PathConfig.java
- Restructured configuration to use versioned subdirectories under `key.home`
- Added `VERSION_COMPARATOR` for comparing version paths (format: `vX.Y`)
- Introduced `KeyPaths` class to encapsulate read/write path sets
- Added `currentPaths` and `previousPaths` for separate read/write configuration handling
- Implemented fallback logic: if the current version config does not exist, finds the most recent previous version to be read from. 

### New System Properties
The defaults are overridable by the user using: 

| Property | Description | Default |
|----------|-------------|---------|
| `key.home` | Base path for configurations | `user.home/.key` |
| `key.config` | Direct path to configuration | `key.home/v<major>.<minor>` |

### logback.xml

- Rollback logback.xml to a previous version because of regression
- logback.xml respect the new versioned configuration folder

### More resilience in configuration reading 
- Every configuration usage class is responsible on their own to deal with `currentPath` and `previousPath`. 

## Backward Compatibility Notes
- Existing configurations in `~/.key/` will be discovered as a fallback
- Empty configuration directories are treated as non-existent for fallback purposes

## Testing
- Unit tests for `VERSION_COMPARATOR` in `PathConfigTest.java`


key.core/src/main/java/de/uka/ilkd/key/settings/PathConfig.java
```diff
@@ -62,53 +122,98 @@ private PathConfig() {
...
-        recentFileStorage = getKeyConfigDir().resolve("recentFiles_v2.json");
-        proofIndependentSettings = getKeyConfigDir().resolve("proofIndependentSettings.props");
-        logDirectory = getKeyConfigDir().resolve("logs");
+    // Resolves the path `<key.home>/v<major>.<minor>` and creates it if necessary.
+    private static KeyPaths getWritingPaths(Path base) throws IOException {
+        var currentVersion = KeYResourceManager.getManager().getVersion().split("\\.");
+        var currentVersionCandidate = "v%s.%s".formatted(currentVersion[0], currentVersion[1]);
+        var currentVersionPath = base.resolve(currentVersionCandidate);
+        if (Files.exists(currentVersionPath)) {
+            Files.createDirectories(currentVersionPath);
```
unp1, line 187: !Files.exists(currentVersionPath)

key.core/src/main/java/de/uka/ilkd/key/settings/PathConfig.java
```diff
@@ -62,53 +122,98 @@ private PathConfig() {
...
 
-    /**
-     * Returns the path to the file that is used to store proof independent settings.
-     *
-     * @return The path to the file.
-     */
-    public static Path getProofIndependentSettings() {
-        return proofIndependentSettings;
+    public static boolean isDifferentReadWriteDirectories() {
+        return Objects.equals(currentPaths.keyConfigDir, previousPaths.keyConfigDir);
```
unp1: !Object.equals

key.core/src/main/java/de/uka/ilkd/key/settings/PathConfig.java
```diff
@@ -62,53 +122,98 @@ private PathConfig() {
...
+                    .filter(Files::isDirectory)
+                    .filter(it -> !Objects.equals(it, currentPaths.keyConfigDir))
+                    .filter(it -> {
+                        try {
+                            return Files.list(it).findAny().isEmpty();
+                        } catch (IOException e) {
+                            return false;
+                        }
+                    })
+                    .max(VERSION_COMPARATOR)
```
unp1, line 207: .filter(it -> {
    try (Stream<Path> entries = Files.list(it)) {
        return entries.findAny().isPresent();   // keep only non-empty config dirs
    } catch (IOException e) {
        return false;
    }
})

key.core/src/main/java/de/uka/ilkd/key/settings/ProofSettings.java
```diff
@@ -206,18 +214,18 @@ public void loadSettingsFromPropertyStream(Reader in) {
...
+                LOGGER.info("Load proof dependent settings from file {}", readPath);
+                loadDefaultJSONSettings();
                 if (isOldFormat) {
-                    loadDefaultJSONSettings();
                     loadSettingsFromPropertyStream(in);
                 } else {
-                    loadDefaultJSONSettings();
                     loadSettingsFromJSONStream(in);
                 }
             } catch (IOException e) {
```
unp1: this can also lead to a runtimeexception e.g. via 
loadSettingsFromJSONStream → Configuration.load → ParsingFacade.parseConfigurationFile → p.getErrorReporter().throwException() → BuildingException

such an occurrence would lead to an ExceptionInInitlizer error as loadSettings is called from a static field and prevent KeY from starting

key.core/src/main/java/de/uka/ilkd/key/settings/ProofIndependentSettings.java
unp1: can also be reached by a RuntimeException, causing KeY not to start up due to an ExceptionInInitilizerException down the line (see ProofSettings comment)

unp1 (changes requested):
First, thanks a lot! Just some minor issues. See comments

key.core/src/main/java/de/uka/ilkd/key/settings/ProofIndependentSettings.java
```diff
@@ -30,14 +31,23 @@ public class ProofIndependentSettings {
...
-        if (file.exists()) {
-            DEFAULT_INSTANCE = new ProofIndependentSettings(file);
-        } else {
-            var old = PathConfig.getProofIndependentSettings();
-            DEFAULT_INSTANCE = new ProofIndependentSettings(old.toFile());
+        var write = PathConfig.currentPaths.proofIndependentSettings;
+        var differs = PathConfig.isDifferentReadWriteDirectories();
+
+        // Path to read from
+        var read = differs ? PathConfig.currentPaths.proofIndependentSettings : write;
```
unp1: Both branches of this ternary resolve to the same value: `write`
The `differs` branch should point at the previous dir:

```suggestion
        var read = differs ? PathConfig.previousPaths.proofIndependentSettings : write;
```

`ProofSettings` is unaffected — it uses `getProofSettingsReadPath()`, which already includes the `previousPaths` files in its search list.

unp1: The last one to address :-)

unp1 (approved):
Thanks!

PR #3891, opened 2026-07-05
Fix two soundness gaps in select/permission taclets

unp1:
## Intended Change

Two independent soundness issues in taclets, both discovered while generating  taclet proof obligations (`TacletProofObligationInput`).

### 1. `simplifySelectOfMemset(EQ)`: missing cast on the stored value

The rule misses a cast (`(beta)x`) in the then-part of the conditional, cf. the `selectOfMemset` axiom. Without the cast the taclet proof obligation is not valid for `x` outside `beta`.

### 2. `permissionRules.key`: twelve assumes-rewrites without update-context restriction

Taclets have an assumes and a find term that matches on term (not top level), hence, the `\sameUpdateLevel` modifier is needed.

## Plan

* [x] Fix `simplifySelectOfMemset(EQ)` (cast) and add `\sameUpdateLevel` to the twelve permission taclets
* [ ] Add `\lemma` annotations + taclet soundness proofs (`tacletProofs/`) for the affected rules, where the lemma translation supports them (the `\addrules`-based memset rules may not be translatable)

## Type of pull request

## Ensuring quality

- I have tested the feature as follows: full `testRAP` green (in particular all `permissionHeap*` loadability and provability proofs, which exercise the twelve permission rules heavily), `testProveRules` green.
- I have checked that runtime performance has not deteriorated (RAP wall time unchanged).

### DIFF
--- a/key.core/src/main/resources/de/uka/ilkd/key/proof/rules/heapRules.key
+++ b/key.core/src/main/resources/de/uka/ilkd/key/proof/rules/heapRules.key
@@ -553,7 +553,7 @@
 
         \replacewith(
             \if(elementOf(o, f, s) & f != java.lang.Object::#$created)
-                \then(x)
+                \then((beta)x)
                 \else(select<[beta]>(h, o, f))
             = sk ==>)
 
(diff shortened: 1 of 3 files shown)


wadoon (approved):

WolframPfeifer:
This PR once again makes me wonder if `\sameUpdateLevel` should not be the default ...

PR #3892, opened 2026-07-05
Fix GUI crashes when no proof is loaded, and remove dead MainWindowTabbedPane

unp1:
## Related Issue

Not related to a specific issue.

## Intended Change

This PR fixes two GUI crashes that happen when no proof (or a proofless task) is involved.

* **`getSelectedGoal()` returns `null` when no proof is loaded, instead of throwing.** `KeYSelectionModel.getSelectedGoal()` threw `IllegalStateException("No proof loaded.")` when no proof was selected, even though its own javadoc promises to return `null` when there is no goal, and its sibling getters `getSelectedProof()` / `getSelectedNode()` are nullable. Callers therefore null-check it and crash when it throws — e.g. a selection listener firing when the last proof is closed:

  ```
  IllegalStateException: No proof loaded.
      at de.uka.ilkd.key.core.KeYSelectionModel.getSelectedGoal(KeYSelectionModel.java:175)
      at de.uka.ilkd.key.core.KeYMediator.getSelectedGoal(KeYMediator.java:442)
      at de.uka.ilkd.key.gui.InfoView$1.lambda$update$0(InfoView.java:90)
  ```

  It now returns `null`, consistent with the documented contract and the nullable siblings. This fixes the whole class of callers at once (~30) rather than one at a time.

* **`DefaultTaskFinishedInfo.toString()` tolerates a null proof.** A task can legitimately finish without a proof — `ProblemLoader` constructs the info with a `null` proof for a proofless/failed load, and `getProof()` is nullable. `toString()` (which builds the status-bar message) dereferenced `proof` unconditionally and threw on the AWT thread:

  ```
  NullPointerException: Cannot invoke "de.uka.ilkd.key.proof.Proof.isDisposed()" because "this.proof" is null
      at de.uka.ilkd.key.prover.impl.DefaultTaskFinishedInfo.toString(DefaultTaskFinishedInfo.java:138)
      at de.uka.ilkd.key.gui.WindowUserInterfaceControl.taskFinishedInternal(WindowUserInterfaceControl.java:217)
  ```

  It now returns `"No proof"` when `proof` is `null`, honouring the nullable-proof contract.

* **`InfoView` clears itself when no proof is selected.** As the persistent global selection listener, `InfoView.update()` left stale content when neither a goal nor a proof was selected (i.e. after the last proof closed). It now explicitly clears the info tree in that case (and fetches the selected goal once instead of twice).

* **Remove the unused `MainWindowTabbedPane`.** This class is dead code as it was replaced by the DockingFrames layout and is no longer used. This PR removes the class.

## Type of pull request

## Ensuring quality

- I have tested the feature as follows: manual test, neither exception is thrown any more when a proofless task finishes / a selection event fires with no proof loaded, and the InfoView clears when the last proof is closed; `spotlessCheck` is clean.

Generated with AI tooling support


key.core/src/main/java/de/uka/ilkd/key/prover/impl/DefaultTaskFinishedInfo.java
```diff
@@ -135,6 +135,12 @@ public Proof getProof() {
      */
     @Override
     public String toString() {
+        if (proof == null) {
```
wadoon, line 142: Would it not be best to add nullness annotation to this class now?

wadoon (approved):

PR #3893, opened 2026-07-05
RFC / KEP-1: Taclet-generating transformers (lemmas instead of metaconstructs, built-in rules, etc.)

unp1:
**Status: RFC / proof-of-concept.** 

Companion documentation (concept, GUI walkthrough,
beyond-OSS): [https://keyproject.github.io/key-docs/keps/kep-0001-taclet-generating-transformers/](https://keyproject.github.io/key-docs/keps/kep-0001-taclet-generating-transformers/).

Reference to Discussion 2019 (original idea @wadoon): [Taclet generation Language](https://git.key-project.org/key/key/-/wikis/Taclet-Generation-Language) (at the moment internal only)

Addresses similar issue as #3707 by @Drodt but from a different angle
KEP moniker coined by @WolframPfeifer ;-)

## Motivation

Several parts of KeY perform many calculus steps at once in Java: term
transformers (metaconstructs like `#add`) and built-in rules like the One Step
Simplifier (OSS). 

Their advantage is performance and the ability to perform complex transformations.
But they are implemented in Java and therefore harder to inspect as a taclet.

This PR explores a middle path: instead of a transformer that mutates the
proof, a **generator** produces a specialized, metaconstruct-free **taclet**
that captures the transformation. It is **introduced** (a recorded step),
**applied** (an ordinary recorded taclet application), and **provable** (a
soundness obligation is created and discharged in the base calculus). The
performed transformation becomes an inspectable, separately certifiable
artifact — and a changed result surfaces immediately, because the recorded
name hashes the outcome.

The state of the implementation is suitable for trying it out and discussing 
whether to carry it forward or not. It is not finalized. Also the show case in 
this example is the OSS, the framework is more general and not restricted to OSS.

In principle one could write complex transformations that establish Groebner bases in
one step and proof that transformation later correct as part of the soundness proof
obligation for the taclet.

## What this changes, at a glance

- `de.uka.ilkd.key.rule.lemma` — the framework: `LemmaTacletGenerator`
  (contract: context-free via `\assumes`, in the provable fragment,
  deterministic content-derived names), `LemmaIntroductionRule` (built-in),
  `GeneratedLemmaRegistry`, `GeneratedLemma` (lazy soundness obligation),
  `LemmaProver` (batch discharge), `TransparentProofSaver`,
  `LemmaElaboratingProofReplayer` (a-posteriori elaboration).
- `OssLemmaGenerator` + `OneStepSimplifier.computeSimplification(...)` — OSS as
  the first real generator (a pure extract-method exposing its core; no change
  to opaque OSS behavior).
- Strategy option **One Step Simplification: Enabled / Disabled / Transparent**.
- `ProofCorrectnessMgt` tracks unproven generated lemmas
  (`CLOSED_BUT_LEMMAS_LEFT`).
- GUI: a **Lemmas** tab in the proof-management dialog (per proof, grouped by
  generator + content, batch-provable), `Save Transparent Proof…`, and an
  "abandon all proofs of an environment" cleanup.
- CLI: `--save-transparent`.

## Two proof steps 

`p & true & (q | true) -> true & p`, OSS **Enabled**:

```
0: One Step Simplification      (one opaque node, hides 5 rewrites)
  1: closeTrue
```

Same problem, OSS **Transparent**:

```
0: introduce_ossLemma
  1: ossLemma_fbf1a98ed8f4__0    // \find(p & true & (q|true) -> true & p) \replacewith(true)
    2: closeTrue
```

## How generators map to lemmas

`OssLemmaGenerator` turns an aggregated `F ↝ F'` into a ground taclet; the
context formulas of OSS's replace-known steps become an `\assumes` clause with
`\inSequentState` (mirroring OSS's own discipline). Modality-containing
formulas are outside the provable fragment and stay with the ordinary calculus.
Obligations are created lazily (off the search path) and proved with OSS **off** from the individual base rules the lemma aggregates, never by the aggregation under scrutiny.

## Two operating modes

- **In search** — the `Transparent` strategy option. In transparent mode OSS is
  applicable nowhere and does not steal its rule sets from the goals, so
  modality-free formulas go through generated lemmas while everything else is
  simplified by the ordinary calculus in individual steps. Proofs contain **no
  opaque OSS node at all**. A small focus-conditional cost penalty makes the
  lemma path win the scheduling race on eligible formulas without disturbing
  symbolic-execution scheduling.
- **A posteriori** — find the proof with fast opaque OSS, then elaborate the
  finished proof into a transparent one (this is what "Save Transparent Proof"
  does). Search stays fast; transparency is produced on demand.

## Design points for discussion

1. **Explicit two-step recording is permanent; implicitness is display-only.**
   No proof-format migration; folding intro+apply into one node would be a
   view filter (not in this PR).
2. **Generated taclets are not yet serialized into the proof file.** Transparent
   proofs load and replay-to-closed, but replay *re-runs the generator*; the
   content-hash names make this deterministic, and a changed result fails loudly
   ("taclet not found"). Serializing the taclets — so a checker needs only the
   taclet text + its obligation and the generator drops out of the trusted base
   — is the natural next step, deliberately out of scope.
3. **Fragment boundary.** Modality-free (incl. update-laden) formulas are
   lemma-eligible; program modalities are not. Generalizing update-prefix
   simplification (schematizing the unchanged modality kernel) would widen the
   fragment and is a designed follow-up, not built here.
4. **Bound-uniqueness check scoped to schema variables.** Taclets generated from
   proof formulas legitimately re-bind concrete variables; the check's own
   documentation says the restriction is about schema variables, so this aligns
   code with intent rather than weakening an invariant.

## Trying it out

- Toy generator: `./gradlew :key.ui:run`, load `lemmagen-demo.key`, right-click a
  `5 + 7` subterm → `introduce_addLitLemma`, then apply the generated taclet.
- OSS in the GUI: set One Step Simplification to *Transparent*, run a FOL/arith
  proof, inspect the `ossLemma_…` steps; open Proof Management → By Proof →
  **Lemmas** to prove the obligations.
- CLI: `./gradlew :key.ui:run --args="--auto --save-transparent <file>.key"`.
- Reuse measurement: `./gradlew :key.core:runOssReuseProbe`.

## Testing

New tests under `de.uka.ilkd.key.rule.lemma` cover: introduction / application /
obligation / replay / reuse (toy); OSS lemmas with and without assumptions,
including update-formula obligations; transparent-mode automode on
computation / SumAndMax / BinarySearch (and that stock mode is unchanged);
non-circular obligations (OSS off); quantifier-obligation generation (regression
for the `DefaultLemmaGenerator` fix); a-posteriori elaboration; batch proving,
content grouping, status transition and no-flooding. `ProveRulesTest` confirms
the shared translation change is inert for the standard taclet base.

## Not for merge as-is / follow-ups

- `lemmagen-demo.key` sits at the repo root for reviewer convenience; move under
  `key.ui/examples/` (or drop) before merge.
- `\rules` serialization of generated taclets (generator-independent checking).
- Cross-proof / environment-level lemma library (reuse across proofs).
- Update-prefix kernel generalization (widen the fragment past modality-free).
- Proof-tree view filter to collapse intro+apply visually.
- The `No proof loaded` selection-listener crash exposed by bulk-abandon is
  fixed upstream in #3892 (not duplicated here).
  
Code with AI tooling support.


wadoon:
There was also a wiki entry on this from the discussion in 2019: https://git.key-project.org/key/key/-/wikis/Taclet-Generation-Language

unp1:
> There was also a wiki entry on this from the discussion in 2019: https://git.key-project.org/key/key/-/wikis/Taclet-Generation-Language

Thanks! I put the link at the top of PR description.

PR #3897, opened 2026-07-06
Interactive Proof Replayer

wadoon:
## Intended Change

A user interface that allows you to replay stored proofs interactively. 

The idea is to (a) allow restoring stored proofs partially, (b) on error try to message your current state into a replayable state (e.g. hide formulas etc.), and (c) restart the proof replay.  

image

## Feature

* [x] Load Proofs from file and store them into a data structure
* [x] Nice beautiful proof tree, similar to the "Proof Tree".
* [ ] Clicking on a tree node allows you to 
  * [x] Apply selected rules on the current proof node
  * [ ] Apply a sub-tree to a proof node
  * [ ] If applied, jump to the (last) corresponding proof node.
* [x] Applied with success => "green" marked tree node
* [x] Applied with error => "red" marked tree node
* [ ] Shows stored information of rule application 
* [ ] Shows why rule application was not applicable

## Type of pull request

## Ensuring quality

  `.github/workflows/tests.yml`


mattulbrich:
I converted to draft due to the number of open check boxes.

PR #3902, opened 2026-07-09
Add user-defined strategy presets to the strategy options pane

unp1:
## Intended Change

The strategy pane offered already a few built-in presets. This PR extends the functionality by allowing the user to save their own presets. 

A "stash" options allows to quickly save the current strategy options so that one can return quickly back to them.  Stashes are tought for quick, non-persistent capturing of the current strategy properties without giving them a name and/or making them persistent.

Functionalities:
- save the current strategy settings as a named preset ("Save current as…"),
- stash them into a single reusable "Stash" slot (overwritten each time),
- rename a preset, and
- delete a preset via a small "x" on its row in the combo box dropdown.

Presets capture the current strategy properties and max. rule
applications, are stored in the proof-independent settings and survive restarts.

## Type of pull request

## Ensuring quality

    
- I have tested the feature as follows: manual GUI tests

### DIFF
--- a/key.ui/src/main/java/de/uka/ilkd/key/gui/StrategySelectionView.java
+++ b/key.ui/src/main/java/de/uka/ilkd/key/gui/StrategySelectionView.java
@@ -4,10 +4,14 @@
 package de.uka.ilkd.key.gui;
 
 import java.awt.*;
+import java.awt.event.MouseAdapter;
+import java.awt.event.MouseEvent;
 import java.util.*;
 import java.util.List;
 import java.util.Map.Entry;
+import java.util.function.Supplier;
 import javax.swing.*;
+import javax.swing.plaf.basic.ComboPopup;
 
 import de.uka.ilkd.key.core.KeYMediator;
 import de.uka.ilkd.key.core.KeYSelectionEvent;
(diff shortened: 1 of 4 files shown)


key.ui/src/main/java/de/uka/ilkd/key/gui/StrategySelectionView.java
```diff
@@ -581,6 +956,125 @@ public Icon getIcon() {
         return IconFactory.PROOF_SEARCH_STRATEGY.get(MainWindow.TAB_ICON_SIZE);
     }
 
+    /**
+     * A single entry of the preset combo box: either the built-in "Defaults", one of the profile's
+     * further built-in presets, or a user-defined preset. User-defined entries can be deleted and
+     * renamed; built-in entries cannot.
+     */
+    private static final class PresetEntry {
```
wadoon, line 1012: Why not wrap the configuration object?

wadoon (commented):
First glance: 

* ENH: Store the selected preferred strategy in the configuration.
* ENH: ProofSettings should be loadable from file (e.g., proof files), and the clipboard. 
  * makes sharing and comparison easier 
* ENH: Copy proof setting to clipboard.
* ENH: Keep the StrategyPanel simple. Add rather a SettingsPanel in the SettingsDialog.

unp1:
Thanks a lot.
> * ENH: Store the selected preferred strategy in the configuration.
> * ENH: ProofSettings should be loadable from file (e.g., proof files), and the clipboard.
>   * makes sharing and comparison easier
> * ENH: Copy proof setting to clipboard.

will do.

> * ENH: Keep the StrategyPanel simple. Add rather a SettingsPanel in the SettingsDialog.

That is a bit against my usage scenario as I want to quickly save (at least stash) settings during case studies. But I agree completely that the strategy pane needs rework. Either an item for the KeY KA meetings or for the KeY Symposium

wadoon:
When is `Apply` required to be pressed?

image

wadoon (approved):
rebased

PR #3904, opened 2026-07-10
Make the strategy cost structure explicit

unp1:
## Intended Change

This PR is a precursor to a careful refactoring of cost values assigned to rules.
It does not change any computed costs and is identical to current main. Its goal is too make
the current cost structure visible and to allow follow up PRs to consolidate them systematically.

The PR makes the cost structure explicit by naming the costs and organizing them in 
three layers (for a detailed explanation see https://keyproject.github.io/key-docs/devel/StrategyCosts/) :

- **`CostBand`**: cost bands express cost tiers for categories that are cross theory 
    like (`CLOSE`, `REWRITE`, `SUBST`, …, `DEFER`, `LAST_RESORT`) with `cost()` and `at(delta)` for fine ordering within a tier. Each tier documents its meaning and, where present, hard ordering constraints (e.g. `BLOCK_CONTRACT` must stay cheaper than `REWRITE` and all
  symbolic-execution rules).
- **`CombinationCost`** (new, `key.core`): costs whose meaning spans theories,
  with the two admission mechanisms documented — (1) rule sets that more than
  one strategy assigns a cost to (conflict-dispatched by
  `ModularJavaDLStrategy`; `ORDERED_REWRITING` for `apply_equations`,
  `CNF_CONVERSION` for `apply_equations_andOr`), and (2) taclets whose rule
  sets belong to different strategies, so the tuned quantity is the *sum*
  (`APPLY_SELECT_EQ_EFFECTIVE` for the `applyEq` taclet; the heap-side constant
  is derived from it, turning a previously comment-only "≈ −5700" coupling into
  checked arithmetic). A taclet/rule-set sweep verified these are the only such
  cases.
- **Theory-local cost tiers** (new, `key.core`): `FOLCost`, the integer-arithmetic
  holders (`PolynomialCost`, `LinearEquationCost`, `LinearInequationCost`,
  `NonlinearArithmeticCost`, `DivModCost` — grouped anticipating a future split
  of the integer strategy), `SymExCost`, `StringCost`, `JavaCardDLCost` and
  `HeapSelectCost` (the pull-out-select pipeline, kept separate to pre-stage a
  future heap/select module). Constants are named for what they achieve;
  related values are expressed as arithmetic on their sibling
  (`MERGE_POINT_SKIP = MERGE_RULE - 1000`). Theory-local constants are
  deliberately absolute (not anchored to a band): retuning a band moves exactly
  the rules placed on it.

The five strategies (`FOLStrategy`, `IntegerStrategy`, `SymExStrategy`,
`StringStrategy`, `JavaCardDLStrategy`) are converted to this vocabulary;
theory-local constants are imported statically so call sites read as bare
names, while `CostBand`/`CombinationCost` stay qualified as a signal that a
shared level is involved.

The commits are per-theory checkpoints (FOL, Integer, SymEx, String,
JavaCardDL, then consolidation and readability) and each compiles and passes
runAllProofs (to check that all proofs stay identical to current main) on its own.

## Type of pull request

## Ensuring quality

- The refactoring is **byte-identical by construction**: every named constant
  equals the literal it replaces, verified per theory.
- Full `runAllProofs` after each theory checkpoint and after the consolidation
  and readability steps: 674/674 proofs, 0 failures.
- Model-Search node-for-node comparison on an arithmetic proof corpus against
  the unmodified strategies: identical proofs (same node counts on all
  compared proofs).
- I made sure that introduced/changed code is well documented (javadoc on every
  band, holder and constant; cross-references between the layers).
- I made sure that new/changed end-user features are well documented: a new
  developer-guide page **"Strategy Costs and the Cost-Band Ladder"**
  (https://keyproject.github.io/key-docs/devel/StrategyCosts/) is pushed to
  key-docs together with this PR; the page carries an open-PR notice
  referencing this pull request, and this PR and the page reference each other.

This PR is deliberately value-preserving. Follow-up PRs (independent of each
other, based on this one) will normalize individual values where the assessment
found them accidental — e.g. the relative order of goal closure and
loop-invariant application, the `Long.MIN_VALUE` / `-20000` sentinel extremes,
and near-duplicate levels — each gated by its own runAllProofs and
node-for-node comparison.

PR created with AI tooling support


key.core/src/main/java/de/uka/ilkd/key/strategy/FOLStrategy.java
```diff
@@ -321,28 +332,30 @@ protected void setupFormulaNormalisation(RuleSetDispatchFeature d) {
...
-                applyTF("assoc1", ff.clauseSet), applyTF("assoc2", ff.clause), longConst(-10)));
+                applyTF("assoc1", ff.clauseSet), applyTF("assoc2", ff.clause),
+                longConst(CNF_RESTRUCTURE + 25)));
 
         bindRuleSet(d, "cnf_dist",
             SumFeature.createSum(applyTF("distRight0", ff.clauseSet),
                 applyTF("distRight1", ff.clauseSet), ifZero(applyTF("distLeft", ff.clause),
-                    longConst(-15), applyTF("distLeft", ff.clauseSet)),
-                longConst(-35)));
+                    longConst(CNF_RESTRUCTURE + 20), applyTF("distLeft", ff.clauseSet)),
```
wadoon, line 353: Ah... the magic 20... of course.

wadoon (approved):
It was a mystery before this PR. And it will be a mystery after this PR.

Drodt (approved):
Thanks!

Issue #3711, opened 2026-01-12
"Reload" loads the wrong file


wadoon:
## Description

The "Reload" action (File -> Reload or C-r) does not load the last opened proof sometimes. 

The action loads the first file in the "Most recent file" menu. The first file does not have to be the last opened file, as a rearrangement of files is missed if the file is already in the list.

## Solution

The "Most recent file" menu should reflect the *history* of the opened file. If a file is opened, it should be in the first place of the list. This requires removing and adding it up front. 

## Reproducible
  
> Is the issue reproducible?
*always*

### Steps to reproduce 

> Describe the steps needed to reproduce the issue.

1. Open two arbitrary KeY files A.key, and B.key.
2. Open A.key again. 
3. The order in the "Most recent files" menu is [B.key, A.key, …]. 
4. Hitting "File -> Reload" now loads B.key again. 

> What is your expected behavior and what was the actual behavior?

File "A.key" is loaded instead and the MRF menu order is [A.keY, B.key, …]

wadoon:
Waiting for #3788

PR #3909, opened 2026-07-10
Fix recent file menu; fix #3898

### LINKED ISSUE #3898: Recent files entries broken
## Description

The dynamically generated menu of recent files used to show the filename only if unique (full path shown in tooltip). If not unique, the least prefix that makes it unique was added. This is not the case any more. Instead, parentheses occur (in parentheses actually seems to be the part of the path that would be unique) and the whole absolute path is shown. Compare the screenshot:

Image
(Note: The longest line is due to the actual filename being very long, and I think there is no really good solution for this problem. However, without the directory prefix it would be shorter.)

In addition, when loading an example shipped with KeY (i.e., a file from inside the jar), it is unzipped into a temp directory. The menu entry refers to this now. Since the entries are persisted in a settings file, and the latest after reboot the temp files are deleted typically, the menu entries point to non-existing files sooner or later.

Maybe it would make sense to switch menu to use URLs instead of paths? SourceView and IssueDialog do that, so the information should be there.

## Reproducible
 
always (at least when starting KeY via the shadow jar, observed to not occur when starting via gradle)

wadoon:
## Related Issue

This pull request resolves #3898 

## Intended Change

1. Make filenames as short as possible, but as long as needed for uniqueness, again.
2. Show profile information directly and options in the tooltip. 
3. Do not show entries if they were temporary and already deleted. 

## Type of pull request

## Ensuring quality

- Looking at it, visually.

### DIFF
--- a/key.ui/src/main/java/de/uka/ilkd/key/gui/RecentFileMenu.java
+++ b/key.ui/src/main/java/de/uka/ilkd/key/gui/RecentFileMenu.java
@@ -4,7 +4,9 @@
 package de.uka.ilkd.key.gui;
 
 import java.awt.event.ActionEvent;
-import java.io.*;
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -19,6 +21,7 @@ import de.uka.ilkd.key.proof.init.Profile;
 import de.uka.ilkd.key.settings.Configuration;
 import de.uka.ilkd.key.settings.PathConfig;
 
+import com.google.common.html.HtmlEscapers;
 import org.jspecify.annotations.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -104,15 +107,7 @@ public class RecentFileMenu {
             insertFirstEntry(entry);
 
             // Recalculate unique names
-            final String[] paths =
-                recentFiles.stream().map(RecentFileEntry::getAbsolutePath).toArray(String[]::new);
-            final ShortUniqueFileNames.Name[] names = ShortUniqueFileNames.makeUniqueNames(paths);
-
-            // Set the names
-            for (int i = 0; i < names.length; i++) {
-                var text = names[i].toString();
-                recentFiles.get(i).createMenuItem().getAction().putValue(Action.NAME, text);
-            }
+            computeShortestUniqueFilenames();
         }
     }
 
@@ -194,8 +189,17 @@ public class RecentFileMenu {
             List<Configuration> recent = file.asConfigurationList();
             this.recentFiles.clear();
             this.mostRecentFile = null;
+
+            String tmpFolder = System.getProperty("java.io.tmpdir");
             for (var c : recent) {
                 final var e = new RecentFileEntry(c);
+
+                // if file not present, and it was stored in the tmp folder,
+                // we allow us to silently ignore it.
+                if (e.path.startsWith(tmpFolder) && !Files.exists(Paths.get(e.path))) {
+                    continue;
+                }
+
                 // The list is stored most-recent-first, so the first entry is the most recent.
                 // (Previously this condition was inverted and overwrote mostRecentFile with every
                 // entry, leaving it null after startup -- issue #3711.)
@@ -205,6 +209,8 @@ public class RecentFileMenu {
                 recentFiles.add(e);
                 menu.add(e.createMenuItem());
             }
+
+            computeShortestUniqueFilenames();
         } catch (FileNotFoundException ex) {
             LOGGER.info("Could not read RecentFileList. Did not find file {}", filename);
         } catch (IOException ioe) {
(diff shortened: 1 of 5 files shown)


key.ui/src/main/java/de/uka/ilkd/key/gui/RecentFileMenu.java
```diff
@@ -194,8 +188,17 @@ public final void loadFrom(Path filename) {
...
             this.recentFiles.clear();
             this.mostRecentFile = null;
+
+            String tmpFolder = System.getProperty("java.io.tmpdir");
             for (var c : recent) {
                 final var e = new RecentFileEntry(c);
+
+                // if file not present, and it was stored in the tmp folder,
+                // we allow us to silently ignore it.
+                if (e.path.startsWith(tmpFolder) && !Files.exists(Paths.get(e.path))) {
```
unp1, line 203: Breaks `RecentFileMenuTest` (#3711): `@TempDir` sits under `java.io.tmpdir` and the test never creates `A.key`/`B.key` on disk, so both match the guard and are skipped and `getMostRecent()` returns null (CI stays red).

One problem tobe aware with `java.io.tmpdir` is that it is unreliable under macOS `/var` is a symlink to `/private/var`, so `java.io.tmpdir` reads as `/var/folders/…` while the same file, once canonicalized, is `/private/var/folders/…`

key.ui/src/main/java/de/uka/ilkd/key/gui/RecentFileMenu.java
```diff
@@ -222,13 +227,41 @@ public String getMostRecent() {
...
+
+        var names = ShortUniqueFileNames.makeUniqueNames(map);
+
+        for (int i = 0; i < actions.size(); i++) {
+            var a = actions.get(i);
+            var p = a.fileEntry.profile;
+            var option = a.fileEntry.additionalOption;
+            a.putValue(Action.NAME, names[i].getName() +
+                    ((p == null || p.equals("null")) ? "" : " (Profile: %s)".formatted(p)));
+            a.putValue(Action.SHORT_DESCRIPTION,
```
unp1, line 257: Path and `Configuration.toString()` are not HTML-escaped, this migth lead to problems with paths containing strange symbols  `<`, `>` or `&`

key.ui/src/main/java/de/uka/ilkd/key/gui/ShortUniqueFileNames.java
```diff
@@ -93,10 +94,10 @@ private static void prepareForInsertionOf(
      * @param files unique list of files
      * @return named files
      */
-    public static Name[] makeUniqueNames(String[] files) {
-        Name[] names = new Name[files.length];
-        for (int i = 0; i < files.length; i++) {
-            Name entry = new Name(files[i]);
+    public static Name[] makeUniqueNames(List<String> files) {
```
unp1, line 97: `makeUniqueNames` expects unique paths, but the same file under two profiles passes its path twice, the shortener then cannot distinguish them and lengthens both to the full path. The `(Profile: …)` suffix already distinguishes them, compute the short name over the *distinct* paths and reuse it for each entry (keeping both entries).

key.ui/src/main/java/de/uka/ilkd/key/gui/RecentFileMenu.java
```diff
@@ -222,13 +227,41 @@ public String getMostRecent() {
         return mostRecentFile != null ? mostRecentFile.getAbsolutePath() : null;
     }
 
+    public void computeShortestUniqueFilenames() {
```
unp1:
```suggestion
    private void computeShortestUniqueFilenames() {
```

key.ui/src/main/java/de/uka/ilkd/key/gui/RecentFileMenu.java
```diff
@@ -222,13 +227,41 @@ public String getMostRecent() {
...
+        var pathSeparator = File.separatorChar;
```
unp1: Unused — remove.
```suggestion
```

unp1 (changes requested):
Thanks. Just a few minor issues. The problem with tmp is not a blocker just sth to be aware of.

key.ui/src/main/java/de/uka/ilkd/key/gui/RecentFileMenu.java
wadoon: But also this test is not good -- AI slob. I made `RecentFileMenu` testable, and did it right. 

The simple fix  would have been a `Files.createFile(...)`.

unp1 (changes requested):
The html rendering in the tooltip seems not to work

Image

WolframPfeifer:
> The html rendering in the tooltip seems not to work
> Image

Weird: For me it renders correctly.

Is this the tooltip of the recent file entry in the menu? Then I don't understand where the `<html>` comes from. The current implementation seems to just add the absolute path (as a String) as tooltip.

unp1:
> > The html rendering in the tooltip seems not to work
>>
> Weird: For me it renders correctly.
> 
> Is this the tooltip of the recent file entry in the menu? Then I don't understand where the `<html>` comes from. The current implementation seems to just add the absolute path (as a String) as tooltip.

I think on macos
```
a.putValue(Action.SHORT_DESCRIPTION,  
                "<html>" + HtmlEscapers.htmlEscaper().escape(a.fileEntry.path) +
                    (option == null ? "" : "<br>" + option)
                    + "</html>");
```
the the string is rendered as normal text not as html leaving the tags to be visible.

Yes, it is from the recent file menu. But maybe that is just a macos problem. I'll approve the PR as that should not be a blocker. We can still push a fix if there is one either on the 3.0 (or 3.x) branch.

unp1 (approved):

PR #3910, opened 2026-07-10
Removal of proof_references and symbolic_execution modules

wadoon:
## Intended Change

Removal of `key.core.symbolic_execution{.example}` and `key.core.proof_references`. 
Both modules do not survive the tests anymore, and will not be fixed.


unp1 (approved):
Thanks!

PR #3914, opened 2026-07-11
Matcher Follow-Up: Remove monolithic matcher for Java programs (compiler) + move more matching logic to ncore

unp1:
## Intended Change

The previous introduction of the matcher based on "compilation" (partial evaluation) used a monolithic program matcher with fall-backs to
the matching code in the AST classes. This PR breaks the monolithic matcher in single instructions and removes any dependency on AST
internal matching methods.

Advantages:
* it prepares for the removal of the interpreter matcher as the compiler does not need its infrastructure for program matching and as safety fallback anymore.
* it moves generic parts of the program matching logic to ncore facilitating reuse across provers
These changes make the compiler adaption easier for other provers like RustyKeY and SolidiKeY

In addition JavaDoc comments were partially reworded to make them easier to read. In particular wording referring to the 
development process of the PR have been removed. They are now more focused on describing the "is" state than on how the "is" 
state was achieved. 

### The framework (`key.ncore.matcher`)

* The module is renamed from `key.ncore.compiler` and the interpreter VM framework moves into it, so both halves of the match framework live in one module (package FQNs unchanged; pure `git mv`). Dependency direction is preserved — nothing in the ncore modules points at `key.core`.
* **Term side.** The single-source plans (`MatchPlan`, `OperatorPlan`, `SchemaVarPlan`, `MatchHead`) already lived here; the dispatch skeleton now joins them: `MatchPlanBuilder` owns the recursion over a find pattern (operator, subterms, bound variables) and the fail-at-load contract. `ModalityHead` is language-agnostic — the modal kind and the program are the modality operator's two syntax children (`Modality.kind()` / `Modality.programBlock()`, an abstraction every prover's modalities share).
* **Program side.** The symmetric half for the program inside a modality: `ProgramMatchPlan` nodes (`ProgramLeafPlan`, `ProgramStructuralPlan`, `ProgramListSVPlan`) with the mixed fixed/list child walk in `ProgramChildSequence` over a generic positional cursor (`ProgramChildCursor`). Plans form a small immutable tree; each node renders its subtree via `toString` for debugging.
* **What a language supplies** enters through three small service provider interfaces plus two dispatches:
  * `BinderMatcher`: how bound variables are bound and unbound. Each operation appears once: `binder(boundVars)` is an element-based instruction shared by both back-ends; `unbind(mc, boundVars)` is a pure state transformation that the framework itself wraps into an interpreter instruction. `unbind` receives the bound variables because closing a scope is not scope-structured in every binding representation — a front-end that keeps a counted stack (e.g. for de-Bruijn-indexed terms) pops exactly `boundVars.size()` entries, while a nested renaming table pops one scope and ignores the argument.
  * `ProgramMatchHook`: how the program of a modality is entered from the term skeleton.
  * `ListSVMatcher`: which program elements a list schema variable may absorb, and how the matched run is recorded.
  * a `MatchPlanBuilder` subclass (schema-variable instructions, per-operator heads, term-label wrap) and a program dispatch (one `instanceof` case per construct).
* Everything else, plan skeletons, child walks, the greedy list-SV run, instruction emission and compilation is shared. The package doc is the plugging-in guide.

### The Java front-end (`key.core`)

* `JavaMatchPlanBuilder` supplies only the Java-DL decisions (schema-variable instruction factory, the head dispatch for elementary updates / parametric function instances / modalities, `JavaBinderMatcher`, the term-label wrap); the recursion and both back-end derivations are inherited.
* `JavaProgramMatchPlanBuilder` describes every Java construct once: fixed-arity structural elements, value leaves, schema type references, dimension-guarded type references / variable specifications, schematic field references, concrete program variables/methods, variable-arity list schema variables, and **context statement blocks** — including the `.. ...` prefix descent, the inner execution context (`.#ex..` / `.#pm@#t(#v)..`, matched by ordinary sub-plans), the active statements, and the recorded prefix/suffix positions. The compiled back-end reads only the structure of pattern and source; it calls **no AST `match(SourceData, MatchConditions)` method at all**.
* The legacy compiled matcher (`JavaProgramCompiler`) is deleted. A program the dispatch does not describe fails at taclet-load time with the framework's clear "no head for this find pattern" error — the same contract the term side has — instead of silently matching through the AST. No construct expressible in the taclet language hits this.
* The interpreter back-end is untouched (its converted-instruction mode and the monolithic `MatchProgramInstruction` remain); it is the reference the compiled matcher is validated against.

## Type of pull request

## Ensuring quality

  * **Byte-identity** validated with a differential oracle that ran the monolithic AST matcher alongside every plan match over the whole RunAllProofs corpus: **53,419,581 match datapoints, 0 mismatches** — re-run after every stage of the series (context integration, list SVs, variable-arity contexts, safety-net retirement, framework extraction, term-side generalization, binder rework). The oracle is development tooling and is kept out of this branch.
  * Instrumented fallback count over the full taclet base: **0** interpreter fallbacks and **0** compiled legacy fallbacks before the net was removed; the removal itself is compile-guaranteed unreachable for the shipped base (RunAllProofs loads every taclet).
  * Full **RunAllProofs** green; full `key.core` test suite green 
  * **Node-identical proofs**: the 8 benchmark proofs of the mt-goals series close with identical node counts before/after, seq wall-time within noise (±3.5 %). Capped smoke runs (100k rule applications) on large SE-heavy proofs (TimSort binarySort/mergeLo/mergeHi/sort, a wide-branching case study) are node-identical as well.
  * The modality head's compiled program entry reads `Modality.programBlock()` — the same object `JTerm.javaBlock()` returns for a modality term; the binder rework routes unbinding through the framework wrapper at the same point in the instruction stream. Both re-validated by the oracle run above.
  * Every commit in the series compiles on its own; matcher unit tests and spotless pass on the tip.
  * Adversarially reviewed: independent line-by-line equivalence reviews of the context-match and list-SV translations against the AST originals (7 review lenses); the two findings they produced are the documented clean-fail-vs-NPE corner and the re-match alignment above.


unp1:
Superseded by https://github.com/KeYProject/key/pull/3922, which is identical (same 17 commits, tip bc26ce719a) but sourced from a branch on KeYProject/key directly instead of the unp1 fork. Re-homed at the maintainers' preference; no code change.

PR #3916, opened 2026-07-11
Add test for Java parsing coverage

unp1:
## Intended Change

This PR adds 48 test cases for Java constructs that are parser supported (it is not checked whether KeY can deal with them on a calculus level).

The tests record the current status quo, this means they fail when a supported construct can no longer be supported and they fail if a construct becomes supported (in the latter case the error message says to change the test case to refelct that a new feature is supported)

One thing that stands out is that we can parse ``var x = 0;``  if the right-hand side is a primtiive type but something with a reference type ``var x = "hi";`` fails.

The tests are an artifact  of determining the Java coverage for KeY 3.0 (see https://keyproject.github.io/key-docs/user/SupportedFeatures/JavaCoverage/)

## Type of pull request

- Other: Adds additional tests

## Ensuring quality
    
- I have tested the feature as follows: The added feature are tests

Created with AI tooling suuport

### DIFF
--- a/key.core/src/test/java/de/uka/ilkd/key/java/JavaFeatureCoverageTest.java
+++ b/key.core/src/test/java/de/uka/ilkd/key/java/JavaFeatureCoverageTest.java
@@ -1,0 +1,143 @@
+/* This file is part of KeY - https://key-project.org
+ * KeY is licensed under the GNU General Public License Version 2
+ * SPDX-License-Identifier: GPL-2.0-only */
+package de.uka.ilkd.key.java;
+
+import java.io.IOException;
+import java.net.URISyntaxException;
+import java.nio.file.Files;
+import java.nio.file.Path;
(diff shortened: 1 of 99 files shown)


wadoon:
Would it not be nice to confirm this via `jacoco`?

unp1:
Not sure I can follow. How would you use Jacoco to measure how much Java constructs are supported by KeY? Measuring how many visit/listener methods are executed?

wadoon:
> Not sure I can follow. How would you use Jacoco to measure how much Java constructs are supported by KeY? Measuring how many visit/listener methods are executed?

You can measure the coverage quantitatively by `jacoco` to monitor the line hits in the `GeneratedJavaParser` class. This gives you a rough percentage of taken paths in the parser, correlating with the grammar rules.

wadoon (approved):

PR #3919, opened 2026-07-12
Support for multi-catch-clauses in try-statements

wadoon:
### Intended Change
This PR implements a new transformer for the `KeYJavaTransformationPipeline` that desugars modern Java exception handling constructs into their equivalent basic forms: 

**MultiCatchReducer** transforms multi-catch clauses into multiple single catch clauses (per JLS 14.20). 

These transformations enable KeY to verify code using these Java 7+ features by reducing them to the core try/catch/finally constructs already supported by the verification infrastructure.

### Technical Details

For a multi-catch:

```java
try {
    body
} catch (ExceptionType1 | ExceptionType2 e) {
    handler
}
```

Produces:
```java
try {
    body
} catch (ExceptionType1 e) {
    handler
} catch (ExceptionType2 e) {
    handler
}
```

Supports:
- Any number of exception types in union
- Mixed single and multi-catch in same try
- Preserves `finally`-blocks
- Marks exception parameters as final (as required by multi-catch semantics)

### Type of pull request
- [x] Refactoring (behavior should not change or only minimally change)
- [x] There are changes to the Java code

### Ensuring quality
- [x] I made sure that introduced/changed code is well documented (Javadoc and inline comments).
    - `MultiCatchReducerTest.java` - Tests for 2-3 exception types, mixed catches, and finally blocks
- [x] I have tested the feature as an example: `key.ui/examples/Java/MultiCatch/`


unp1 (changes requested):
Thanks.
Remaining issues:
1. **`MultiCatchReducer` is not added to the pipeline.** It is defined under `…/transformations/pipeline/` but not registered in `KeYJavaPipeline.createDefault(...)`, so it never runs during loading

2. **Two or more multi-catch clauses in the same `try` are handled incorrectly.** `rewrite(...)` iterates a snapshot (`originalClauses`) but mutates the live `catchClauses` via `remove(i)` / `add(j++, …)`; after the first multi-catch expands, the snapshot index no longer matches the live list. As a result one clause is silently dropped and an un-expanded `UnionType` is left behind (which then fails to load). Reproducer:

   ```java
   public class T {
       void m() {}
       int f() {
           try { m(); }
           catch (IllegalArgumentException | NullPointerException e) { return 1; }
           catch (ArithmeticException | ArrayStoreException e)       { return 2; }
           return 0;
       }
   }
   ```

   With the reducer registered this fails to load with `Unsupported element … UnionType` on the leftover `ArithmeticException | ArrayStoreException`, and the transformer output shows a clause from the first multi-catch is dropped. A single multi-catch (even mixed with single-type catches) works. Fix: track the live insertion position instead of removing by the snapshot index (e.g. rebuild the clause list, or remove the specific node).

3. **The bundled example does not load.** `key.ui/examples/Java/MultiCatch/MultiCatch.java` uses `java.sql.SQLException`, which is not in KeY's default classpath (`JavaRedux`)

4. **Nit:** the Javadoc note about ordering catch clauses by subtype relationship doesn't apply to a multi-catch's own alternatives, they must be disjoint (JLS §14.20), so there's no "more specific first" among them.

wadoon:
> Thanks. Remaining issues:
addressed

key.core/src/main/java/de/uka/ilkd/key/java/transformations/pipeline/MultiCatchReducer.java
```diff
@@ -0,0 +1,88 @@
...
+                    newParam.addModifier(Modifier.DefaultKeyword.FINAL);
+                    newCatchClause.setParameter(newParam);
+
+                    // Clone the body for each catch clause
+                    newCatchClause.setBody(catchClause.getBody().clone());
+
+                    // Insert after the current position (will be before due to reverse iteration)
+                    catchClauses.add(j++, newCatchClause);
+                }
+            }
```
unp1, line 87: This
```
catch (IllegalArgumentException | NullPointerException e) { return 1; }
catch (ArithmeticException | ArrayStoreException e)       { return 2; }
```
becomes
```
catch (IllegalArgumentException e) { return 1; }
catch (ArithmeticException e)      { return 2; }
catch (ArrayStoreException e)      { return 2; }
catch (NullPointerException e)     { return 1; }

```
which would be problematic if the second group would contain a supertype of an exception in the first group

unp1 (approved):

key.core/src/main/java/de/uka/ilkd/key/java/transformations/pipeline/MultiCatchReducer.java
unp1: Fixed it in the latest commit, but please check. Otherwise thanks and I approved the PR.

wadoon:
@mattulbrich @WolframPfeifer This might be something for *soundiness*. 

How do we describe Java programs, which are JLS-invalid but are accepted to a certain extend in KeY?

Invalid would be: 
```
try { m(); }
catch (IllegalArgumentException | NullPointerException | Exception  e) { return 1; }
```

due to violated required type disjointness in union types. 


KeY transformed it and should work with it properly.

PR #3923, opened 2026-07-14
Fix model fields resolution when multiple heaps present

unp1:
## Intended Change

Model fields without a represents or accessible clause could not be used in JavaRedux
as permissions could not build the term (wrong arity).

Found out when trying to specify addSuppressed in Throwable of PR #3918.

This PR should fix it.

## Type of pull request

## Ensuring quality

- I have tested the feature as follows: CI test suite, MErged on local branch with PR 3918 and activated model fields in Throwable

### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/speclang/translation/SLAttributeResolver.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/speclang/translation/SLAttributeResolver.java
@@ -3,7 +3,6 @@
  * SPDX-License-Identifier: GPL-2.0-only */
 package de.uka.ilkd.key.speclang.translation;
 
-import java.util.ArrayList;
 import java.util.List;
 
 import de.uka.ilkd.key.java.JavaInfo;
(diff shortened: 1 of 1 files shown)


wadoon (approved):

key.core/src/main/java/de/uka/ilkd/key/speclang/translation/SLAttributeResolver.java
```diff
@@ -133,7 +136,18 @@ protected SLExpression doResolving(SLExpression receiver, String name, SLParamet
                     final Function fieldSymbol =
                         heapLDT.getFieldSymbolForPV((LocationVariable) attribute, services);
                     JTerm attributeTerm;
-                    if (attribute.isStatic()) {
+                    if (attribute.isModel()) {
```
wadoon, line 139: Does this not depend on `no_state` and `two_state` modifiers?

unp1: do they exist for model fields too? I though just for model methods. Otherwise, I look into that and thanks for bringing it up

wadoon: mhm... On the other side, are we not modelling "model fields" using "model methods"?

unp1: The fix is not 3.0 critical. I have a look into it and will not merge before that is setlled. I'll discuss it with Mattias (he knows model fields/methods best, I think) on Friday at the KeY meeting. In principle, the translation is at least similar and if not should be made identical. But I am not sure whether model fields should be allowed to use the modifiers (from a usage point of view).

mattulbrich: I do not think that we have `two_state` and `no_state` for model fields. We do not desugare model fields to model methods (last time I checked).

unp1: Thanks!

PR #3928, opened 2026-07-14
Remove proof-tree node border

flo2702:
## Intended Change

Removes the ugly border around proof-tree nodes that was introduces during the GUI rework last year in #3599

Before:

Screenshot_2026-07-14_18-41-32

After:

Screenshot_2026-07-14_18-39-52

## Type of pull request

- Non-breaking GUI change


wadoon (approved):

wadoon:
You know, that you can directly push on the prerelease branch, especially in such a case.

WolframPfeifer:
I was wondering: Since #3599 introduced this line that got changed here (I think), what was the reason for this? Does it still look nice with a LaF other than Metal?

PR #3944, opened 2026-07-25
Cache subtype computations in KeYProgModelInfo

unp1:
## Intended Change

Cache subtype relations of the Java model. Speeds up lookup in loading and proof search.
In general, but in particular because at each lookup JavaParser generates and uses pretty printed Strings
to decide assignability between types.

## Type of pull request

## Ensuring quality
    
Part of going through the remnants of different performance fixes.

Partially gnerated/identified with AI support.

### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/java/KeYProgModelInfo.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/java/KeYProgModelInfo.java
@@ -56,6 +56,21 @@ public class KeYProgModelInfo {
     private final JP2KeYTypeConverter typeConverter;
     private final Map<KeYJavaType, Map<String, IProgramMethod>> implicits = new LinkedHashMap<>();
 
+    /**
+     * Caches subtype lookup. First it is a general win, but second JavaParser pretty prints AST
+     * nodes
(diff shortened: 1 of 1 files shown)


key.core/src/main/java/de/uka/ilkd/key/java/KeYProgModelInfo.java
```diff
@@ -507,8 +522,38 @@ private List<Field> getVisibleArrayFields(KeYJavaType arrayType) {
 
     /**
      * returns all proper subtypes of class <code>ct</code> (i.e. without <code>ct</code> itself)
+     *
+     * @return an unmodifiable list of all proper subtypes
      */
     private List<ResolvedReferenceTypeDeclaration> getAllRecoderSubtypes(KeYJavaType ct) {
```
wadoon, line 528: In 20 years, our ancestor will look at the code, and ask, what the heck is "recoder".

wadoon (approved):

key.core/src/main/java/de/uka/ilkd/key/java/KeYProgModelInfo.java
unp1: :-)

PR #3945, opened 2026-07-26
Fix wellFormedEQ* taclets and some improvements for large heaps

unp1:
## Intended Change

1. Fixes some wellFormedEQ* taclets that got out of sync with their counterpart
    - I would suggest we add a taclet flag \generateEQ that automatically greates the EQ version, which would allow us to keep the rule files smaller and to avoid these rules getting out of sync (this is **not** done in this PR)
2. Improves reasoning with large heaps in updates (e.g. as extreme example: heap_array_400 in examples/performance-test/updateSimplification)

## Type of pull request

- Breaking change (fix or feature that would cause existing functionality to change): the rule fixes might break some saved proofs

## Ensuring quality
- I have tested the feature as follows: added additional proofs to runAllProofs and checked that no existing problem deteriorates


Drodt (approved):
Thanks!

PR #3941, opened 2026-07-24
Migrate KeY v3.0 fixes to main

unp1:
## Intended Change

Cherry-picked version fo the fix commits for KeY v3.0 (PR #3925) w/o the changes 
to CI configuration files.

1. Please check if correct.
2. Do we wnat it that way or do your prefer a squash from PR #3925, but then we have to drop there the CI changes

## Type of pull request

- Bug fix (non-breaking change which fixes an issue): The bugfixes on the 3.0 branch

## Additional information and contact(s)

     

The contributions within this pull request are licensed under GPLv2 (only) for inclusion in KeY.

PR #3946, opened 2026-07-27
Forgotten commit from 3.0 release

WolframPfeifer:
In #3941, the commit c7e163d31e was forgotten, so the borders in the proof tree reappeared (see #3928).


unp1 (approved):
Thanks

PR #3949, opened 2026-07-30
Automatically Generate EQ Taclets

Drodt:
## Related Issue

This idea was mentioned in #3945 by @unp1.

## Intended Change

We have many "EQ taclets" in our taclet base that all follow exactly the same schema. However, as they represent essentially a duplication of logic, it is easy for the "normal" taclets and its EQ version to go out of sync. In #3945 this was observed w.r.t. variable conditions.

This PR makes all of that superfluous, as EQ taclets are generated automatically.

To generate an EQ taclet, add the `\generate(\EQ(...))` modifier to it. It takes a term---the term to be put into the `\assumes` part. Let's look at two examples:
```
\lemma
lenOfSeqEmpty {
    \find(seqLen(seqEmpty))
    \replacewith(0)
    \heuristics(concrete)
    \generate(\EQ(seqEmpty))
}
```
generates the EQ taclet
```
\lemma
lenOfSeqEmptyEQ {
    \schemaVar \term Seq EQ;
    \assumes(seqEmpty = EQ)
    \find(seqLen(EQ))
    \replacewith(0)
    \heuristics(concrete)
}
```
Note that all properties of the taclet (annotations, rule sets, etc.) are copied over. The name is automatically extended with "EQ", an `\assumes` part is added, and `seqEmpty` is replaced with `EQ` in the `\find` and (original) `\assumes` parts. (Also for the `\find` of added rules).

You can override the rule sets of the taclet as follows:
```
\lemma
getOfSeqSingleton {
    \schemaVar \term any x;
    \schemaVar \term int idx;
    \find(seqGet<[alpha]>(seqSingleton(x), idx))
    \replacewith(\if(idx = 0)
            \then((alpha)x)
            \else((alpha)seqGetOutside))
    \heuristics(simplify)
    \generate(\EQ(seqSingleton(x) : no_self_application, simplify_enlarging))
}
```
which generates
```
\lemma
getOfSeqSingletonEQ {
    \schemaVar \term any x;
    \schemaVar \term int idx;
    \schemaVar \term Seq EQ;
    \find(seqGet<[alpha]>(EQ, idx))
    \replacewith(\if(idx = 0)
            \then((alpha)x)
            \else((alpha)seqGetOutside))
    \heuristics(no_self_application, simplify_enlarging)
}
```

One side effect of all this: Some taclets change names, as the EQ was not at the end of its name previously. Personally, I prefer the consistency.

## Plan

* [x] Document code in `TacletPBuilder`
* [x] Discuss syntax
* [ ] Once approved: Document in KeY Docs

## Type of pull request

## Ensuring quality

    
- I have tested the feature as follows: Examined generated `taclets.old.txt`


unp1 (changes requested):
HeapSimplificationMacro and SupportedTaclets use the old names of subsetOf... and reach... that did not follow the normal naming approach

Same for saved SMT settings in: 
key.ui/examples/completionscopes/ and key.core/src/test/resources/testcase/merge/

key.core/src/main/java/de/uka/ilkd/key/rule/tacletbuilder/NoFindTacletBuilder.java
```diff
@@ -80,4 +80,18 @@ public NoFindTaclet getTaclet() {
...
+        var rb = new NoFindTacletBuilder();
+        rb.setAnnotations(tacletAnnotations);
+        rb.setAssumesSequent(assumesSeq);
+        rb.setChoices(choices);
+        rb.setDisplayName(attrs.displayName());
+        rb.setName(name);
+        rb.setTacletGoalTemplates(goals);
+        rb.setTrigger(attrs.trigger());
+        rb.setRuleSets(ruleSets);
+        return null;
```
unp1: return rb;

key.core/src/main/java/de/uka/ilkd/key/rule/tacletbuilder/AntecTacletBuilder.java
```diff
@@ -94,4 +94,32 @@ public AntecTaclet getAntecTaclet() {
...
+            rb.addVariableCondition(vc);
+        }
+        rb.addVarsNew(varsNew);
+        rb.addVarsNotFreeIn(varsNotFreeIn);
+        for (var vc : varsNewDependingOn) {
+            rb.addVarsNewDependingOn(vc.first(), vc.second());
+        }
+        rb.setFind((Sequent) find);
+        rb.setChoices(choices);
+        rb.goal2Choices =
```
unp1: goalchoices can be null when using \closegoal so we might get here an NPE, ok most likely not for the kind of taclets where we apply generateEQ

key.core/src/main/java/de/uka/ilkd/key/rule/tacletbuilder/RewriteTacletBuilder.java
```diff
@@ -92,4 +92,33 @@ public void addGoalTerm(JTerm goalTerm) {
...
+            rb.addVariableCondition(vc);
+        }
+        rb.addVarsNew(varsNew);
+        rb.addVarsNotFreeIn(varsNotFreeIn);
+        for (var vc : varsNewDependingOn) {
+            rb.addVarsNewDependingOn(vc.first(), vc.second());
+        }
+        rb.setFind((JTerm) find);
+        rb.setChoices(choices);
+        rb.goal2Choices =
```
unp1: goalchoices can be null when using \closegoal so we might get here an NPE, ok most likely not for the kind of taclets where we apply generateEQ

key.core/src/main/java/de/uka/ilkd/key/rule/tacletbuilder/SuccTacletBuilder.java
```diff
@@ -90,4 +90,32 @@ public SuccTaclet getSuccTaclet() {
...
+            rb.addVariableCondition(vc);
+        }
+        rb.addVarsNew(varsNew);
+        rb.addVarsNotFreeIn(varsNotFreeIn);
+        for (var vc : varsNewDependingOn) {
+            rb.addVarsNewDependingOn(vc.first(), vc.second());
+        }
+        rb.setFind((Sequent) find);
+        rb.setChoices(choices);
+        rb.goal2Choices =
```
unp1: goalchoices can be null when using \closegoal so we might get here an NPE, ok most likely not for the kind of taclets where we apply generateEQ

key.core/src/main/resources/de/uka/ilkd/key/proof/rules/sequence/seqRules.key
```diff
@@ -309,6 +266,7 @@
         \replacewith(\if(from < to) \then(to - from) \else(0))
 
         \heuristics(simplify)
+        \generateEQ(seqSub(seq, from, to) : obsolete)
```
unp1: rule set not what it is at the moment

unp1 (changes requested):
Thanks a lot! Only smaller changes.

One general comment:
Should we check whether the term given in generateEQ occurs in the find term, otherwise a wrong taclet might be easily generated (or is that done and I overlooked it)

key.core/src/main/resources/de/uka/ilkd/key/proof/rules/sequence/seqRules.key
Drodt: True. That's a point for discussion. I did not really implement a path for setting the rule sets to empty; that seemed pointless to me. `obsolete` at least keeps it out of automation.

Either we find a suitable rule set or we bring back the old EQ version.

Drodt:
> Should we check whether the term given in generateEQ occurs in the find term, otherwise a wrong taclet might be easily generated (or is that done and I overlooked it)

That's not done. I can add it.

key.core/src/main/resources/de/uka/ilkd/key/proof/rules/sequence/seqRules.key
unp1: For this one, I am actually not sure if it is not just an oversight that it is not in simplify as well, there is in my opinion no reason why not. 

If the need arises to have an empty marker we can do it then, "obsolete" has more the connotation that the rule is free to be deleted e.g. at the next major version or so.

Drodt:
> subsetOf

Done

key.core/src/main/resources/de/uka/ilkd/key/proof/rules/sequence/seqRules.key
Drodt: Done

unp1 (approved):
Thanks!

unp1 (approved):

mattulbrich:
Looks great! Thanks!

PR #3952, opened 2026-08-01
Determinism hardening

unp1:
## Intended Change

Replaces some HashMaps with potential determinism problems by LinkedHashMaps and invalidates
cost reuse cache entries when strategy changes during proof (by a user).

Minor cleanup for EqualtyConstraint: removal of unused join cache

## Type of pull request

## Ensuring quality
    
- I have tested the feature as follows: test suite, multiple RAP runs to ensure determinism

### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/java/ServiceCaches.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/java/ServiceCaches.java
@@ -18,6 +18,7 @@ import de.uka.ilkd.key.proof.TermTacletAppIndex;
 import de.uka.ilkd.key.proof.TermTacletAppIndexCacheSet;
 import de.uka.ilkd.key.rule.metaconstruct.arith.Monomial;
 import de.uka.ilkd.key.rule.metaconstruct.arith.Polynomial;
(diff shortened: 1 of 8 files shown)


wadoon:
I do not think it is a wise idea to change `HashMap` to `LinkedHashMap`. This is just a pushing of non-determinism to the front. The order then just depends on `put` order. Would a `TreeMap` not be wiser?

unp1:
> I do not think it is a wise idea to change `HashMap` to `LinkedHashMap`. This is just a pushing of non-determinism to the front. The order then just depends on `put` order. Would a `TreeMap` not be wiser?

* Problem in the Macro case is that TreeMap have O(log(n)) and not amortized constant like HashMaps that has performance impacts on code running during proof search.
* For the LoopContractImpl, performance should be a non-issue so I can use the nameHash value as Sorting key in case of a TreeMap (will check)

wadoon:
But in the end, we are not really hardnening KeY against non-determinism. We are just pushing this upfront, and in the end the order is determined by, e.g., the order of definition in KeY file etc...

wadoon (approved):

wadoon (approved):

PR #3956, opened 2026-08-03
modify the Hash of `ResolvedTypeWrapper`

PiIsRational:
## Intended Change

Modify the Hash and equality functions of `ResolvedTypeWrapper`, such that Reference Types with the same 
qualified name get treated as the same type. This tries to resolve a problem in the current implementation
that also uses the `equals` and `hash` methods of the underlying `ResolvedType`, which can fail in certain cases 
for two elements which correspond to the same type. As an example with this change, the test case [var-reference-type](https://github.com/KeYProject/key/tree/main/key.core/src/test/resources/de/uka/ilkd/key/java/coverage/rejected/var-reference-type), which is currently rejected, passes.

## Type of pull request

## Ensuring quality

    
    - I have run the test suite


wadoon (approved):

WolframPfeifer:
Thanks for this PR, @PiIsRational! Looks like the tests are successful, and I (and apparently also @wadoon) think that the changes make sense. The only thing left would be formatting (you can run the gradle task `spotlessJavaApply`, to fix that).

If this is ready, you can remove the Draft status.

PR #3958, opened 2026-08-09
introduce the "script-aware script" button in the auto selection

mattulbrich:
## Intended Change

As planned: Add the script-aware auto macro as one of the options in the dropdown menu

## Plan

This has been commented out in the sources. But now with scripts available in 3.0, this button should also be available in 3.0.

I suggest this be backported to 3.0.1 as well.

Done:
* Added a "J" version of the button for the script aware macro
* Fixed the outline computation in the icon factory, and added 2 px more space

## Type of pull request

- New *micro* feature (non-breaking change which adds functionality)

## Ensuring quality

- I checked the macro behaved as intended.

Should go to 3.0.1!


WolframPfeifer (approved):
Thanks!

Two minor remarks: It was not really clear what the "J" stands for. I guess "JML script"? However, I don't have a better idea, since "S" is already taken.

The baseline of the letter looks a bit high now imo, but this is personal taste:
Image

PR #3960, opened 2026-08-12
Definitional Skolem constants

unp1:
As mentioned  at the devloper meeting. This is the first PR of three cleaning up the AI supported work.

Make definitional Skolem constants aware of their introduction time (removes need for caches) and
formalizes the notion of definitional formerly known as small constants. Definitional Skolem constants
are smaller then most terms in termordering. 

Speeds up Skolem heavy proof by up to 10% bet, 5% on the whole runAllFunProof collection.

## Type of pull request

## Ensuring quality

    
- I have tested the feature as follows: Several RAP runs, unit tests
- I have checked that runtime performance has not deteriorated. It improved.


Drodt (approved):
Thanks!

PR #3961, opened 2026-08-12
Quantifier and Heap Reasoning Improvements

### LINKED ISSUE #3972: Incompleteness of quantifier treatment
## Description

KeY is weak on nested `∀∃` quantifications and cannot proof simple statements that cvc5 and z3 can do in .01 s.

## Reproducible
  
```
 a.length = i_0,
 \forall int u; (u < v_0 & u >= 0 -> \exists int j; (j < i_0 & j >= 0 & u = a[j]))
==>
 \forall int u;
   (u >= 0 & u < v_0 -> \exists int j; ((j >= 0 & j < a.length & u = a[j])<<SC>>))
```

Image

is a MVE sequent that cannot be closed by KeY's automation. It can be closed by manual interaction and by SMT solvers.

### Steps to reproduce 

The above example can be encoded into a KeY file. User interaction is just skolemisation of u and then obvious instantiation.

### Additional information

The automation somehow gets stuck and does not instantiate the LHS ∀u with the skolem obtained from the RHS ∀u. The trigger mechanism seems incomplete here.

This is not a soudness issue, but a completeness issue.

---

* Commit: 555a763233a8e17553af5e90c1db94b137fdcc9a

unp1:
Second of three of the last AI supported work 

* Improves the quantifier heuristics and heap reasoning by providing triggers if array access is an affine integer expression.
The following problem
```
\programVariables {
    int[] a;
    int[] b;
    int k;
    int m;
    int n;
}

\problem {
       \forall int t; (0 <= t & t < n -> b[k + t] = a[m + t])
   ==> \forall int x; (k <= x & x < k + n -> b[x] = a[x - k + m])
}
```
can now be proven automatically.

* It also makes inequality of objects explicit in the sequents enabling more heap simplification rules to be applied automatically. The added taclet has been proven sound and the proof has been added.

* This PR also addresses issue #3972 using the reworked infrastructure of this PR.

## Type of pull request

## Ensuring quality

    
- I added new test case(s) for new functionality. New problem for RAP being able to close an Adjacency list problem which stayed open before.
- I have tested the feature as follows: RAP, testing
- I have checked that runtime performance has not deteriorated. No regressions on RAP, can solve more proofs now. In the worst case the Strategy Options allow to switch back to the old behavior (select GOOD)


Drodt (approved):
Thanks!

PR #3962, opened 2026-08-13
Formalize LocSet as Set of Locations

Drodt:
## Intended Change

Rely on the KeY formalization of parametric `Set` to express `LocSet`. We have:
```
\alias Loc = Pair<Object, Field>;
\alias LocSet = Set<Loc>;
```
Rules are adjusted accordingly.

Additionally, this PR provides
- Many additional rules for sets and locsets (proven sound with KeY by @unp1)
- Additional strategies for sets and locsets, allowing easy use for other KeY instances in the future
- Proper LDT support for parametric sorts
- Improved and fixed matching for parametric functions with parametric sorts
- `Pair` is now parametric and has some additional rules
- SMT translation and Well-Definedness checks now consider set operations, not just LocSet

## Plan

* [x] Fix tests.

## Type of pull request

- Other: 

## Ensuring quality

    

- I have tested the feature as follows: Tested using existing test suite


key.core/src/main/java/de/uka/ilkd/key/ldt/ParametricLDT.java
```diff
@@ -0,0 +1,65 @@
...
+     *
+     * Is it possible to implement LDTs as singletons? (Kai Wallisch 04/2014)
+     */
+    public static Map<Name, ParametricLDT> getNewLDTInstances(Services s) {
+        // TreeMap ensures the map is sorted according to the natural order of its keys.
+        Map<Name, ParametricLDT> ret = new TreeMap<>();
+
+
+
+        return ret;
```
unp1, line 51: That should now contain the SetLDT and/or LocSetLDT?

key.core/src/main/java/de/uka/ilkd/key/ldt/LDT.java
```diff
@@ -33,98 +24,43 @@
...
     // -------------------------------------------------------------------------
 
     protected LDT(Name name, TermServices services) {
-        sort = services.getNamespaces().sorts().lookup(name);
+        super(name);
+        var sort = services.getNamespaces().sorts().lookup(name);
         if (sort == null) {
-            throw new RuntimeException("LDT " + name + " not found.\n"
-                + "It seems that there are definitions missing from the .key files.");
+            sort = services.getNamespaces().sortAliases().lookup(name).aliasedSort();
```
unp1: if "services.getNamespaces().sortAliases().lookup(name)" set is null we get an NPE instead of a telling error message (also of course it should never the case).

unp1 (changes requested):
Thanks a lot! Great work. Only a few minor remarks/questions.

key.core/src/main/java/de/uka/ilkd/key/ldt/ParametricLDT.java
unp1: Solved (?) that in commit 0f62a74c14c541f1cb1b04f8acc603c755fd9e08

unp1:
Please check if you want to keep the last two commits otherwise revert
They reeanble automatic proving of setExamples.key by moving the succedent version setEquaityToElementRight to generic sets (before only for locsets). That is not the general setEqualityToElement rule, which I think you did not want to be enabled automatically). If that includes also the more targeted right version, we can simply move it back or disable it for generic sets and add a duplicate for location sets.

The second commit restricts the elementOf decomposition rules to those cases where the decomposition is necessary. Such that the decomposition rules are only applied where it makes sense and otherwise using the normal one without decomposition. 

Feel free to revert both commits or to only pick what you want.

unp1 (approved):
Thanks a lot. I approved. But please make sure you are fine with my last commit (I wanted to have it done yesterday, but decided to get some sleepp at 5 a.m.) It makes the strategies a bit more useful (comm, assoc and distribute are now set strategy internal not relying on the quantifier heuristic) as well as additional simplification rules (+ their soundness proofs). Sorry that the PR adds now more lines then before, but that is misleading, still much more deleted and the vast majorit are the saved soundness proofs.

mattulbrich:
Thanks, guys for your effort here! Sounds very reasonable and I am looking forward to having a decent set datatype in KeY!

key.core/src/main/resources/de/uka/ilkd/key/proof/rules/locset/locSetsRules.key
```diff
@@ -360,6 +361,267 @@
...
+        \replacewith(f1 != f2)
+        \heuristics(concrete)
+        \displayname "disjointAllObjects"
+    }
+
+    \lemma
+    arrayRangeEmpty {
+        \schemaVar \term Object o, o2;
+        \schemaVar \term int lower, upper;
+        \assumes( ==> o = o2)
```
Drodt: Why do we assume this? `o2` appears only here. Is it an error or for performance?

Drodt: If it is not an error, I can undo my last commit ([9eeedc6](https://github.com/KeYProject/key/pull/3962/commits/9eeedc6f953947515c5b530c087b31eee66b9384))

unp1: The problem was that I could not proof the taclet without the assumption (Z3 could), I did not find the proof argument that there are at least two elements of type java.lang.Object :-) The only one I could construct is 'null'.  I am not sure if Z3s ability to close the proof came from translating taclets (in which case I could just not combine it) or some hard encoded information that the set of objects infinite.

I wanted to discuss it first, if everyone is fine to add an axiom stating that the java.lang.Object domain is infinite. I would like to state it as a general property (and not to hardcode it as a wellformed property). The wellformedness was also why I could not use instanceAllocation to get a second object. 

But in the meantime I did not want to give up on the taclet and used the assumes as a workaround.

unp1: But you seem to have managed it 👍  So no objections from my side, the taclet is much more powerful that way.

PR #3963, opened 2026-08-13
Small feature: go to node with given serial number

WolframPfeifer:
This PR adds a small feature: An entry "Go to Node..." to the proof tree settings menu (gear symbol) and the proof menu which prompts the user jumps to the proof node with the given serial number (if it exists in the proof).

This is particularly useful with the slicing and caching extensions (but also with proof replay), when you receive an error message with a node number.

Note that one could also use the "Search in Proof Tree" as a workaround (which I noticed only after implementing the feature), but this is less precise: If you search for "100" for instance, also the nodes "1001", "1002", ..., "1100", etc. are found.

## Type of pull request

## Ensuring quality

- I have tested the feature as follows: Manually via the GUI

### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/proof/NodeInfo.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/proof/NodeInfo.java
@@ -85,7 +85,7 @@ public class NodeInfo {
     private boolean uselessApplication = false;
 
     /** User-provided plain-text annotations to the node. */
-    private String notes;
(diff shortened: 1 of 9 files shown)


mattulbrich:
Nice feature! Thanks!
I have added a bit of functionality which should be self-explanatory by trying it out.
[curated ChatGPT code]

mattulbrich (approved):
Like it! thanks!

@WolframPfeifer please check the changes are improvements.

PR #3966, opened 2026-08-15
Migrate to JMLtk

wadoon:
## Intended Change

With this PR, `wadoon/key-javaparser` becomes obsolete in favour of the JMLtk fork!

This PR **does not**

* use JMLtk to parse JML.
* use JML-AST classes. 

**Regressions:**

* Positioning on invalid `ExpressionStmt` (`int x = ;`)

**Changes to `key-javaparser`**

* JML modifiers are printed escaped => (`/*@ ghost */ int x;`)
* `KeyRangeExpression` gone
* `KeyMarkerStatement` gone
* Consistent names are in favour of JavaParser (e.g., `XStatement` vs. `XStmt`)

## Plan

Things to do after review: 

* decide when to make JMLtk first release
  * update `K` counter on JMLtk

* discuss sacrifices in JMLtk for KeY compat: 
  `/*@ghost*/ int #x;` vs. `ghost int #x;`

* How to proceed further with exchanging JML parsing

## Type of pull request

## Ensuring quality
    
- I have tested the feature as follows: unit-tests


wadoon:
Squash before merge

key.core/pipelineTests/records/expected/01_TextblockTransformer/SimpleRecord.java
```diff
@@ -1,4 +1,4 @@
-public record SimpleRecord(/*@ nullable */ String name) implements Serializable {
+public record SimpleRecord([[JML modifiers]] String name) implements Serializable {
```
unp1, line 1: That does not look right

key.core/pipelineTests/records/expected/02_MultiCatchReducer/SimpleRecord.java
```diff
@@ -1,4 +1,4 @@
-public record SimpleRecord(/*@ nullable */ String name) implements Serializable {
+public record SimpleRecord([[JML modifiers]] String name) implements Serializable {
```
unp1, line 1: I think that should also be nullable

key.core/pipelineTests/records/expected/03_TryWithResourceReducer/SimpleRecord.java
```diff
@@ -1,4 +1,4 @@
-public record SimpleRecord(/*@ nullable */ String name) implements Serializable {
+public record SimpleRecord([[JML modifiers]] String name) implements Serializable {
```
unp1, line 1: /*@ nullable @*/

key.core/pipelineTests/records/expected/04_EnumClassBuilder/SimpleRecord.java
```diff
@@ -1,4 +1,4 @@
-public record SimpleRecord(/*@ nullable */ String name) implements Serializable {
+public record SimpleRecord([[JML modifiers]] String name) implements Serializable {
```
unp1, line 1: /*@ nullable @*/

key.core/pipelineTests/records/expected/05_RecordClassBuilder/SimpleRecord.java
```diff
@@ -2,10 +2,10 @@
 public final class SimpleRecord extends Record {
 
     @javax.annotation.processing.Generated("RecordClassBuilder")
-    private final /*@ nullable */ String name;
+    private final [[JML modifiers]] String name;
```
unp1, line 5: /*@ nullable @*/

key.core/pipelineTests/records/expected/05_RecordClassBuilder/SimpleRecord.java
```diff
@@ -2,10 +2,10 @@
...
 
     @javax.annotation.processing.Generated("RecordClassBuilder")
-    public final /*@ nullable */ String name() {
+    public final [[JML modifiers]] String name() {
```
unp1, line 8: /*@ nullable @*/

key.core/pipelineTests/records/expected/06_JMLTransformer/SimpleRecord.java
```diff
@@ -2,10 +2,10 @@
 public final class SimpleRecord extends Record {
 
     @javax.annotation.processing.Generated("RecordClassBuilder")
-    private final /*@ nullable */ nullable String name;
+    private final [[JML modifiers]] /*@ nullable */ String name;
```
unp1, line 5: /*@ nullable @*/

key.core/pipelineTests/records/expected/06_JMLTransformer/SimpleRecord.java
```diff
@@ -2,10 +2,10 @@
...
 
     @javax.annotation.processing.Generated("RecordClassBuilder")
-    public final /*@ nullable */ nullable String name() {
+    public final [[JML modifiers]] /*@ nullable */ String name() {
```
unp1, line 8: /*@ nullable @*/

unp1 (changes requested):
Thanks! I think all the mentioned issues are just one comment. Otherwise the changes to error reporting seem to be a regression in the reported position, so we should fix that position reporting not the tests. I'll look into that myself.

unp1:
Improved the position recording slightly. But the parser now reports "int" as last successfull token and no longer "=", so the position is no longer as precise as before where the dialog underlined the empty space after '='

unp1:
Sorry my fault, restored to open.

First pushed to the wrong branch due to a typo jmtk and then deleted the correct one....
Nothing got lost everything restored

key.core/pipelineTests/records/expected/01_TextblockTransformer/SimpleRecord.java
wadoon: As JMLtk also treats JML, there are some fundamental differences. 

You can observe here the difference between a `JmlDocModifier` and a `JML_NULLABLE`.

`[[JML modifier]]` signals that there are `JmlDocModifier` which are not processed into the appropriate modifier. 

In the next phase (JML processing), JMLtk looks for `JmlDocs` in the AST and processes them. Very similar to `JMLTransformer` in KeY, but then translates the `JmlDocs` into AST nodes and puts them into the tree. 


Yes, this is not correct Java, but it also signals unprocessed JML.

wadoon:
The question is rather: Can we get rid of the stupid ghost and model modifier on Java-level? 
See here: https://github.com/jmltoolkit/jmltk/blob/168dfa039c2d26582e57450c7029da434242daa8/javaparser-key-testing/src/test/java/com/github/javaparser/ast/key/GhostTest.java

Correctly, it should be `/@ghost int x;*/` as the whole thing has JML semantics. 
To be clearly, `/*@ ghost #t #a = #se;*/` should be translated into `/*@ ghost #t #a;*/ /*ghost #a = #se; */`.

unp1:
> The question is rather: Can we get rid of the stupid ghost and model modifier on Java-level? See here: https://github.com/jmltoolkit/jmltk/blob/168dfa039c2d26582e57450c7029da434242daa8/javaparser-key-testing/src/test/java/com/github/javaparser/ast/key/GhostTest.java
> 
> Correctly, it should be `/@ghost int x;*/` as the whole thing has JML semantics. To be clearly, `/*@ ghost #t #a = #se;*/` should be translated into `/*@ ghost #t #a;*/ /*ghost #a = #se; */`.

I assume you mean  `/*@ ghost #t #a;*/ /*@ set #a = #se; */`? But yes, it would be better if JML is not on the Java level.

gradle/libs.versions.toml
```diff
@@ -5,16 +5,16 @@
 # Plugins
 spotless = "8.9.0"
 checkerframework-gradle = "0.6.61"
-maven-publish = "0.37.0"
+maven-publish = "0.36.0"
```
unp1, line 8: Is the downgrade intentional? Otherwise 0.37.0

unp1 (approved):
Thanks a lot. Fine from my side. Only question if you want to have a second review from someone else in addition.

gradle/libs.versions.toml
wadoon: Yes, I got an error that 0.36.0 was not found in the Github CI by Gradle. 
On Maven Central it is available. On clear, whether this was a hiccup, or not...

wadoon:
> Thanks a lot. Fine from my side. Only question if you want to have a second review from someone else in addition.

We need to have a discussion on the needed `ghost` modifier at the Java level. 

In JMLtk, I have the syntax `/*@ ghost <stmt>; */` on the statement level. The following is valid:

```
public void foo() {
  //@ ghost int x = e;
  //@ ghost if(x>2) x = -2;
  //@       else    x =  1/0;
}
```

Allowing JML in Taclets would now allow writing Taclets specific to ghost code, i.e., appropriate handling of exceptions (WD problem). 


All in all, `ghost` should not be considered as a Java modifier for long in JMLtk.

PR #3968, opened 2026-08-16
Fix leaking in ImmutableList#fromArray

### LINKED ISSUE #3967: Immutable lists are mutable
## Description

The immutable data structures are strictly immutable by design.

However currently, they allow indirect modification after initialisation.

## Reproducible

always  

### Steps to reproduce 

Add the following test cases to class `ImmutableListArrayTest`

```java
    @Test
    void immutability1() {
        Integer[] array = { 0, 1, 4};
        ImmutableListArray<Integer> list = new ImmutableListArray<>(array);
        array[2] = 2;
        assertThat(list.get(2)).isEqualTo(4);
    }

    @Test
    void immutability2() {
        Integer[] array = { 0, 1, 4};
        ImmutableList<Integer> list = ImmutableList.fromArray(array);
        array[2] = 2;
        assertThat(list.get(2)).isEqualTo(4);
    }
```

They fail. In both cases the value `list.get(2)` is 2.

### Additional information

---

* Commit: 555a763233a8e17553af5e90c1db94b137fdcc9a

wadoon:
## Related Issue

This pull request closes #3967.

## Type of pull request

## Ensuring quality

    
- I have tested the feature as follows: test cases

### DIFF
--- a/key.util/src/main/java/org/key_project/util/collection/ImmutableList.java
+++ b/key.util/src/main/java/org/key_project/util/collection/ImmutableList.java
@@ -111,7 +111,7 @@ public interface ImmutableList<T extends @Nullable Object>
     /// @param list the list to wrap
     /// @return an ImmutableList containing the elements of the input list
     static <T extends @Nullable Object> ImmutableList<T> fromList(List<T> list) {
-        return new ImmutableListList<>(list);
+        return new ImmutableListList<>(new ArrayList<>(list));
     }
 
     /// Creates an [ImmutableList] from an array.
@@ -123,9 +123,8 @@ public interface ImmutableList<T extends @Nullable Object>
     /// @param array the array to wrap
     /// @return an ImmutableList containing the elements of the input array
     static <T extends @Nullable Object> ImmutableList<T> fromArray(T[] array) {
-        return new ImmutableListArray<>(array);
+        return new ImmutableListArray<>(array.clone());
     }
-
 
     /// Return an empty immutable list.
     ///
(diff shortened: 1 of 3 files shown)


key.util/src/main/java/org/key_project/util/collection/ImmutableList.java
```diff
@@ -246,7 +258,8 @@ default ImmutableList<T> prependReverse(Iterable<T> collection) {
     /// @param array the array to prepend
     /// @return a new list with the array's elements at the beginning
     default ImmutableList<T> prepend(T... array) {
-        return new ImmutableListConcat<>(ImmutableList.fromArray(array), this);
+        // weigl: here we can use the leaky version
+        return new ImmutableListConcat<>(new ImmutableListArray<>(array), this);
```
mattulbrich, line 249: I do not think we can use this here. See the added test cases ...

key.util/src/main/java/org/key_project/util/collection/ImmutableList.java
```diff
@@ -123,6 +123,18 @@ public interface ImmutableList<T extends @Nullable Object>
     /// @param array the array to wrap
     /// @return an ImmutableList containing the elements of the input array
     static <T extends @Nullable Object> ImmutableList<T> fromArray(T[] array) {
+        return new ImmutableListArray<>(Arrays.copyOf(array, array.length));
```
mattulbrich: suggest to use array.clone().

mattulbrich (approved):

key.util/src/main/java/org/key_project/util/collection/ImmutableList.java
mattulbrich: (to make the checker framework happy)

PR #3976, opened 2026-08-19
Support for `<expr> instanceof <Type> <var>`

wadoon:
## Intended Change

This PR adds support for the simple form of the `instanceof` pattern types that came in Java 16: `<expr> instanceof <Type> <var>`. 

Please consult: [JLS 6.3.1](https://docs.oracle.com/javase/specs/jls/se25/html/jls-6.html#jls-6.3.1) for the binding rules. 

## Plan

* [x] Add AST and translation support
* [x] Add first rules 
* [ ] Make the rules parseable
* [ ] Discuss flow analysis as varcond.
* [ ] Add the more complex rules for `while`, `for`, `do`. 
* [x] Example
* [ ] Git history cleanup

## Type of pull request

## Ensuring quality
   

key.core/src/main/resources/de/uka/ilkd/key/proof/rules/java5.key
```diff
@@ -113,3 +115,98 @@
...
     */
 }
+
+\rules {
+// --- instanceof pattern matching binding (Java 16+) ----------------------
+    // These rules handle the variable binding semantics of pattern matching instanceof
+    // Syntax: if(expr instanceof Type varName) { /* varName is bound to expr */ }
+
+    instanceof_pattern_if {
+            \find(\modality{#allmodal}{.. if( #nse instanceof #ty #id ) { #then; } ...}\endmodality (post))
```
mattulbrich: Compare to `ifElseUnfold` in javaRules.key line ~1278.
That should be copyable.

wadoon:
State of [PR 3976](https://github.com/KeYProject/key/pull/3976) zu `instanceof`:

- added varcond \`isBindingExpr(#expression, #posBinding, #negBindings) (true iff one of the bindings is not empty)
- added varcond `alwaysAbnormallyTerminates(#stmts)` (true iff #stmts are never reaches the natural end, e.g., `if(x) { assert false; }`)  _under-approximation_
- added rule: `ifElseUnfoldPattern`, `ifUnfoldPatternUnreachable`, `ifElseUnfoldPattern`
- General-Idea: `if(#bnse) #slist1 else #slist2    ~~~~~> boolean #boolv; {#boolv = #bnse} if(#bnse) { #posBinding; #slist1} else {#negBinding; #slist2}`
- added rules for `#boolv = #nse instanceof #T #id` which **does not match**.

**Todo**:

- [ ] forbid `\ifElseUnfold` for binding expressions

unp1:
Hi, 


> * added rules for `#boolv = #nse instanceof #T #id` which **does not match**.

The rule matches for me in the example below:

```
 /*@ normal_behavior
      @ requires obj != null;
      @ ensures \result instanceof String;
      @*/
    public static Object getString(Object obj) {
        return "ha";
    }

    /*@ normal_behavior
      @ requires obj != null;
      @ ensures \result;
      @*/
    public static boolean isString(Object obj) {
        boolean result = getString(obj) instanceof String id;
        return result;
    }
```

I added the example and a minor fix (replacewith of the `instance_eval_id` taclet)

PR #3977, opened 2026-08-20
Add support for generation of varcond documentation

wadoon:
## Intended Change

Let class `de.uka.ilkd.key.nparser.varexp.TacletBuilderManipulators#main()`
print out Markdown documentation for the current known varconds.

Added therapi-javadoc to write Javadoc into resources to be read at runtime.

## Type of pull request

## Ensuring quality
    
  `.github/workflows/tests.yml`


wadoon:
We might want to discuss whether we want to introduce this library https://github.com/dnault/therapi-runtime-javadoc

Which brings annotations and an annotation processor that writes annotations capturing the Javadoc to `Clazz` and other Java entities for runtime usage

key.core/src/main/java/de/uka/ilkd/key/nparser/varexp/TacletBuilderManipulators.java
```diff
@@ -433,4 +444,118 @@ public static List<TacletBuilderCommand> getConditionBuildersFor(String name) {
                 .collect(Collectors.toList());
     }
     // endregion
+
```
unp1: Could we factor the static methods out in their own class GenerateVarCondDoc or similar?

unp1 (changes requested):
Thanks. No blockers just discussion points.

key.core/src/main/java/de/uka/ilkd/key/nparser/varexp/TacletBuilderManipulators.java
```diff
@@ -3,24 +3,33 @@
...
 import org.key_project.logic.sort.Sort;
 import org.key_project.prover.rules.VariableCondition;
-
+import org.key_project.prover.rules.conditions.NewDependingOn;
+import org.key_project.prover.rules.conditions.NotFreeIn;
+import org.key_project.util.collection.Pair;
+
+import com.github.javaparser.ParserConfiguration;
+import com.github.javaparser.StaticJavaParser;
+import com.github.therapi.runtimejavadoc.ClassJavadoc;
```
unp1: therapi did not receive an update since 3 years and in particular:
`https://github.com/dnault/therapi-runtime-javadoc/issues/82`
might be problematic for the future?

key.core/src/main/java/de/uka/ilkd/key/nparser/varexp/TacletBuilderManipulators.java
wadoon: Maybe we need a general strategy. I think for proof scripts we have a similar problem. 

The strategy could also be `key-tacletdoc`. 

cc: @mattulbrich

key.core/src/main/java/de/uka/ilkd/key/nparser/varexp/TacletBuilderManipulators.java
wadoon: That is a good question. On the one hand, it seems to be stable. On the other hand, it is quite small and forkable.

My other approach was to use JavaParser to read the Javadoc, but then you need to have access to the sources; not so much fun for key-tacletdoc.

wadoon: I see the following entities that should have user documentation: Macros, Proof Script Commands, Variable Conditions, Settings, ...

Everything that comes from key files is documentable in KeY files.

key.core/src/main/java/de/uka/ilkd/key/nparser/varexp/TacletBuilderManipulators.java
wadoon: I put them into a separate package

key.core/src/main/java/de/uka/ilkd/key/nparser/varexp/TacletBuilderCommandInfo.java
```diff
@@ -0,0 +1,287 @@
...
+    /// @return `true` if such a constructor exists, `false` otherwise
+    private static boolean lastArgumentOfFirstConstructorIsBoolean(
+            Class<?> clazz, ArgumentType[] argTypes) {
+        if (findConstructor(clazz, getConstructorClasses(argTypes, true))) {
+            return true;
+        }
+        if (findConstructor(clazz, getConstructorClasses(argTypes, false))) {
+            return false;
+        }
+        throw new IllegalStateException();
```
unp1, line 253: Some text mentioning the clazz and looked for argument types would help to find an issue faster

key.core/src/main/java/de/uka/ilkd/key/nparser/varexp/TacletBuilderCommandInfo.java
```diff
@@ -0,0 +1,287 @@
...
+    ///
+    /// If no constructor matching [#getConstructorClasses()] can be found, argument
+    /// names are filled with empty strings and empty documentation placeholders are used
+    /// instead of failing.
+    private void findDocumentation() {
+        ClassJavadoc classDoc = RuntimeJavadoc.getJavadoc(clazz.getName());
+        generalDocumentation = classDoc;
+
+        try {
+            var constr = clazz.getConstructor(getConstructorClasses());
```
unp1: needs also findConstructor... to find the relevant constructors

key.core/src/main/java/de/uka/ilkd/key/nparser/varexp/TacletBuilderCommandInfo.java
```diff
@@ -0,0 +1,287 @@
...
+        }
+        return argumentDocumentation;
+    }
+
+    /// Computes the parameter types of the constructor that this command's arguments (and,
+    /// if applicable, its negation flag) would map to.
+    ///
+    /// @return the array of expected constructor parameter types
+    Class<?>[] getConstructorClasses() {
+        return getConstructorClasses(argTypes, isNegationSupported == Boolean.TRUE);
```
unp1: - Boolean.TRUE.equals(isNegationSupported) 
- iNegationSupported is currently null for constructor based builders: Vardoc calls argNames() -> getConstructorClasses() in line 67 before isNegationSupported() which is called afterwards and hence could not set the field

key.core/src/main/java/de/uka/ilkd/key/nparser/varexp/TacletBuilderCommandInfo.java
```diff
@@ -0,0 +1,287 @@
...
+
+            final var parameters = constr.getParameters();
+            argNames = new String[argTypes.length];
+            for (int i = 0; i < argTypes.length; i++) {
+                argNames[i] = parameters[i].getName();
+            }
+
+            constructorDeclaration.ifPresent(it -> {
+                List<ParamJavadoc> params = it.getParams();
+                for (int i = 0; i < params.size(); i++) {
```
unp1: could cause array index out of bounds exception as params size might be larger than argTypes size, when negation flag is set and documented

key.core/src/main/java/de/uka/ilkd/key/doc/VarcondDoc.java
```diff
@@ -0,0 +1,163 @@
...
+                cleanJavadoc(new CommentFormatter().format(generalDocumentation.getComment())));
+
+            System.out.format("\n**Signatures**\n\n");
+            for (TacletBuilderCommandInfo cmd : cmds) {
+
+                final var arguments = Streams.zip(
+                    Arrays.stream(cmd.argNames()),
+                    Arrays.stream(cmd.argumentTypes())
+                            .map(Enum::toString),
+                    "%s%s"::formatted)
```
unp1: %s%s::formatted put no separator between name and type

key.core/src/main/java/de/uka/ilkd/key/doc/VarcondDoc.java
```diff
@@ -0,0 +1,163 @@
...
+                argumentInformation.getParams().stream()
+                        .map(tag -> "   * `%s` %s".formatted(tag.getName(), tag.getComment()))
+                        .forEach(System.out::println);
+                System.out.println();
+            }
+        }
+    }
+
+    private static String indent(String spaces, String it) {
+        return spaces + it.replace("\n", spaces);
```
unp1: spaces + it.replace("\n", "\n"+spaces) ?

key.core/src/main/java/de/uka/ilkd/key/nparser/varexp/TacletBuilderCommandInfo.java
```diff
@@ -0,0 +1,287 @@
...
+                    argNames[i] = params.get(i).getName();
+                }
+            });
+
+            argumentDocumentation = constructorDeclaration
+                    .orElse(new MethodJavadoc(constr.getName(), null, null, null, null, null, null,
```
unp1: MethodJavadoc.createEmpty(constr)

key.core/src/main/java/de/uka/ilkd/key/nparser/varexp/TacletBuilderCommandInfo.java
```diff
@@ -0,0 +1,287 @@
...
+                        null));
+        } catch (NoSuchMethodException e) {
+            argNames = new String[argTypes.length];
+            Arrays.fill(argNames, "");
+            argumentDocumentation = new MethodJavadoc("", null, null, null, null, null, null, null);
```
unp1: MethodJavadoc.createEmpty(constr)

key.core/src/main/java/de/uka/ilkd/key/doc/VarcondDoc.java
```diff
@@ -0,0 +1,163 @@
...
+                System.out.printf("* `%s(%s)`\n", name, arguments);
+
+                if (cmd.isNegationSupported()) {
+                    System.out.printf("* `\\not%s(%s)`\n", name, arguments);
+                }
+
+                System.out.println();
+                final var argumentInformation = cmd.getArgumentInformation();
+
+                System.out.println(indent("   ", cleanJavadoc(argumentInformation.toString())));
```
unp1: argumentInformation.toString() returns a not nicely readable JSON like representation.

key.core/src/main/java/de/uka/ilkd/key/nparser/varexp/TacletBuilderManipulators.java
unp1, line 264: EnumTypeCondition declares only two parameters

unp1 (changes requested):
Thanks! Concerning the failing WD test, that constructor for ConstantCondition takes a JAbstractSortedOperator which is not a parseable variable. 

(also "\reference" does not take a variable)

key.core/src/main/java/de/uka/ilkd/key/nparser/varexp/TacletBuilderManipulators.java
unp1: Thanks

key.core/src/main/java/de/uka/ilkd/key/nparser/varexp/TacletBuilderCommandInfo.java
wadoon: At this point we have already checked whether we expect negation or not. Hence, the expected argument classes are determined. We either find the constructor directly or not. If not present, the call returns exceptional.

key.core/src/main/java/de/uka/ilkd/key/nparser/varexp/TacletBuilderCommandInfo.java
wadoon: Yeah, the previous version just looked at the last parameter of the  first constructor. Now all arguments must be met.

PR #3978, opened 2026-08-21
Rendering performance improvements

unp1:
## Intended Change

Improves rendering speed of proof tree and sequent view by
* rendering the applied on information (potentially long term) first when tooltip is shown (proof tree)
* caching locally declared fields (sequent view and proof tree)

The proof tree part could lead to a minute long wait even for rather small proofs on at least macOS
where a JTree is walked by the OSes accessibility feature to provide navigation via voice over. The walk
caused to render the term part of the tooltip for each node.

The second part helps the proof tree and sequent view by caching an information often accessed by 
the pretty printer to decide how to print an attribute name.

## Type of pull request

## Ensuring quality

    
- I have tested the feature as follows: the render is now instantenuous on macos and the shown information on the proof tree and tooltips is identical.

### DIFF
--- a/key.core/src/main/java/de/uka/ilkd/key/java/KeYProgModelInfo.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/java/KeYProgModelInfo.java
@@ -69,6 +69,11 @@ public class KeYProgModelInfo {
     /** Size of the type model {@link #subtypeCache} was computed from. */
     private int subtypeCachedSize = -1;
 
+    /**
+     * Caches locally declared fields; does not need to be invalidated as
+     * an existing class declaration is immutable
+     */
+    private final Map<KeYJavaType, List<Field>> localDeclaredFields = new LinkedHashMap<>();
 
 
     public KeYProgModelInfo(JavaService service) {
@@ -460,9 +465,20 @@ public class KeYProgModelInfo {
         if (ct.getJavaType() instanceof ArrayType) {
             return getVisibleArrayFields(ct);
         }
-        return getReferenceType(ct)
+        List<Field> fields;
+        synchronized (localDeclaredFields) {
+            fields = localDeclaredFields.get(ct);
+            if (fields != null) {
+                return fields;
+            }
+        }
+        fields = Collections.unmodifiableList(getReferenceType(ct)
                 .map(r -> asKeYFieldsR(r.getDeclaredFields().stream()))
-                .orElseGet(ArrayList::new);
+                .orElseGet(ArrayList::new));
+        synchronized (localDeclaredFields) {
+            localDeclaredFields.put(ct, fields);
+        }
+        return fields;
     }
 
 
--- a/key.core/src/main/java/de/uka/ilkd/key/java/KeYProgModelInfo.java
+++ b/key.core/src/main/java/de/uka/ilkd/key/java/KeYProgModelInfo.java
@@ -4,6 +4,7 @@
 package de.uka.ilkd.key.java;
 
 import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
@@ -73,7 +74,8 @@ public class KeYProgModelInfo {
      * Caches locally declared fields; does not need to be invalidated as
      * an existing class declaration is immutable
      */
-    private final Map<KeYJavaType, List<Field>> localDeclaredFields = new LinkedHashMap<>();
+    private final Map<KeYJavaType, ImmutableList<Field>> localDeclaredFields =
+        new ConcurrentHashMap<>();
 
 
     public KeYProgModelInfo(JavaService service) {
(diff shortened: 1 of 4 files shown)


key.core/src/main/java/de/uka/ilkd/key/java/KeYProgModelInfo.java
```diff
@@ -69,6 +69,11 @@ public class KeYProgModelInfo {
     /** Size of the type model {@link #subtypeCache} was computed from. */
     private int subtypeCachedSize = -1;
 
+    /**
+     * Caches locally declared fields; does not need to be invalidated as
+     * an existing class declaration is immutable
+     */
+    private final Map<KeYJavaType, List<Field>> localDeclaredFields = new LinkedHashMap<>();
```
wadoon: Why not a `ConcurrentMap` implementation? `LinkedHashMap` seems unnecessarily here. 

Also we have Guava, offering thread-safe [caches](https://javadoc.io/static/com.google.guava/guava/33.7.1-jre/com/google/common/cache/Cache.html)

key.core/src/main/java/de/uka/ilkd/key/java/KeYProgModelInfo.java
```diff
@@ -460,9 +465,20 @@ public List<Field> getAllFieldsLocallyDeclaredIn(KeYJavaType ct) {
         if (ct.getJavaType() instanceof ArrayType) {
             return getVisibleArrayFields(ct);
         }
-        return getReferenceType(ct)
+        List<Field> fields;
+        synchronized (localDeclaredFields) {
```
wadoon: Check if read access really needs to protected. Normally ReadWriteLock would be better.

key.core/src/main/java/de/uka/ilkd/key/java/KeYProgModelInfo.java
wadoon: We also could consider to make this `ThreadLocal` and have every thread one cache for its own.

key.ui/src/main/java/de/uka/ilkd/key/gui/prooftree/ProofTreeView.java
```diff
@@ -1226,7 +1227,16 @@ private static String renderTooltip(Style.Tooltip tooltip) {
             } else {
                 result.append(" ");
             }
-            result.append(fragment.value);
+            if (fragment.value instanceof final PosInOccurrence pio) {
+                if (pio != null) {
```
wadoon: instanceof checks also for nullness

key.ui/src/main/java/de/uka/ilkd/key/gui/prooftree/Style.java
```diff
@@ -107,15 +110,15 @@ public List<Fragment> getAdditionalInfos() {
         }
 
         /** wrapper class for additional infos */
-        public static final class Fragment {
+        public static final class Fragment<T> {
```
wadoon: I do not see the point, why this is now parameterized. You can just make this to an `Object`.

key.ui/src/main/java/de/uka/ilkd/key/gui/prooftree/ProofTreeView.java
```diff
@@ -1197,7 +1198,7 @@ public void valueChanged(TreeSelectionEvent e) {
         }
     }
 
-    private static String renderTooltip(Style.Tooltip tooltip) {
+    private static String renderTooltip(Style.Tooltip tooltip, Services services) {
         String title = tooltip.getTitle();
         List<Style.Tooltip.Fragment> fragments = tooltip.getAdditionalInfos();
```
wadoon, line 1203: `Fragment<?>`? Just remove the type parameter, no value here.

wadoon (changes requested):

key.core/src/main/java/de/uka/ilkd/key/java/KeYProgModelInfo.java
unp1: Thanks. Changed to ConcurrentMap. The races when reading and writing to the cache are harmless as in the worst case the same list is computed twice. Decided in this case against ThreadLocal as after an initial phase, the cache is usually only read and threads can benefit from each others work.

key.ui/src/main/java/de/uka/ilkd/key/gui/prooftree/Style.java
unp1: Thanks fixed.

key.ui/src/main/java/de/uka/ilkd/key/gui/prooftree/ProofTreeView.java
unp1: Thanks fixed.

key.ui/src/main/java/de/uka/ilkd/key/gui/prooftree/ProofTreeView.java
unp1: Oops. Thanks fixed.

key.core/src/main/java/de/uka/ilkd/key/java/KeYProgModelInfo.java
unp1: After change to ConcurrentMap, no synchronization needed as races are harmless leading in the worst case to double work.

wadoon (approved):

PR #3979, opened 2026-08-22
Tweaks for Immutable* data structures

unp1:
## Intended Change

Minor tweaks to Immutable data structures:

- DefaultImmutableSet: ensure set property "no duplicates" maintained when constructing from a Collection
- ImmutableList: avoid unnecessary copying of list
- DefaultImmutableMap: 
       Use hashcode computation as by java.util.HashMap to reduce collision.
       Data from a 'pathological' case taken from TPTP:
       **Context**: The problem is in examples/standard_key/pred_log/tptp/NUM/NUM325+1.key 
       It cannot be solved at the moment by KeY as it leads after ~180 nodes to an OOM (that is part of another optimisation effort)
       **Results**:
         - New hashcode: OOM reached within 105s (187 nodes)
        - Old hashcode: OOM reached after 900s (187 nodes)
      So we reach an OOM faster :-) after same amount of work.

## Type of pull request

## Ensuring quality

    
- I have tested the feature as follows: standard CI tests 

### DIFF
--- a/key.util/src/main/java/org/key_project/util/collection/DefaultImmutableSet.java
+++ b/key.util/src/main/java/org/key_project/util/collection/DefaultImmutableSet.java
@@ -63,9 +63,12 @@ public class DefaultImmutableSet<T extends @Nullable Object> implements Immutabl
     }
 
     public static <T> ImmutableSet<T> fromCollection(Collection<T> seq) {
-        // ensure set property, each element only once and order of collection maintained
-        // to avoid regression with previous behavior
-        return seq.isEmpty() ? nil() : fromSet(new LinkedHashSet<>(seq));
+        if (seq instanceof Set<T> set) {
+            return fromSet(set);
+        } else {
+            // ensure set property "each element only once"
+            return seq.isEmpty() ? nil() : fromImmutableList(ImmutableList.fromList(seq));
+        }
     }
 
     public static <T> ImmutableSet<T> fromSet(Set<T> set) {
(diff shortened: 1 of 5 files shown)


key.util/src/main/java/org/key_project/util/collection/DefaultImmutableSet.java
```diff
@@ -63,7 +63,14 @@ private DefaultImmutableSet(ImmutableList<T> elementList) {
...
     public static <T> ImmutableSet<T> fromCollection(Collection<T> seq) {
-        return new DefaultImmutableSet<>(ImmutableList.fromList(seq));
+        // ensure set property, each element only once and order of collection maintained
+        // to avoid regression with previous behavior
+        return seq.isEmpty() ? nil() : fromSet(new LinkedHashSet<>(seq));
+    }
+
+    public static <T> ImmutableSet<T> fromSet(Set<T> set) {
+        // we can rely on the set property here
+        return set.isEmpty() ? nil() : new DefaultImmutableSet<>(ImmutableList.fromList(set));
```
wadoon, line 76: Double copy. Immutable.fromList copies.

wadoon (changes requested):

key.util/src/main/java/org/key_project/util/collection/DefaultImmutableSet.java
unp1: Just that I fix on the right place: The double copying is when coming from the code path via `fromCollection(..)` right or do I miss something? As we need the copy by fromList when invoking fromSet directly with a provided set.

wadoon: yeah, via `fromCollection()`. You can just call the constructor there.

unp1: "fromCollection" takes now the same route as "fromList" and duplicates are removed via Immutables removeDuplicates (no longer via a full copy)

wadoon (approved):

wadoon (approved):

PR #3981, opened 2026-08-23
Fix for slicing in case of dynamic taclet applications and renamings

unp1:
## Intended Change

Analysing whether some proofs from the StipuLa case study can be made smaller using slicing, 
an exception was encountered concerning a dynamically loaded taclet could not be found.

The reasons was that the replay for slicing omitted to also check the locally introduced taclet information
in nodes. Adding this revealed a second problem with renaming (taclets with addprogvars).

Both issues have been fixed.

Added also the StipuLa case study problems and a test for slicing.

## Type of pull request

## Ensuring quality

    
- I have tested the feature as follows: slicing tests, standard tests by CI


Drodt (approved):
Thanks!

